PR #72, opened 2015-04-25
add david-dm npm package dependencies badges

PeterDaveHello:


metaraine:
Thanks! I'm not sure I see the benefit of this. With how quickly node packages evolve, there is nearly always going to be something out of date. I am not a full-time developer so I cannot update it every day. It seems to me that updating the dependencies at every normal release (whether it is major, minor, or patch) is a reasonable approach.

I am open to being convinced otherwise, with good reasoning.

PeterDaveHello:
I think the best feature of it is the insecure dependencies detection, like this
Imgur

I like it!

PeterDaveHello:
I think npm-check-updates was designed to keep every think up-to-date, and I noticed the .travis.yml, if you enabled the CI build test(if it exist), the upgrade process will be easier.

metaraine:
Thanks, that's fair. I like the "insecure" warning, too.

I think @tjunnone has to enable CI.

PR #93, opened 2015-07-03
Require get-stdin-promise only if needed

olebed:
If the get-stdin-promise has been 'required' but not used,
it can prevent nodejs process from being terminated.


metaraine:
Good point, thanks!

PR #96, opened 2015-07-11
Changed behavior according to #94

### LINKED ISSUE #94: Weird behavior: in unstable, ^1.1.0 is not updated to ^1.2.1
The unstable version (v2.0.0-alpha.11) returns different results from the stable version (v1.5.1).
In my package.json I have the following dependency:
```
  "dependencies": {
    ...
    "bindings": "^1.1.0",
    ...
  }
```
The `bindings` latest version is 1.2.1 (ATTOW). When I run the unstable version of `ncu` it doesn't list the update, but when I run the stable version it does.
The reason it doesn't list the update is because in the `isUpgradeable` function we test if the latest version is satisfied by the current version and if it does, we skip it (1.2.1 is satisfied by ^1.1.0). I guess the reason this test was added is because if the latest version is satisfied by the current version, all you have to do to get it is `npm install`. 
Some users, myself included, expect `ncu` to let us know of any update. If I had run `npm i bindings --save` when the version was 1.1.0 and then I run `ncu` today when the version is 1.2.1, I won't get any warning...
Also, the documentation in the readme file (for v2), in the "How dependency updates are determined" section states that `ncu` will do the following:
> ^1.2.0 => ^1.3.0

But it's not how v2 currently works.

I suggest removing the `!semver.satisfies(latest, version)` so we will be notified of any update. 
At the very list we should update the documentation to make it clear that v2 doesn't show updates to versions that are satisfied by the current version.

etiktin:
Resolves #94.

Change-log:
- Default behavior without -u to shows all upgrades (e.g. for satisfied versions we show `"<package>" satisfies current dependency (Installed: <old>, Latest: <new>)`)
- Default behavior with -u only upgrades those that are not satisfied (like the current behavior of v2 alpha)
- Using -ua (-a) / --upgradeAll flags to upgrade even satisfied versions (like the behavior of v1)
- The note at the end of the version upgrades was changed to:
```
Run with '-u' to upgrade your package.json
Run with '-ua' to upgrade even those that satisfy the declared version range
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ results
 npm-debug.log
 
 node_modules
+
+.idea
--- a/bin/npm-check-updates
+++ b/bin/npm-check-updates
@@ -9,7 +9,6 @@ program
     .usage('[options] [filter]')
     .option('-d, --dev', 'check only devDependencies')
     .option('-e, --error-level <n>', 'set the error-level. 1: exits with error code 0 if no errors occur. 2: exits with error code 0 if no packages need updating (useful for continuous integration). Default is 1.', cint.partialAt(parseInt, 1, 10), 1)
-    .option('-f, --force', 'force upgrade even when the latest version satisfies the declared semver dependency')
     .option('-g, --global', 'check global packages instead of in the current project')
     // program.json is set to true in programInit if any options that begin with 'json' are true
     .option('-j, --jsonAll', 'output new package.json instead of human-readable message')
@@ -19,6 +18,7 @@ program
     .option('-s, --silent', "don't output anything")
     .option('-t, --greatest', "find the highest versions available instead of the latest stable versions")
     .option('-u, --upgrade', 'upgrade package.json dependencies to match latest versions (maintaining existing policy)')
+    .option('-a, --upgradeAll', 'upgrade package.json dependencies even when the latest version satisfies the declared semver dependency')
 
 program.parse(process.argv);
 
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -35,7 +35,7 @@ function upgradePackageDefinitions(currentDependencies) {
     var dependencyList = Object.keys(currentDependencies);
     return vm.getLatestVersions(dependencyList, {
         versionTarget: options.greatest ? 'greatest' : 'latest',
-        registry: options.registry ? options.registry : null,
+        registry: options.registry ? options.registry : null
     }).then(function (latestVersions) {
         var upgradedDependencies = vm.upgradeDependencies(currentDependencies, latestVersions);
         return [upgradedDependencies, latestVersions];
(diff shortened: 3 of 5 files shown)


etiktin:
Still need to add tests and update the documentation.

metaraine:
Awesome!!! Thank you, this looks great!

Careful of your auto-linter... the extra noise in the diff made it harder to see the real changes. Please stick to the conventions (or lack of conventions) of the repo authors.

I may tweak the wording for clarity. It would be nice to still see the version numbers when it satisfies the current dependency.

Thanks for doing this! Well done!

etiktin:
Your welcome :)

You're right. I should have waited with the style changes (maybe a different PR) because it distracts from the meaningful changes. It's just that my editor had a hard time with the combination of 2 space tabs and 4 space tabs.

I agree that the wording my need some more work and that we should show the current version range. Maybe something like:
```
"bindings" version range ^1.1.0, satisfies the latest version (Installed: 1.1.0, Latest: 1.2.1)
```

P.S.
I just committed an update to the documentation to reflect the changes, but I guess once you merge the PR, I can't push to it anymore. Any way, if you're intrested the commit is at: https://github.com/etiktin/npm-check-updates/commit/85ae9db9b809da0cda826f0c4db9d76746b0e0fd

metaraine:
Thanks!!! I updated the README before this came in, so I think we're set.

I double checked the semantics of "satisfies" in the [node-semver](https://github.com/npm/node-semver) readme, and it is "version *satisfies* range." Maybe something like: `"bindings" latest version satisfies ^1.1.0 (Installed: 1.1.0, Latest: 1.2.1)`.

It may not be worth fine-tuning the message at the moment, however, as I am hoping to rewrite the printing logic to make it similar to [salita](https://github.com/tbranyen/salita). I really like their output. Thoughts?

etiktin:
The message you're suggested is better than the current one, so let's use it for the time being.

I like the way `salita` presents the information. The wording is very simple and the tabs and coloring makes it easy to consume. The status at the beginning of each line is also a nice touch. It's missing installed and latest which I think are important.

How about displaying the information in a table format instead of trying to phrase the changes?

For example:

Update range?  | Dependency | Current | Latest
------------- | ------------- | ------------- | -------------
Update  | express | 3.10.1 (^3.0.0) | 4.11.0 (^4.11.0)
Keep  | bindings | 1.10.1 (^1.10.1) | 1.21.1 (^1.21.1)

- For latest, isn't it always going to be `latest (^latest)` for updates and `latest (current range)` for satisfies? If that's the case, it's possible to drop the latest range because users can figure it out from a combination of the status column + current/latest.  
- Regarding the status (update range), maybe it should use symbols instead of text (like in git where `+` is addition, `-` is removal etc.) and maybe it makes more sense if it was the last column instead of the first. 
- The verbs in the example can change according to type of run (e.g. "Update" to "Updated", "Keep" to "Kept").

What I like about the table format is that it's noob friendly :) You don't really need to understand ranges and semver.

What do you think, is it too radical :) ?

metaraine:
Moving discussion back to #94.

PR #100, opened 2015-07-14
Fix old alias handling

### LINKED ISSUE #90: ncu vs npm-check-updates [Can someone with Windows help?]
I saw this output in another issue, but I don't have Windows to test. Can someone with Windows test if you get the following message with npm-check-updates@2.0.0-alpha.10?

    $ ncu
    You can now use "ncu" for less typing!

The "You can now use..." message is only supposed to be shown when you type out the full "npm-check-updates". 

Thank you! If you'd like to be *even more helpful*, can you verify what process.argv returns? The comparison for this is `path.basename(process.argv[1]) === 'npm-check-updates'`, and maybe there is different behavior on Windows.

etiktin:
This resolves #90

Instead of using argv to check if `npm-check-updates` was called or `ncu`, we load different files for each bin. When `npm-check-updates` is called, `/bin/old-alias` is loaded and prints:
"You can now use "ncu" for less typing!"
We then require the regular startup script and continue as usual.

### DIFF
--- a/bin/old-alias
+++ b/bin/old-alias
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
 
-process.env.ncu_old_alias = 'true';
+console.log('You can now use "ncu" for less typing!');
 
 require('./npm-check-updates');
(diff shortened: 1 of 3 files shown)


metaraine:
I love easy fixes ^_^.

Thanks Eran!

PR #107, opened 2015-07-25
Update msgs style for consistency

etiktin:
Added some color where it makes sense.
Fixed some spacing issues with the upgrades table (e.g. if all upgrades are satisfied there's no need for 3 lines spacing before the satisfied message).
Basically, I just tried to make sure that everything is consistent.

Related to #94.


metaraine:
:+1:

PR #114, opened 2015-08-01
Output the -u msg only if unsatisfied > 0

etiktin:
If you only have satisfied updates, we still show the "Run with -u to upgrade your package.json". I think we shouldn't show it, because it won't do anything in this specific case.


metaraine:
Thanks! Implemented these in a [separate commit](https://github.com/tjunnone/npm-check-updates/commit/70b508310b3c45f1828ec779d033073029276fc7) since I would have had to deal with the merge conflict  anyway.

Changes published in `v2.0.3.`

PR #127, opened 2015-08-14
Update CLI flag for upgradeAll

lauritzsh:
Should match the output when running `ncu --help` which is just `-a, --upgradeAll`.


lauritzsh:
...in the README, should have made that clear.

metaraine:
Thanks

PR #132, opened 2015-09-12
Provide helpful errors on JSON.parse exceptions

etiktin:
This commit uses `json-parse-helpfulerror` module to parse JSONs instead of `JSON.parse`. The benefit of using it, is that when the parse fails and throws, it provides clues to why it failed (e.g. "Trailing comma at 15:3"). So it's easier to figure out, what's wrong and fix it. 
The performance should pretty much be the same, since it first tries to use `JSON.parse` and only if it throws, we analyzes the cause.

`closestPackage.sync` that we call when we try to find the closest package,json, can throw exceptions because it tries to parse the content of the file, so I wrapped it in a try catch to give a better error message. I also opened a PR [there](hughsk/closest-package#1), so they will also throw more helpful errors (e.g. the new errors will include the path to the package.json that caused the exception, so users will know what to fix).

This commit continues the work started in https://github.com/tjunnone/npm-check-updates/commit/a98bef5ecda0113aff6324ffc193412fff24c70d

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -14,6 +14,7 @@ var chalk = require('chalk');
 var fs = Promise.promisifyAll(require('fs'));
 var vm = require('./versionmanager');
 var versionUtil = require('./version-util');
+var jph = require('json-parse-helpfulerror');
 
 //
 // Helper functions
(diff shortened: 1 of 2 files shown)


metaraine:
Looks great! Thanks!

PR #133, opened 2015-09-12
Added Node.js v4 and io.js v3 to .travis.yml

etiktin:


metaraine:
Can you explain the recommendation of specifying multiple versions in this section of the .travis.yml? Does it cause travis to run the unit tests on each node version separately? Is that the recommended approach? Thanks!

etiktin:
Sure. Specifying multiple versions, means travis will run the tests using each one of them (look at the build log [here](https://travis-ci.org/tjunnone/npm-check-updates/builds/80009525), in each one it states what was the version used for testing). This way we know that the code works on each stable version of node. It answers questions such as "did we add a module that breaks support for v0.10?", or "did we use ES6 syntax that is not supported in v0.10 and v0.12 - at least without flags?". 
This is the approach used by most modules.

Travis also supports running on the latest stable version, by adding the keys 'stable' and 'iojs', but I think it's better to be explicit about the versions you want (if we didn't, when v4 was released we would have lost testing on v0.12).

etiktin:
This was replaced by #139.

PR #142, opened 2015-09-23
Add update notifier

### LINKED ISSUE #141: Update notifications
What do you think about adding the [`update-notifier`](https://www.npmjs.com/package/update-notifier) module to `ncu`?

The idea is that when you run `ncu` through the cli, the module will check with npm, if there's an update to `ncu`. If there is an update it will print a notification to the console (see screenshot in the module page).
If the users are using recent versions, it means they get bug fixes faster and it will be easier to provide support when an issue arises.

etiktin:
Added the `update-notifier` module to `ncu`. It will check for updates to `ncu` upto once a day. If there's an update it will show a message (only in CLI).

This resolves #141.


etiktin:
To test it, you need to change the package.json version and run it. Then add to your system clock at least a day and re-run the tool twice and you will see the notification.
Explantation: The first run saves the current timestamp, so the first update test will happen at least 24 hours later. The second run figures out that there's an update, but since it runs in a separate process it will show the update only on the next run, so that's why you need the 3 runs to test it.

This is how it looks like on Windows with ConEmu:
updatenotification

I think the spacing between the last dependency and the notification seems a bit much, but other than that it fits pretty good.
Raine, what do you think?

metaraine:
Looks great, thanks!

Can we output the notifications first, before anything else? (precedent: yeoman)

etiktin:
Yep. Here it is:

updatenotificationpre

It looks better and the user gets it faster (before, you had to wait until ncu finished testing dependencies which could take some time).

PR #255, opened 2016-07-14
Specify `ncu` where instructions appear in output

r24y:
I'd like to add `ncu` to a postinstall step of a work project. The current output ends in `Run with -u to upgrade package.json` which might lead developers (coworkers who may not be as familiar with npm) to try `npm install -u`; not the intended effect!

This PR clarifies that it is `ncu` that must be run, not `npm install`.

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -286,7 +286,7 @@ function printUpgrades(args) {
         }, {
             greatest: options.greatest
         });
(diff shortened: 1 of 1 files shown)


raineorshine:
Makes sense! Thanks!

billyjanitsch:
@r24y A small mistake was added here: `chalk.cyan('ncu ') + chalk.blue(options.packageManager + ' update')` yields `ncu npm update` instead of the previously-correct `npm update`.

r24y:
Whoops! Sorry about the error; thank you @billyjanitsch for catching it.

PR #257, opened 2016-07-19
Revert specifying `ncu` for non-ncu command

billyjanitsch:
Fixes the mistake introduced in #255 (explained [here](https://github.com/tjunnone/npm-check-updates/pull/255#issuecomment-233529203)).


raineorshine:
Thanks! I will republish as soon as I am at my computer again.

PR #297, opened 2016-11-26
Fix Windows global prefix not set

anantoghosh:
Sets prefix manually to %AppData%\npm on windows in global mode. Fixes https://github.com/tjunnone/npm-check-updates/issues/163.


raineorshine:
Looks good! Do you think we need to test it on other Windows OS's? What would be the best way to do that?

anantoghosh:
I searched a bit, it does look like default global install location is %appdata%\npm across all versions windows. This should be safe.

raineorshine:
Okay!

PR #302, opened 2016-12-04
Fixes npm ENOENT error on Windows.

anantoghosh:
Windows requires spawn to execute npm with the .cmd extension.
Fixes https://github.com/tjunnone/npm-check-updates/issues/301


raineorshine:
Good catch! Thanks!

PR #309, opened 2016-12-30
Auto Dependency Updates ...

lguzzon:


raineorshine:
Thanks for the pull request!

There are a few considerations here:

- Despite the behavior of ncu itself, it is unnecessary to upgrade the lower bound on dependency ranges, i.e. ^3.4.3 does not need to be upgraded to ^3.4.7. The range specifier already includes all versions up to but not including 4.0.0. A standard `npm update` will install newer versions within the specified range. Forcing the version to the latest here adds to the commit diff without any functional change.
- `find-up` needs to be pinned to v1. v2 has breaking changes that are not supported by ncu.
- There may be larger implications of upgrading npm, which must match the npm of the user's system. This a known issue. The solution is to discontinue use of the internal npm and only use the user's system-installed npm.

In conclusion, the only upgrade I would consider at this time is npm, and that needs to come with some changes to the architecture in order to use the system install of npm.

lguzzon:
Ok,
keep going ...

raineorshine:
Here's how we use npm [currently](https://github.com/tjunnone/npm-check-updates/blob/master/lib/package-managers/npm.js#L21):

```
var npm = require('npm');
...
npm.commands.viewAsync([packageName, field], true).then(...)
npm.commands.listAsync(args || [], true).then(...)
...
```

One problem is that the `npm` was never intended to be used as a node module, as their documentation has made clear. The bigger problem is that this uses the version of npm in npm-check-updates's dependencies, which could be different than the user's installed version. It will fix a lot of issues to instead spawn a child process that uses the system npm to make these calls.

This is actually [already done](https://github.com/tjunnone/npm-check-updates/blob/master/lib/package-managers/npm.js#L94) in one place, but needs to be done everywhere:

```
spawn('npm', ['ls', '--json', '-depth=0']).then(...)
```

That's the bigger change that would allow us to remove npm as a direct dependency completely, and instead treat npm as a peer dependency. Does that makes sense?

PR #311, opened 2017-01-10
Optionally include prereleases

### LINKED ISSUE #308: [FEATURE REQUEST] Exclude prerelease tags while checking for updates
I was wondering if an extra flag to exclude prerelease tags is something you would consider adding? 

_Example_:

- Current `package.json`:

```json
...
"dependencies": {
    "react": "0.14.6"
}
```

- Update suggested by `ncu -p --semverLevel major -u` -> `0.15.0-alpha.1`

- Actual version I want: `0.14.8` (which is the latest after `0.15.0-alpha.1`)

I'd be happy to work on it and get it in!

karanjthakkar:
Fixes #308

@raineorshine This is  basic impl of handling the `--pre` flag for `latest` versions. Let me if I'm on track :)

### DIFF
--- a/bin/ncu
+++ b/bin/ncu
@@ -11,7 +11,6 @@ updateNotifier({pkg: pkg})
     .notify({defer: false});
 
 program
-    .version(require('../package').version)
     .description('[filter] is a list or regex of package names to check (all others will be ignored).')
     .usage('[options] [filter]')
     .option('-d, --dev', 'check only devDependencies')
@@ -36,7 +35,9 @@ program
     .option('-x, --reject <matches>', 'exclude packages matching the given string, comma-delimited list, or regex')
     .option('-a, --upgradeAll', 'include even those dependencies whose latest version satisfies the declared semver dependency')
     .option('--semverLevel <level>', 'find the highest version within "major" or "minor"')
-    .option('--pre', 'include prereleases');
+    .option('--pre', 'include prereleases')
+    .option('-v, --version', pkg.version, function() { console.log(pkg.version); process.exit(0); })
+    .option('-V', '', function() { console.log(pkg.version); process.exit(0); });
 
 program.parse(process.argv);
 
--- a/bin/ncu
+++ b/bin/ncu
@@ -36,6 +36,7 @@ program
     .option('-a, --upgradeAll', 'include even those dependencies whose latest version satisfies the declared semver dependency')
     .option('--semverLevel <level>', 'find the highest version within "major" or "minor"')
     .option('--pre', 'include prereleases')
+    .option('--removeRange', 'remove version ranges from the final package version')
     .option('-v, --version', pkg.version, function() { console.log(pkg.version); process.exit(0); })
     .option('-V', '', function() { console.log(pkg.version); process.exit(0); });
 
--- a/bin/ncu
+++ b/bin/ncu
@@ -35,7 +35,8 @@ program
     .option('-u, --upgrade', 'overwrite package file')
     .option('-x, --reject <matches>', 'exclude packages matching the given string, comma-delimited list, or regex')
     .option('-a, --upgradeAll', 'include even those dependencies whose latest version satisfies the declared semver dependency')
-    .option('--semverLevel <level>', 'find the highest version within "major" or "minor"');
+    .option('--semverLevel <level>', 'find the highest version within "major" or "minor"')
+    .option('--pre', 'Include prereleases in the final result');
 
 program.parse(process.argv);
 
--- a/bin/ncu
+++ b/bin/ncu
@@ -36,7 +36,7 @@ program
     .option('-x, --reject <matches>', 'exclude packages matching the given string, comma-delimited list, or regex')
     .option('-a, --upgradeAll', 'include even those dependencies whose latest version satisfies the declared semver dependency')
     .option('--semverLevel <level>', 'find the highest version within "major" or "minor"')
-    .option('--pre', 'Include prereleases in the final result');
+    .option('--pre', 'include prereleases');
 
 program.parse(process.argv);
 
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -122,7 +122,7 @@ function upgradePackageDefinitions(currentDependencies) {
     return vm.getLatestVersions(currentDependencies, {
         versionTarget: versionTarget,
         registry: options.registry ? options.registry : null,
-        includePrerelease: options.pre
+        pre: options.pre
     }).then(function (latestVersions) {
 
         var upgradedDependencies = vm.upgradeDependencies(currentDependencies, latestVersions);
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -300,7 +300,7 @@ function printUpgrades(args) {
             greatest: options.greatest,
             newest: options.newest
         });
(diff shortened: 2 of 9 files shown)


lib/package-managers/npm.js
```diff
@@ -107,7 +107,19 @@ module.exports = {
             npm.commands.listAsync(args || [], true); // silent:true
     },
 
-    latest: cint.partialAt(view, 1, 'dist-tags.latest'),
+    latest: function (packageName, currentVersion, includePrerelease) {
+        if (includePrerelease) {
```
raineorshine: This needs to work a little differently. It is possible that there is a higher number released just not on the latest tag. We don't want to accidentally pick that up. You should always try the latest tag first. If that ends up being a prerelease, then implement the logic to get the highest non-prerelease number instead (unless includePrerelease is specified).

bin/ncu
```diff
@@ -35,7 +35,8 @@ program
     .option('-u, --upgrade', 'overwrite package file')
     .option('-x, --reject <matches>', 'exclude packages matching the given string, comma-delimited list, or regex')
     .option('-a, --upgradeAll', 'include even those dependencies whose latest version satisfies the declared semver dependency')
-    .option('--semverLevel <level>', 'find the highest version within "major" or "minor"');
+    .option('--semverLevel <level>', 'find the highest version within "major" or "minor"')
+    .option('--pre', 'Include prereleases in the final result');
```
raineorshine: Add a corresponding option to the README under Advanced Options. I may modify the description to make it as clear as possible.

raineorshine:
It would be great to have unit tests for this functionality, but I think it would require mocking npm which is a little more work. At the very least we should note somewhere the missing code coverage, perhaps in a simple TODO comment.

bin/ncu
karanjthakkar: Done.

lib/package-managers/npm.js
karanjthakkar: Ahhh, got it. Fixed.

karanjthakkar:
@raineorshine I've completed the functionality for the `semverLevel` filters as well. I think it can be refactored a bit by moving the filtering of prerelease versions to the `version-util`. 

Also, as for testing, if there were a way to mock npm I would be happy to write tests for this. I'm not too sure how to go about doing that yet.

raineorshine:
Sure! Do whatever you are comfortable with! Great to have your help!

One idea for mocking npm is to provide an optional input of `npm` to each of the package manager methods. If this is given, you can use that instead of the real npm. Then you would just have to create a mock npm module with a view method that emulates the real npm, but returns fixed results for testing. 

When I get done traveling later this month, I will be able to do a more thorough review or contribute to what you've done.

lib/package-managers/npm.js
```diff
@@ -107,25 +107,67 @@ module.exports = {
             npm.commands.listAsync(args || [], true); // silent:true
     },
 
-    latest: cint.partialAt(view, 1, 'dist-tags.latest'),
+    latest: function (packageName, currentVersion, includePrerelease) {
+        return view(packageName, 'dist-tags.latest')
+            .then(function (version) {
+                if (versionUtil.getPrecision(version) === 'release' &&
```
raineorshine: Almost! If getPrecision is not `'release'`, we always want to return the dist-tags version.

karanjthakkar: Sorry! That was very silly of me. I've updated the check for this.

karanjthakkar:
I moved the filtering of release versions to `version-util` :) Let me know if I've missed something out.

lib/package-managers/npm.js
```diff
@@ -107,25 +107,60 @@ module.exports = {
...
     },
 
-    latest: cint.partialAt(view, 1, 'dist-tags.latest'),
+    latest: function (packageName, currentVersion, includePrerelease) {
+        return view(packageName, 'dist-tags.latest')
+            .then(function (version) {
+                if (
+                    versionUtil.getPrecision(version) !== 'release' ||
+                    (
+                        versionUtil.getPrecision(version) === 'release' &&
```
raineorshine: You can omit line 116. It is guaranteed to be true here since you've already checked if it's not "prerelease" in the other branch of the OR.

lib/version-util.js
```diff
@@ -179,6 +179,16 @@ function findGreatestByLevel(versions, current, level) {
         .value();
 }
 
+/**
+ * @param versions  Array of all available versions
+ * @Returns         An array of versions with the release versions filtered out
+ */
```
raineorshine: Great, this is a perfect helper function to reuse this functionality. However, I don't think it belongs in a separate file, since 1) it is only used locally within npm.js, and 2) the overhead of increasing the surface area of the version-util API doesn't justify the benefit in this case. It would be great to move it into npm.js directly. Hope that makes sense!

lib/package-managers/npm.js
karanjthakkar: Done

lib/version-util.js
karanjthakkar: Done. But now after moving it to npm.js, I'm not sure how to write tests for it since it will local and not exported via anything.

lib/npm-check-updates.js
```diff
@@ -121,7 +121,8 @@ function upgradePackageDefinitions(currentDependencies) {
 
     return vm.getLatestVersions(currentDependencies, {
         versionTarget: versionTarget,
-        registry: options.registry ? options.registry : null
+        registry: options.registry ? options.registry : null,
+        includePrerelease: options.pre
```
raineorshine: Didn't notice this before, but let's stick to using the same name for command line options and internal options. It's confusing to suddenly switch from `pre` to `includePrelease`. The README and binary provide enough explanation of what `pre` does for it to be understandable deeper in the code.

lib/package-managers/npm.js
```diff
@@ -107,25 +117,57 @@ module.exports = {
...
+    },
 
-    newest: function (packageName) {
+    newest: function (packageName, currentVersion, includePrerelease) {
         return view(packageName, 'time')
             .then(_.keys)
-            .then(_.last);
+            .then(function (versions) {
+                if (!includePrerelease) {
+                    versions = filterOutPrereleaseVersions(versions);
```
raineorshine: This is bad practice to overwrite a local variable with a different value. Could you combine these into a single return statement?

```
return _last(includePrelease ? versions : filterOutPrereleaseVersions(versions));
```

lib/package-managers/npm.js
```diff
@@ -107,25 +117,57 @@ module.exports = {
...
 
-    greatest: function (packageName) {
-        return view(packageName, 'versions').then(_.last);
+    greatest: function (packageName, currentVersion, includePrerelease) {
+        return view(packageName, 'versions')
+            .then(function (versions) {
+                if (!includePrerelease) {
+                    versions = filterOutPrereleaseVersions(versions);
+                }
+                return _.last(versions);
```
raineorshine: Same as suggestion in `newest`.

lib/package-managers/npm.js
```diff
@@ -107,25 +117,57 @@ module.exports = {
...
+            });
     },
 
-    greatestMajor: function (packageName, currentVersion) {
+    greatestMajor: function (packageName, currentVersion, includePrerelease) {
         return view(packageName, 'versions').then(function (versions) {
+            if (!includePrerelease) {
+                versions = filterOutPrereleaseVersions(versions);
+            }
             return versionUtil.findGreatestByLevel(versions, currentVersion, 'major');
```
raineorshine: Since there is more logic here, you may want to not do it all on one line. Instead of overwriting an existing variable, however, declare a new variable such as `resultVersions`.

karanjthakkar:
@raineorshine Sorry for the super late update to the PR. Have been swamped at work. But I have made the changes that you requested. Let me know if this looks good :)

lib/package-managers/npm.js
```diff
@@ -107,26 +117,54 @@ module.exports = {
...
+            });
     },
 
-    greatestMajor: function (packageName, currentVersion) {
+    greatestMajor: function (packageName, currentVersion, pre) {
         return view(packageName, 'versions').then(function (versions) {
-            return versionUtil.findGreatestByLevel(versions, currentVersion, 'major');
+            var resultVersions = versions;
+            if (!pre) {
+                resultVersions = filterOutPrereleaseVersions(versions);
```
raineorshine: Minor suggestion: A ternary statement would be much clearer here:

```
var resultVersions = pre ? versions : filterOutPrereleaseVersions(versions);
```

karanjthakkar:
@raineorshine Done. :)

PR #314, opened 2017-01-23
[WIP] Remove version ranges from final version based on flag

### LINKED ISSUE #313: Feature please make an attribute that removes the accents in version
when i run ncu on a project the output i get is:

 body-parser  ^1.15.2  →  ^1.16.0
 eslint       ^3.12.2  →  ^3.13.1

this is because on npm install the dependency was added with the accent (^). 

It would be nice to automatically remove these accents and use hard versions instead, based on an attribute say: --remove-accents.

result would be:

 body-parser  ^1.15.2  →  1.16.0
 eslint       ^3.12.2  →  3.13.1

karanjthakkar:
Fixes #313.

@raineorshine I'm not sure if this (https://github.com/tjunnone/npm-check-updates/pull/314/files#diff-15ec6f0a7089133029d1784f738d3a27R95) is the right place to skip the range addition to the final version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ Do not use these unless you know what you are doing! Not needed for typical usag
                              latest stable versions
     -a, --upgradeAll         include even those dependencies whose latest
                              version satisfies the declared semver dependency
+    --removeRange            remove version ranges from the final package version
 
 Integration
 --------------
--- a/bin/ncu
+++ b/bin/ncu
@@ -35,7 +35,8 @@ program
     .option('-u, --upgrade', 'overwrite package file')
     .option('-x, --reject <matches>', 'exclude packages matching the given string, comma-delimited list, or regex')
     .option('-a, --upgradeAll', 'include even those dependencies whose latest version satisfies the declared semver dependency')
-    .option('--semverLevel <level>', 'find the highest version within "major" or "minor"');
+    .option('--semverLevel <level>', 'find the highest version within "major" or "minor"')
+    .option('--removeRange', 'remove version ranges from the final package version');
 
 program.parse(process.argv);
 
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -124,7 +124,9 @@ function upgradePackageDefinitions(currentDependencies) {
         registry: options.registry ? options.registry : null
     }).then(function (latestVersions) {
 
-        var upgradedDependencies = vm.upgradeDependencies(currentDependencies, latestVersions);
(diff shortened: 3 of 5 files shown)


raineorshine:
Thanks! I will take a closer look as soon as I have a chance.

raineorshine:
Here are my thoughts:

1. You should check `removeRange` as close to the beginning of `upgradeDependencyDeclaration` as possible. Most of the operations in that function are irrelevant if we are returning the latest version directly.

2. We still have to put it after [the section that parses the latest version](https://github.com/karanjthakkar/npm-check-updates/blob/7daf922c7829e92ff535cc4cffc3c1c875e0933b/lib/versionmanager.js#L44-L49), because we don't want to return it if it's invalid.

3. I think it makes the most sense for `--removeRange` to take precedence over [preserving lone wildcards](https://github.com/karanjthakkar/npm-check-updates/blob/7daf922c7829e92ff535cc4cffc3c1c875e0933b/lib/versionmanager.js#L39-L42
) (i.e. Normally a dependency declared just as "*" would persist, but if they specify `--removeRange` we should convert it to an explicit version) so we need to make sure the `isWildCard` check comes *after* the `removeRange` check.
```
    if (option.removeRange) {
        return latestVersion;
    }
    else if (versionUtil.isWildCard(declaration)) {
        return declaration;
    }
```

Let me know if you have any questions or if you have another opinion on any of that!

raineorshine:
Unit tests for the new functionality can go here: https://github.com/karanjthakkar/npm-check-updates/blob/7daf922c7829e92ff535cc4cffc3c1c875e0933b/test/test-versionmanager.js#L103

joelharkes:
I just ran it locally, i see it also works on the print function 👍 

image

karanjthakkar:
@raineorshine Thanks for the inputs. I had the same thought that we could return early if `removeRange` was specified. But i wasnt sure if I'd break existing functionality while doing that so I didn't do it. Now that I have a better idea of what everything does, I'll fix up my implementation.

As for handling wildcards, yes, I totally agree that wildcards should be converted to explicit versions if `removeRange` is specified.

raineorshine:
When you have unit tests, you don't have to worry about breaking things
unexpectedly—the tests will tell you! ;)

karanjthakkar:
@raineorshine Sorry again for the delay in getting this updated! Have made the appropriate changes and added a test to test for them.

raineorshine:
Very nice! Looks great!

Issue #317, opened 2017-01-27
--silent doesn't work


raineorshine:
* node version: 6.1.0 
* npm version: 3.8.6
* npm-check-updates version: 2.8.9

---------------------------

`--silent` should behave the same as `loglevel --silent`. This appears to be a regression.

karanjthakkar:
@raineorshine This seems to occur because the loglevel option defaults to `warn`. Because of that, [this line](https://github.com/tjunnone/npm-check-updates/blob/215792f0714d4f2d28f8db0e732a5a82873e1319/lib/npm-check-updates.js#L347) does not convert `options.silent` to `options.loglevel = 'silent'`. I'm not sure if this is recent regression. The commit where those changes happened are old: https://github.com/tjunnone/npm-check-updates/commit/18b3bb459fb70a31d0af48703f856146715b7287 Am I missing something or did this never work?

raineorshine:
This is indeed a regression and worked only a long time ago. (Missed due to
lack of unit test!)

raineorshine:
Fixed in `v2.10.2`

PR #321, opened 2017-02-04
Fix --silent option regression

karanjthakkar:
Possible fix for #317. 

Two points I want to ask:

1. The current fix will override `loglevel` option when the `silent` option is provided. I think this is how it should work. But not sure if the `!options.loglevel` check was given there for a specific reason. What do you think?

2. How do we test if the silent option is set correctly? Just to make sure this doesn't break in the future again.

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -346,7 +346,7 @@ function programInit() {
         .some(_.propertyOf(options));
 
     // convert silent option to loglevel silent
-    if (!options.loglevel && options.silent) {
+    if (options.silent) {
         options.loglevel = 'silent';
     }
 }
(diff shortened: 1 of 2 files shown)


raineorshine:
Thanks for working on this!!!

1. I agree, `--silent` should take precedence.

2. https://github.com/tjunnone/npm-check-updates/blob/master/test/test-ncu.js#L87

karanjthakkar:
Added a unit test to check that `--silent` works :)

raineorshine:
Looks good! Just remove the `in-progress` label when you feel it's ready to merge.

karanjthakkar:
Done! 😎

Issue #119, opened 2015-08-06
process hangs when spawned through child_process


silverwind:
I'm using [grunt-shell](https://github.com/sindresorhus/grunt-shell) to run `ncu -ua` and it seems like the process hangs indefinitely, same command works from the shell. The shell module config is:

```js
shell: {
  options: {
    stdout: true,
    stderr: true,
    failOnError: true
  },
  update: {
    command: "ncu -ua"
  }
}
```
It basically spawns a child process, so I think it shoud be reproduceable with just `child_process.exec`.
I've tried tracking it down a bit and it looks like the promise returned [here](https://github.com/tjunnone/npm-check-updates/blob/966229163e1bfe3679113dc4cd9a817f776d003c/lib/npm-check-updates.js#L265) never resolves.

metaraine:
Thanks for tracking this down! Since `process.stdin.isTTY` was falsey, it assumes that something is being piped through stdin and it is waiting for it ([as described here](http://stackoverflow.com/a/15485424/480608)).

[get-stdin-promise](/metaraine/get-stdin-promise) is as simple as:
```
var stdin = require('get-stdin')
module.exports = new Promise(function(resolve, reject) {
	stdin(function (data) {
		resolve(data)
	})
})
```

which ultimately leads to [get-stdin](https://github.com/sindresorhus/get-stdin) (all paths lead so to [sindresorhus](/sindresorhus)):

```
module.exports = function (cb) {
	var stdin = process.stdin;
	var ret = '';

	if (stdin.isTTY) {
		setImmediate(cb, '');
		return;
	}

	stdin.setEncoding('utf8');

	stdin.on('readable', function () {
		var chunk;

		while (chunk = stdin.read()) {
			ret += chunk;
		}
	});

	stdin.on('end', function () {
		cb(ret);
	});
};
```

get-stdin [invokes the callback](https://github.com/sindresorhus/get-stdin/blob/master/index.js#L22-L24) (which resolves the promise) as soon as stdin closes.

So, my question is, **how does grunt-shell handle stdin? When does it receive the `end` event?** I don't think it would be waiting if it didn't detect that stdin was open, but it's possible that I'm missing something.

Please let me know if you are able to troubleshoot further. If you can't get any further, I will investigate more myself. I'm not invested in understanding grunt-shell inside-and-out yet, so if you were able to identify how it handles stdin that would be a big help. Thanks!

silverwind:
[...]

PR #325, opened 2017-02-15
fallback to package.json when stdin closed with no content

eighty4:
I'm making this change to with #119 

The change is to support a parent process starting ncu with child_process, closing stdin without piping package.json content and having ncu fallback to search for package.json from fs.

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -378,6 +378,7 @@ function programRunLocal() {
         pkgData = Promise.resolve(options.packageData);
     } else if (options.packageFile) {
         pkgFile = options.packageFile;
+        pkgData = getPackageDataFromFile(pkgFile, pkgFileName);
     } else if (!process.stdin.isTTY) {
         print('Waiting for package data on stdin...', 'verbose');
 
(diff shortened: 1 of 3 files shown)


test/test-ncu.js
```diff
@@ -93,6 +94,17 @@ describe('npm-check-updates', function () {
                 });
         });
 
+        it('should fall back to package.json search when receiving empty content on stdin', function (done) {
+            var childProcess = child.exec('node bin/ncu', function (error, stdout) {
+                if (error) {
+                    throw new Error(error);
```
raineorshine: Throwing an error in a callback will not necessarily bubble up. You should call `done(error)` instead to explicitly fail the test.

lib/npm-check-updates.js
```diff
@@ -387,37 +388,42 @@ function programRunLocal() {
...
-        // print a message if we are using a descendant package file
-        var relPathToPackage = path.resolve(pkgFile);
-        if (relPathToPackage !== pkgFileName) {
-            print('Using ' + relPathToPackage);
-        }
-        if (!fs.existsSync(pkgFile)) {
-            programError(chalk.red(relPathToPackage + ' not found'));
-        }
+    return pkgData.then(function (_pkgData) {
+        print('wtf', _pkgData);
```
raineorshine: Remove

eighty4:
Whoops! That def shouldn't have made it through :)

raineorshine:
Thanks, looks good!

Issue #329, opened 2017-03-03
Exception  "Path must be a string" - Version 2.10.3 (Node 6)


codeuniquely:
Just got a new Laptop (Mac Book Pro)

Doing a fresh fresh install of everything - Node, Npm, Git, Sublime Text, npm-check-updates, ...

Cloned my existing project
npm install - worked OK,

wanted to run ncu to see if it had installed any newer modules (compare to the project on my old MBP)

It threw this exception ....

```
Unhandled rejection TypeError: Path must be a string. Received null
    at assertPath (path.js:7:11)
    at Object.resolve (path.js:1146:7)
    at getPackageDataFromFile (/usr/local/lib/node_modules/npm-check-updates/lib/npm-check-updates.js:417:33)
    at programRunLocal (/usr/local/lib/node_modules/npm-check-updates/lib/npm-check-updates.js:407:19)
    at programRun (/usr/local/lib/node_modules/npm-check-updates/lib/npm-check-updates.js:356:50)
    at tryCatcher (/usr/local/lib/node_modules/npm-check-updates/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/usr/local/lib/node_modules/npm-check-updates/node_modules/bluebird/js/release/promise.js:510:31)
    at Promise._settlePromise (/usr/local/lib/node_modules/npm-check-updates/node_modules/bluebird/js/release/promise.js:567:18)
    at Promise._settlePromise0 (/usr/local/lib/node_modules/npm-check-updates/node_modules/bluebird/js/release/promise.js:612:10)
    at Promise._settlePromises (/usr/local/lib/node_modules/npm-check-updates/node_modules/bluebird/js/release/promise.js:691:18)
    at Promise._fulfill (/usr/local/lib/node_modules/npm-check-updates/node_modules/bluebird/js/release/promise.js:636:18)
    at Promise._resolveCallback (/usr/local/lib/node_modules/npm-check-updates/node_modules/bluebird/js/release/promise.js:431:57)
    at Promise._settlePromiseFromHandler (/usr/local/lib/node_modules/npm-check-updates/node_modules/bluebird/js/release/promise.js:522:17)
    at Promise._settlePromise (/usr/local/lib/node_modules/npm-check-updates/node_modules/bluebird/js/release/promise.js:567:18)
[...]

PR #332, opened 2017-03-07
Fixing #329

eighty4:
Let me know if you have a better idea for the test case. I used the "/" as the cwd for a dir that won't have a package.json in it. Not portable, right?

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -345,18 +345,40 @@ function findPackage() {
 
     var pkgFileName = getPackageFileName();
 
+    /*
+
+     // if pkgFile was set, make sure it exists and read it into pkgData
+     if (pkgFile) {
+     // print a message if we are using a descendant package file
+     var relPathToPackage = path.resolve(pkgFile);
+     if (relPathToPackage !== pkgFileName) {
+     print('Using ' + relPathToPackage);
+     }
+     if (!fs.existsSync(pkgFile)) {
+     programError(chalk.red(relPathToPackage + ' not found'));
+     }
+
+     pkgData = readPackageFile(pkgFile, null, false);
+     }
+
+     // no package data!
+     if (!pkgData) {
+     }
+
+     return pkgData.then(_.partial(analyzeProjectDependencies, _, pkgFile));
+
+     */
+
     function getPackageDataFromFile(pkgFile, pkgFileName) {
         // exit if no pkgFile to read from fs
-        if (pkgFile === null) {
-            programError(chalk.red(pkgFileName + ' not found'));
-        }
-        // print a message if we are using a descendant package file
-        var relPathToPackage = path.resolve(pkgFile);
-        if (relPathToPackage !== pkgFileName) {
-            print('Using ' + relPathToPackage);
-        }
-        if (!fs.existsSync(pkgFile)) {
-            programError(chalk.red(relPathToPackage + ' not found'));
+        if (pkgFile !== null) {
+            // print a message if we are using a descendant package file
+            var relPathToPackage = path.resolve(pkgFile);
+            if (relPathToPackage !== pkgFileName) {
+                print('Using ' + relPathToPackage);
+            }
+        } else {
+            programError(chalk.red('No ' + pkgFileName) + '\n\nPlease add a ' + pkgFileName + ' to the current directory, specify the ' + chalk.blue('--packageFile') + ' or ' + chalk.blue('--packageData') + ' options, or pipe a ' + pkgFileName + ' to stdin.');
         }
 
         return readPackageFile(pkgFile);
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -346,6 +346,10 @@ function findPackage() {
     var pkgFileName = getPackageFileName();
 
     function getPackageDataFromFile(pkgFile, pkgFileName) {
+        // exit if no pkgFile to read from fs
+        if (pkgFile === null) {
+            programError(chalk.red(pkgFileName + ' not found'));
+        }
         // print a message if we are using a descendant package file
         var relPathToPackage = path.resolve(pkgFile);
         if (relPathToPackage !== pkgFileName) {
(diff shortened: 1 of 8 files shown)


lib/npm-check-updates.js
```diff
@@ -346,6 +346,10 @@ function findPackage() {
     var pkgFileName = getPackageFileName();
 
     function getPackageDataFromFile(pkgFile, pkgFileName) {
+        // exit if no pkgFile to read from fs
+        if (pkgFile === null) {
+            programError(chalk.red(pkgFileName + ' not found'));
```
raineorshine: Thanks! Could you have it print the same error that it did before?
https://github.com/tjunnone/npm-check-updates/pull/325/files#diff-1415f5139deb67a37325f1bde416bcbdL416

test/test-ncu.js
```diff
@@ -105,6 +105,16 @@ describe('npm-check-updates', function () {
             childProcess.stdin.end();
         });
 
+        it('should handle no package.json to analyze when receiving empty content on stdin', function (done) {
```
raineorshine: Please use the existing `spawn` package for invoking child processes. It provides a clean, promisified interface. Then we won't need to pull in `child_process` to do the same thing. You see how `spawn` is used elsewhere in the file. e.g. `spawn('pwd').then(...)`. Returning a promise from the `it` block automatically handles errors.

raineorshine:
I'm not sure the best way to simulate a directory without a package.json in it. Probably should either create a temp directory or use a directory in the project itself that we know does not have a package.json.

eighty4:
Using one in the project will cause ncu to scan up the directory structure
in our project until it finds one.

eighty4:
How would you feel about using the package tmp for managing the temporary dir to test against?

test/test-ncu.js
```diff
@@ -95,14 +94,21 @@ describe('npm-check-updates', function () {
...
-                    done(error);
-                }
+            spawn('node', ['bin/ncu']).then(function (stdout) {
                 stdout.toString().trim().should.match(/^Using .+package.json/);
                 done();
             });
-            childProcess.stdin.end();
+        });
+
+        it('should handle no package.json to analyze when receiving empty content on stdin', function (done) {
```
raineorshine: Sorry I didn't catch this before, but is there any reason to use `pwd` here instead of just using `__dirname`?

test/test-ncu.js
```diff
@@ -95,14 +94,21 @@ describe('npm-check-updates', function () {
         });
 
         it('should fall back to package.json search when receiving empty content on stdin', function (done) {
-            var childProcess = child.exec('node bin/ncu', function (error, stdout) {
-                if (error) {
-                    done(error);
-                }
+            spawn('node', ['bin/ncu']).then(function (stdout) {
```
raineorshine: You would want to `return spawn(...)` and remove `done` from the function block so that the mocha can handle the test resolution automatically.

test/test-ncu.js
```diff
@@ -95,14 +94,21 @@ describe('npm-check-updates', function () {
...
                 stdout.toString().trim().should.match(/^Using .+package.json/);
                 done();
             });
-            childProcess.stdin.end();
+        });
+
+        it('should handle no package.json to analyze when receiving empty content on stdin', function (done) {
+            spawn('pwd').then(function (pwd) {
```
raineorshine: Same as comment above regarding `return spawn(...)`.

raineorshine:
tmp package would be fine

eighty4:
I removed the use of '/' and also added test to paths to lint.

package.json
```diff
@@ -71,6 +71,7 @@
     "chai-string": "^1.2.0",
     "chokidar-cli": "^1.2.0",
     "eslint": "^3.4.0",
+    "eslint-plugin-jasmine": "^2.2.0",
```
raineorshine: Is this used by eslint even for mocha tests?

test/test-ncu.js
raineorshine: https://mochajs.org/#working-with-promises

raineorshine:
Thanks! I think when the suggestions from yesterday are addressed we'll be good to go!

I've appreciated your thoroughness and communication throughout the process. I'll add you add as a collaborator so you have a little more freedom. Thanks again!

package.json
eighty4: Nope, slipped in

test/test-ncu.js
eighty4: I switched back to using done() from returning promises because you have more flow control when testing errors and promises. More of a concern when test driving an api that returns promises vs using promisey APIs in tests. I switched it back to returning to be consistent.

test/test-ncu.js
eighty4: __dirname will be the dir of the current module, so for a test it would be npm-check-updates/test but we really need the cwd/top project dir. I did just come across process.cwd() as I was double checking the diff so I can make this a tad bit simpler.

test/test-ncu.js
```diff
@@ -81,31 +80,34 @@ describe('npm-check-updates', function () {
...
                 stdout.toString().trim().should.match(/^Using .+package.json/);
-                done();
             });
-            childProcess.stdin.end();
         });
 
-        it('should output json with --jsonAll', function() {
+        it('should handle no package.json to analyze when receiving empty content on stdin', function () {
+            // run from tmp dir to avoid ncu analyzing the project's package.json
+            return spawn('node', [process.cwd() + '/bin/ncu'], {cwd: tmp.dirSync().name})
```
raineorshine: I think `spawn('node', [process.cwd() + '/bin/ncu'])` is the same as `spawn('node', ['bin/ncu'])`

raineorshine:
@eighty4 I will be out of town 3/25-4/1 so let me know if you want me to merge in the changes and publish a patch before then or if we should wait until I get back. Thanks!

test/test-ncu.js
eighty4: Changing the cwd of the child process to a tmp dir will make it look for 'bin/ncu' in the tmp dir, I'm using the parent process cwd for an abs path

PR #339, opened 2017-04-17
adding --timeout option

eighty4:
In a new test I used done() from a callback to have cleaner error handling/output when an error is expected but the spawn finishes successfully. It felt dirtier having the .catch() have to check if it was the error we expect or another error.

### DIFF
--- a/bin/ncu
+++ b/bin/ncu
@@ -31,7 +31,7 @@ program
     .option('-r, --registry <url>', 'specify third-party npm registry')
     .option('-s, --silent', "don't output anything (--loglevel silent)")
     .option('-t, --greatest', 'find the highest versions available instead of the latest stable versions')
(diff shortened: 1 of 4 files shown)


bin/ncu
```diff
@@ -31,6 +31,7 @@ program
     .option('-r, --registry <url>', 'specify third-party npm registry')
     .option('-s, --silent', "don't output anything (--loglevel silent)")
     .option('-t, --greatest', 'find the highest versions available instead of the latest stable versions')
+    .option('--timeout <seconds>', 'set a global timeout in ms')
```
raineorshine: This should say `--timeout <ms>`

eighty4: Good catch. I started out using seconds. Thanks :)

PR #353, opened 2017-05-22
Object.assign -> _.assign for backwards compat.

calebsevern:
[`Object.assign` is not available in node v0.12.x](http://node.green/#ES2015-built-in-extensions-Object-static-methods), which is listed as this project's minimum required engine.


raineorshine:
Thanks! Good catch!

raineorshine:
Published in [v2.11.2](https://github.com/tjunnone/npm-check-updates/releases/tag/v2.11.2).

PR #369, opened 2017-06-25
[V3] Remove npm dependency

bastienmoulia:
Linked to https://github.com/tjunnone/npm-check-updates/issues/350

### DIFF
--- a/lib/package-managers/npm.js
+++ b/lib/package-managers/npm.js
@@ -13,7 +13,7 @@ function parseJson(result, data) {
     try {
         json = JSON.parse(result);
     } catch (err) {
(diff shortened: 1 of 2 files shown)


yvele:
Yep, having a dependency to `npm-check-updates` makes having his `npm` v3 in bin, and makes `npm scripts` using it instead of the global `npm`.. Resulting in lot of bugs

vdh:
It looks like `npmi` will also need to be removed? Otherwise its dependency to `npm` will still install the second copy.

raineorshine:
Yes, npmi needs to be removed as well.

raineorshine:
Closing as the local npm has been removed in the soon-to-be-released [v3](https://github.com/tjunnone/npm-check-updates/issues/471).

PR #374, opened 2017-07-02
chore(package): add keyword "ncu"

AndersDJohnson:
Adding keyword "ncu" to package so that hopefully it shows up in https://www.npmjs.com/search?q=ncu.


nickmccurdy:
It seems like this won't affect search results until there's a new release.

PR #382, opened 2017-08-09
Avoid bower to check dependencies versions internally

huhgawz:


raineorshine:
Looks good, thank you!

raineorshine:
@huhgawz Any idea why the CI is failing?
https://travis-ci.org/tjunnone/npm-check-updates/jobs/262826611

raineorshine:
@huhgawz Does the test pass for you locally? If we can figure out the CI issue we can merge. Thanks!

raineorshine:
I added the unit tests to ncu v3. Thank you!

I am passing on `offline: true` until I understand better why that would be beneficial. It seems like you would want bower to connect to the network under most circumstances to check for latest versions.

PR #399, opened 2017-10-09
Upgrade global packages through subshell

JamieMason:
Hopefully this is a problem other people share and this PR will be useful.

Given a machine in this state

```
$ ncu -g
 image-to-ascii-cli   1.0.7  →   1.0.8
 jest-codemods       0.13.3  →  0.13.4
 js-beautify         1.6.14  →   1.7.4
 lebab                2.7.5  →   2.7.7
 npm-check-updates   2.12.1  →  2.13.0
 nsp                  2.7.0  →   2.8.1
 prepack              0.2.6  →   0.2.8
 prettier             1.6.1  →   1.7.4
 rimraf               2.6.1  →   2.6.2
```

When running `ncu -ug` the following output is displayed

```
ncu cannot upgrade global packages. Run npm install -g [package] to update a global package
```

This PR would result in the following output being displayed when running `ncu -ug` and upgrades are available

```
ncu cannot upgrade global packages. Run npm install -g image-to-ascii-cli jest-codemods js-beautify lebab npm-check-updates nsp prepack prettier rimraf to update a global package
```

...or the original message if no upgrades are available

```
ncu cannot upgrade global packages. Run npm install -g [package] to update a global package
```

The reason for this request is that it would let me copy the install message and run it to install all global dependencies at once. At the moment I have to run `ncu -g` then copy and edit the output to construct this install command.

The linter and tests are passing.

Thanks a lot for this really useful tool.

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -127,11 +127,6 @@ var writePackageFile = fs.writeFileAsync;
 
 function analyzeGlobalPackages(options) {
 
-    if (options.global && options.upgrade) {
-        programError(options, chalk.cyan('ncu') + ' cannot upgrade global packages. Run ' + chalk.cyan('npm install -g [package]') +
-            ' to update a global package');
-    }
-
     print(options, 'Getting installed packages...', 'verbose');
 
     return vm.getInstalledPackages({
@@ -145,6 +140,12 @@ function analyzeGlobalPackages(options) {
 
             return vm.upgradePackageDefinitions(globalPackages, options)
                 .spread(function (upgraded, latest) {
+
+                    if (options.global && options.upgrade) {
+                        var instruction = Object.keys(upgraded).join(' ') || '[package]';
+                        programError(options, chalk.cyan('ncu') + ' cannot upgrade global packages. Run ' + chalk.cyan('npm install -g ' + instruction) +
+                            ' to update a global package');
+                    }
 
                     print(options, latest, 'silly');
                     printUpgrades(options, {


raineorshine:
Thank you @JamieMason! I wonder if we could actually do one better. We could prompt the user and then execute the command in a subshell:

```js
var spawn = require('spawn-please')
spawn(process.platform === 'win32'? 'npm.cmd' : 'npm', Object.keys(upgraded))
```

Because this is irreversible, I would expand the prompt to the following:

```
Would you like to upgrade these global packages? This cannot be reversed. (y/n)
```

I'm not sure what the easiest way is to provide a prompt of y/n but that shouldn't be hard. What do you think?

JamieMason:
That sounds really cool, much more useful. I've not done prompts like that before but is something I've wanted to try anyway. 

Something that's just occurred to me that I'd forgotten about – It may be a non-issue but I've had problems in the past updating global packages where `npm` itself was one of the packages to update. Ever had that?

If `npm` itself is one of the upgradeable packages, it might be safer to run `npm install -g npm && npm install -g all the others` which I've never had a problem with, but likely the code to achieve this won't be as nice.

raineorshine:
Yes, that might be a complication. We can either ignore npm itself or
change the command as you proposed.

JamieMason:
@raineorshine What renders the `⸨░░░░░░░░░░░░░░░░░░⸩ ⠴ :` please? I have a working prompt but I think I need some way to complete this progress indicator before. Thanks.

raineorshine:
Nice! Unfortunately that's npm itself. Maybe there is a way to turn it off?

kurtextrem:
What's the status of this PR? Is there anything to help with? :)

JamieMason:
Hey @kurtextrem. I've just not found the time but I definitely plan to finish this. If you'd like to help or take it over I'd be happy to chat with you to hand it over? Otherwise I'll get to it as soon as I can, thanks a lot 😊

kurtextrem:
@JamieMason Sorry for the late reply, yes, I'd love to take this one :)

PR #403, opened 2017-11-08
Removed message asking to run ncu -a when it just run

gsiou:
After running ncu -a the message 

> The following dependencies are satisfied by their declared version range, but the installed versions are behind. You can install the latest versions without modifying your package file by using npm update. If you want to update the dependencies in your package file anyway, run ncu -a.

appears again as shown in the image below

ncu-before

I believe that this should not be the case, since I just ran ncu -a and I just want to see what has changed in my package files (not be prompted to run it again), as shown in the image below

ncu-after

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -281,7 +281,9 @@ function printUpgrades(options, args) {
             greatest: options.greatest,
             newest: options.newest
         });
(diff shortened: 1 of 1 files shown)


raineorshine:
Makes sense. Thanks!

PR #413, opened 2017-12-22
fix typo in README

tw0517tw:
name and path, not path and path, I guess?

cc @cavinsmith


cavinsmith:
@tw0517tw sure! thank you!

PR #416, opened 2018-01-18
update to latest

gemal:
With this:
```
nsp check
(+) No known vulnerabilities found
```


raineorshine:
Thanks! There are some complications with this, and I'm not sure it's particularly necessary at the moment. `find-up` has a backwards-incompatible change so it should not be upgraded. I haven't checked the other ones. I might just hold off on this until it's needed.

KurtPreston:
@raineorshine I can't vouch for all the updates, but the update to a newer version of `npmi` would be really helpful.  The version `npm-check-updates` currently references loads npm v3, which causes downstream errors on `npm install` for us.

raineorshine:
`npmi` needs to be removed completely so that `npm-check-updates` simply uses the system `npm`. This is slated for `v3` but I haven't had time to do this. There are currently no other developers working on `npm-check-updates` so sadly it hasn't received much love.

KurtPreston:
@raineorshine Getting off of `npmi` would be nice, but if it's low effort to just bump the version it, I think it'd be a worthwhile update.  We've had to remove `node-check-updates` from our project, because the current version causes an issue requiring the `node_modules` folder to be deleted before every `npm install`.  I'm sure we're not the only ones with this issue, and bumping `npmi` would be a quick fix for this issue.

(The underlying issue is here, associated with npm v3 https://github.com/nodejs/node/issues/13492)

raineorshine:
@KurtPreston Thank you, this definitely needs to be corrected?

It is a low effort change but large impact since it requires a major version upgrade. There are other backwards-incompatible changes that need to be included in the move from v2 to v3. In the mean time, I have bumped `npm` and `npmi` in an alpha release. Maybe this will solve your issue until `v3` is officially released.

```
npm
uninstall -g npm-check-updates && npm install -g
npm-check-updates@v3-alpha
```

KurtPreston:
@raineorshine This works great, much appreciated! No error during `npm install` on the `v3-alpha` branch.  For some context, we use ncu in some gulp tasks, so have it defined as a project dependency instead of using it as a global package. If you try to add the v2 ncu to a project's `package.json`, you may encounter the same error when running `npm install`, since it brings in a buggy version of `dezalgo` as a dependency.

Thanks for your effort!  ncu's a super-useful project.

raineorshine:
Closing as all dependencies will be upgraded in the soon-to-be-released [v3](https://github.com/tjunnone/npm-check-updates/issues/471).

PR #444, opened 2018-06-15
:lock: Try to fix vulnerabilities

quilicicf:
Hi, 

I tried to update all the dependencies I could so that the tests still pass and the vulnerabilities get reduced to a minimum. 

Couldn't upgrade chai-as-promised more because it was breaking the tests.


quilicicf:
The tests failed with node below 6. 
I'm wondering if they should still be supported here. The current lowest active version is [Node 8](https://github.com/nodejs/Release#release-schedule).
I'd trade old versions support over security but I'll leave that to you.

raineorshine:
Thank you! It is good to have the latest versions.

The intention is to remove support for node < 6 in ncu v3, when the codebase is converted to es6. I haven't had time lately and nobody is working on it currently. That change and a couple other breaking changes are needed to get to [v3](https://github.com/tjunnone/npm-check-updates/issues?q=is%3Aopen+is%3Aissue+milestone%3A3.0.0).

raineorshine:
@quilicicf All tests pass for ncu `v2.14.2` in node `v4.9.1` FYI. This is why removing support for older versions of node is part of the move to ncu `v3`.

raineorshine:
@quilicicf Unfortunately, the tests do not pass for me with your updated dependencies. We may have to go back and see which ones can safely be upgraded. Here is my output: https://gist.github.com/raineorshine/8adb32027bb1c4229b637b9d452f24b9.

quilicicf:
I'll try to understand why the tests fail. 
Which version of Node are you using ? I'm using node 10.3.0 (npm 6.1.0).
It's strange that there are errors in the console output and no tests are marked as failed too Oo.

raineorshine:
I found it: `chai-as-promised` has a peer dependency of `chai@>= 2.1.2 < 4`. We need need to either upgrade the tests to work with the latest version of `chai-as-promised` or revert `chai` to `^3.5.0`

quilicicf:
Cool! I'll Downgrade chai for now and try to upgrade both chai and chai-as-promised when I get the time then.

quilicicf:
I've tried to fix the build for Node v5 and v4 and failed miserably. 
This seems to be an endless stream of downgrading to bump into other issues. 

I can give a hand to make that move to v3 if that helps but I'll probably need some assistance as I don't know the code base at all. 

> Note: the code migration to ES6 is not required to bump to v3 and drop support of older versions of Node but that's the area where I'm most likely to be of help as it does not require any knowledge about the code base

raineorshine:
@quilicicf I think the move to `v3` is the best way to move forward. I'd be happy to help out with any questions you have about the codebase. Main issues are converting to es6 syntax and using the system `npm` instead of the dependency `npm`. Then there are just a few smaller fixes that nevertheless are breaking, so it would be nice to include them in this release.

https://github.com/tjunnone/npm-check-updates/issues?q=is%3Aopen+is%3Aissue+milestone%3A3.0.0

raineorshine:
Accidentally merged... manually reset.

PR #446, opened 2018-06-29
chore: use npm ci

DanielRuf:


raineorshine:
Thanks! Although I have some qualms about this change. This forces a global install of the latest version of npm. What if users do not want to upgrade their `npm`? This seems outside the purview of `ncu`.

DanielRuf:
> Thanks! Although I have some qualms about this change. This forces a global install of the latest version of npm. What if users do not want to upgrade their npm? This seems outside the purview of ncu

This just affects GitLab, not normal users.

DanielRuf:
> Thanks! Although I have some qualms about this change. This forces a global install of the latest version of npm. What if users do not want to upgrade their npm? This seems outside the purview of ncu

This just affects Travis CI, not normal users.

raineorshine:
Ah, my bad. Should have noticed it was the .travis.yml file.

I am still confused about the purpose and safety of this change. If there is a before_install script that updates npm to latest, doesn't that defeat the purpose of travis testing across multiple versions of node and npm?

raineorshine:
Closing but happy to reopen if a reason is provided! Thanks!

PR #461, opened 2018-11-13
Update README.md -> docker -> don't wait for stdin

srghma:
don't wait for stdin because this allows to run docker commands in parallel


raineorshine:
Thanks!

PR #462, opened 2018-11-16
Add upgrade instruction for global execution

kurtextrem:
Similar to #399 this adds an upgrade command below the upgrade-able packages.

Whenever someone executes `ncu -n -g` that someone is definitely interested in upgrading packages. Why else would you run this command, so it should always give you an easy way to upgrade.
However, I did not add a prompt, as I think you don't always want to upgrade every package. This way someone can copy & paste the command, remove the package that should stay like it is and move on.

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -149,14 +149,15 @@ function analyzeGlobalPackages(options) {
 
                     if (options.global) {
                     	var instruction = '[package]';
-                    	if (upgraded)
+                    	if (upgraded) {
                         	instruction = Object.keys(upgraded).map(function (package) {
-                        		return package + '@' + upgraded[package]
+                        		return package + '@' + upgraded[package];
                         	}).join(' ');
+                        }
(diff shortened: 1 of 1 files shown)


raineorshine:
Thanks for the PR! Great improvement!

The only thing that needs to be changed is that the promise returned by `printUpgrades` still needs to be returned ([L150](https://github.com/tjunnone/npm-check-updates/pull/462/files#diff-1415f5139deb67a37325f1bde416bcbdL150)) due to the possibility of `writePackageFile` being called ([L299](https://github.com/tjunnone/npm-check-updates/pull/462/files#diff-1415f5139deb67a37325f1bde416bcbdR299)).

kurtextrem:
Thank you! Do you have any idea how to write a message after `printUpgrades` then?

raineorshine:
@kurtextrem I would suggest storing the result of `printUpgrades` in a variable and then returning it at the end.

```
var result = printUpgrades(...)

...

return result
```

kurtextrem:
@raineorshine Great idea, will do that.

raineorshine:
Thanks! I tweaked the wording but otherwise it's great. Published in `v2.15.0`.

PR #467, opened 2018-12-21
Update deps and fix some tab/space style

claudyus:
Update the deps and minor style fix


raineorshine:
Thank you! I think we shouldn't deprecate node versions without a major version upgrade. I will mark this for inclusion in v3. Unfortunately there aren't many active contributors to ncu right now so I am not sure when this will happen.

kibertoad:
@raineorshine Are you searching for new contributors at this point?

raineorshine:
@kibertoad Very much! Could really use the help.

kibertoad:
@raineorshine I'm up for it. Anything you could point out as a priority?

raineorshine:
@kibertoad Yay! Head on over to #471 for priorities and watch for what others are working on to avoid overlap. Thank you!

PR #472, opened 2019-01-20
V3 non-intrusive dep upgrade

kibertoad:
This is less intrusive version of https://github.com/tjunnone/npm-check-updates/pull/467 that could be merged without changing any code.
Note that it raises minimum required Node version to 4 due to Chai dependency upgrade (older version breaks build due to in-range peer dependency change).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,36 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


raineorshine:
@kibertoad In ncu v3 I am planning on targeting node >= 8 given the [node release schedule](https://github.com/nodejs/Release).

PR #487, opened 2019-03-07
Typo in README.md

MoritzHorch:
Pretty obious typo, but still cost me some minutes to figure out. Anyway, thanks for this module. Just using it since several minutes but finding it pretty useful already.


raineorshine:
Thank you!

PR #497, opened 2019-03-11
Replace npm cli calls with package-json

### LINKED ISSUE #488: Crashes while fetching latest versions (Windows)
As reported by @mhosman:

> The console is crashing running ncu and also running ncu --loglevel verbose --packageFile package.json. The console gets stuck in "Fetching latest versions..." and after a while, crashes (this behaviour is only available with 3.0.2 version).

### LINKED ISSUE #496: ncu command gets stuck (3.0.4)

* node version: v10.15.0
* npm version: 6.8.0
* npm-check-updates version: 3.0.4
* OS: OSX 10.14.1

---------------------------

Sometimes `ncu` command gets stuck forever and ever. It seems to happen always in the same project directories. It does not happen in others. I'm trying to get a patron but could just get a repo in which `ncu` always gets stuck. I don't see any option in `ncu` command to do debugging or whatever so not sure why it's failing.

### Steps to reproduce

```bash
$ git clone git@github.com:versatica/mediasoup-client.git

$ npm i

$ time ncu
// => never ends
```

NOTE: Eventually I got it working, but it took SO long. It was just once.

stoically:
~~Added my [custom branch](https://github.com/stoically/package-json/tree/keepalive) of `package-json` to the package.json until https://github.com/sindresorhus/package-json/pull/50 gets merged. Otherwise huge dependencies still take minutes compared to seconds.~~ PR got merged and released as `package-json` 6.1.0.

Fixes #488, Fixes #496

### DIFF
--- a/lib/package-managers/npm.js
+++ b/lib/package-managers/npm.js
@@ -3,6 +3,7 @@ const _ = require('lodash');
 const cint = require('cint');
 const versionUtil  = require('../version-util.js');
 const spawn        = require('spawn-please');
+const packageJson  = require('package-json');
 
 /** Parse JSON and throw an informative error on failure.
  * @param result Data to be parsed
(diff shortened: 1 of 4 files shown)


lib/package-managers/npm.js
```diff
@@ -21,19 +22,27 @@ function parseJson(result, data) {
...
+    return packageJson(packageName, options).then(result => {
+        if (field.startsWith('dist-tags.')) {
+            const split = field.split('.');
+            return result[split[0]][split[1]];
+        } else if (field === 'versions') {
+            return Object.keys(result[field]);
+        } else {
+            return result[field];
+        }
+    }).catch(error => {
```
raineorshine: Unless I am missing something, we should remove the `catch` and allow errors to bubble up here. If an error needs to be ignored, we should match the text of the error message. That said, it should probably still throw and allow it to be caught higher up if necessary.

stoically: Thanks for the fast review! I've added the catch because there are [two tests](https://github.com/tjunnone/npm-check-updates/blob/49e258a6c96b497aec6fb0796a54f6b69f81ff0d/test/test-versionmanager.js#L420-L428) which specify that unknown packages should be ignored, thus returning `false`.

Though I realized that using `console.error` would mess with `--jsonAll` and `--silent` - I've tweaked it to check for `package-json`'s exported `PackageNotFoundError` and throw in any other case.

raineorshine: @stoically FYI I decided to leverage the existing 404 handling: https://github.com/tjunnone/npm-check-updates/commit/529bebad3e4e1e2f26c328d15dc5597f41745995

Thanks so much for making this happen! A huge improvement.

PR #501, opened 2019-03-13
Error handling

### LINKED ISSUE #500: Unhandled rejection for local file:dependency

* node version: 10.15.1
* npm version: 6.4.1
* npm-check-updates version: 3.0.5

---------------------------

I have a local dependency in my package.json that looks like this:

```js
{
  "devDependencies": {
    ...
    "eslint-plugin-internal": "file:devtools/eslint-rules",
    ...
  }
}
```

Since upgrading to v3.0.5, I get this error when running `ncu`:

```
Unhandled rejection PackageNotFoundError: Package `eslint-plugin-internal` could not be found
    at packageJson (/usr/lib/node_modules/npm-check-updates/node_modules/package-json/index.js:70:10)
    at processTicksAndRejections (internal/process/next_tick.js:81:5)
```

stoically:
Check for `package-json.PackageNotFoundError` rejection. Also make travis fail if the tests are failing.

Fixes #500

### DIFF
--- a/lib/versionmanager.js
+++ b/lib/versionmanager.js
@@ -5,6 +5,7 @@ const cint = require('cint');
 const semverutils = require('semver-utils');
 const Promise = require('bluebird');
 const ProgressBar = require('progress');
+const packageJson = require('package-json');
 const versionUtil = require('./version-util.js');
 const packageManagers = require('./package-managers');
 
@@ -324,7 +325,7 @@ function queryVersions(packageMap, options = {}) {
 
     let bar;
     if (!options.json && options.loglevel !== 'silent') {
-        bar = new ProgressBar('[:bar] :percent', {total: packageList.length*2+1, width: 20});
+        bar = new ProgressBar('[:bar] :percent', {total: packageList.length+1, width: 20});
         bar.tick();
     }
 
@@ -358,11 +359,9 @@ function queryVersions(packageMap, options = {}) {
 
     // ignore 404 errors from getPackageVersion by having them return null instead of rejecting
     function getPackageVersionProtected(dep) {
-        if (bar) {
-            bar.tick();
-        }
         return getPackageVersion(dep, packageMap[dep], options.pre).catch(err => {
-            if (err && (err.message || err).toString().match(/E404|404 Not Found|doesn't exist/i)) {
+            if (err && (err.message || err).toString().match(/E404|404 Not Found|doesn't exist/i) ||
+                err instanceof packageJson.PackageNotFoundError) {
                 return null;
             } else {
                 throw err;
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   "scripts": {
     "lint": "eslint bin/* lib test",
     "watch": "chokidar \"lib/**/*.js\" -c \"npm run test\"",
-    "test": "npm run lint && mocha && mocha test/individual; if [ \"$TRAVIS_PULL_REQUEST\" = \"false\" ]; then snyk test; fi"
+    "test": "npm run lint && mocha && mocha test/individual && if [ \"$TRAVIS_PULL_REQUEST\" = \"false\" ]; then snyk test; fi"
   },
   "bin": {
     "npm-check-updates": "./bin/npm-check-updates",


lib/versionmanager.js
```diff
@@ -358,19 +359,17 @@ function queryVersions(packageMap, options = {}) {
...
     // ignore 404 errors from getPackageVersion by having them return null instead of rejecting
     function getPackageVersionProtected(dep) {
-        if (bar) {
-            bar.tick();
-        }
+        bar.tick();
         return getPackageVersion(dep, packageMap[dep], options.pre).catch(err => {
-            if (err && (err.message || err).toString().match(/E404|404 Not Found|doesn't exist/i)) {
+            if (err && (err.message || err).toString().match(/E404|404 Not Found|doesn't exist/i) ||
+                err instanceof packageJson.PackageNotFoundError) {
```
raineorshine: On the previous PR I was getting an runtime error trying to access PackageNotFoundError, which was why I removed it, but clearly I was doing something wrong, as this works now. Thank you!

lib/versionmanager.js
```diff
@@ -358,19 +359,17 @@ function queryVersions(packageMap, options = {}) {
 
     // ignore 404 errors from getPackageVersion by having them return null instead of rejecting
     function getPackageVersionProtected(dep) {
-        if (bar) {
-            bar.tick();
-        }
+        bar.tick();
```
raineorshine: This is the tick that I am questioning. Does this mark any significant time? Or should we just wait for the `getPackageVersion` function to return?

lib/versionmanager.js
```diff
@@ -358,19 +359,17 @@ function queryVersions(packageMap, options = {}) {
...
         return getPackageVersion(dep, packageMap[dep], options.pre).catch(err => {
-            if (err && (err.message || err).toString().match(/E404|404 Not Found|doesn't exist/i)) {
+            if (err && (err.message || err).toString().match(/E404|404 Not Found|doesn't exist/i) ||
+                err instanceof packageJson.PackageNotFoundError) {
+                bar.tick();
```
raineorshine: You can remove this tick regardless, because after the error is caught it will flow into the next `then` block.

lib/versionmanager.js
```diff
@@ -322,7 +323,7 @@ function queryVersions(packageMap, options = {}) {
     // validate options.versionTarget
     options.versionTarget = options.versionTarget || 'latest';
 
-    let bar;
+    let bar = {tick: () => {}};
     if (!options.json && options.loglevel !== 'silent') {
         bar = new ProgressBar('[:bar] :percent', {total: packageList.length*2+1, width: 20});
```
raineorshine: Can you explain the reasoning behind setting the total to `packageList.length*2+1` instead of just `packageList.length`? With the former, some ticks will be longer than others. With the latter, there will be perhaps an initial pause, but then each tick will be a proportional percentage of the total. We could do a 1 or more ticks for the initial pause, i.e. n + c instead of 2n + 1.

lib/versionmanager.js
raineorshine: I must have somehow had `package-json` v6.0.0 before [errors were exported](https://github.com/sindresorhus/package-json/commit/685fef238e6c5c4e053fda219e96cb42fae0e109).

lib/versionmanager.js
stoically: Just wanted the progress bar to do something while waiting for the first `package-json` response to finish. But I'm also fine with only tick'ing when responses come in - changed it accordingly. Though, the response ticks are not necessarily a proportional percentage either (in terms of time) since responses can vary in their completion-time. The first tick is so the bar is actually shown.

raineorshine: I guess I am unsure what you mean by "first" `package-json` finish. Is `package-json` called more than once for each module? Unless I am missing something I am only seeing the call in `queryVersions`. Thanks!

stoically: Sorry, what I actually meant is the [first resolving of `getPackageVersion`](https://github.com/tjunnone/npm-check-updates/blob/c96fc0eca5737c8540a7fcbf9359f1e1e9312752/lib/versionmanager.js#L370) (though, the most time inside `getPackageVersion` will probably be spent waiting for the `package-json` response). I'm saying "first" because `queryVersions` [`map`s](https://github.com/tjunnone/npm-check-updates/blob/c96fc0eca5737c8540a7fcbf9359f1e1e9312752/lib/versionmanager.js#L385) the calls to `getPackageVersion` and one of the promises will be the first one to resolve.

With the [`bar.tick` inside the `getPackageVersionProtected`](https://github.com/tjunnone/npm-check-updates/blob/c96fc0eca5737c8540a7fcbf9359f1e1e9312752/lib/versionmanager.js#L362) every `map` call lets the bar progress for every package - and so the bar wouldn't sit at 0% until the first `getPackageVersion` call resolves.

Issue #502, opened 2019-03-13
Implement --interactive flag to ask each dependency "do you want to upgrade?"


mkungla:
Ref #95, why could not these be merged https://github.com/th0r/npm-upgrade

- [x ] I have read the list of [known issues](https://github.com/tjunnone/npm-check-updates#known-issues) before filing this issue.
- [x ] I have searched for [similiar issues](https://github.com/tjunnone/npm-check-updates/issues) before filing this issue.

* node version:  *
* npm version: *
* npm-check-updates version: *

---------------------------

raineorshine:
That would be a great addition! I'm happy to review a PR for this.

PR #505, opened 2019-03-14
Add --interactive flag quick implementation

mkungla:
ref #502 Not familiar with that project so I just made a quick solution for it.
There may be better solution for implementing this feature.

### DIFF
--- a/lib/versionmanager.js
+++ b/lib/versionmanager.js
@@ -245,18 +245,19 @@ function upgradePackageDefinitions(currentDependencies, options) {
 async function upgradePackageData(pkgData, oldDependencies, newDependencies, newVersions, options = {}) {
     for (const dependency in newDependencies) {
         if (!options.minimal || !isSatisfied(newVersions[dependency], oldDependencies[dependency])) {
-            if (options.upgrade && options.interactive) {
+            if (options.interactive) {
                 const to = versionUtil.colorizeDiff(newDependencies[dependency] || '', oldDependencies[dependency]);
-                const q = {
+                const response = await prompts({
                     type: 'confirm',
                     name: 'value',
                     message: `Do you want to upgrade: ${dependency} ${oldDependencies[dependency]} → ${to}?`,
                     initial: true
-                };
-                let response = await prompts(q);
+                });
                 if (!response.value) {
+                    // Continue loop to next dependency and skip updating newPkgData
+                    // delete newDependencies[dependency] so that this dependency will be omitted from
+                    // printUpgrades(..., {..., upgraded, ..., ..., ..., ...})
                     delete newDependencies[dependency];
-                    delete oldDependencies[dependency];
                     continue;
                 }
             }
--- a/lib/versionmanager.js
+++ b/lib/versionmanager.js
@@ -8,6 +8,7 @@ const ProgressBar = require('progress');
 const packageJson = require('package-json');
 const versionUtil = require('./version-util.js');
 const packageManagers = require('./package-managers');
+const prompts = require('prompts');
 
 // keep order for setPrecision
 const DEFAULT_WILDCARD = '^';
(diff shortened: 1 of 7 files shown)


lib/versionmanager.js
```diff
@@ -241,12 +242,28 @@ function upgradePackageDefinitions(currentDependencies, options) {
  * @param {Object} [options={}]
  * @returns {string} The updated package data, as utf8 text
  */
-function upgradePackageData(pkgData, oldDependencies, newDependencies, newVersions, options = {}) {
+async function upgradePackageData(pkgData, oldDependencies, newDependencies, newVersions, options = {}) {
     for (const dependency in newDependencies) {
         if (!options.minimal || !isSatisfied(newVersions[dependency], oldDependencies[dependency])) {
+            if (options.upgrade && options.interactive) {
```
raineorshine: Let's remove the `options.upgrade` requirement. It will still work with it but it can be treated independently.

lib/versionmanager.js
```diff
@@ -241,12 +242,28 @@ function upgradePackageDefinitions(currentDependencies, options) {
...
+                const to = versionUtil.colorizeDiff(newDependencies[dependency] || '', oldDependencies[dependency]);
+                const q = {
+                    type: 'confirm',
+                    name: 'value',
+                    message: `Do you want to upgrade: ${dependency} ${oldDependencies[dependency]} → ${to}?`,
+                    initial: true
+                };
+                let response = await prompts(q);
```
raineorshine: `const response`

lib/versionmanager.js
```diff
@@ -241,12 +242,28 @@ function upgradePackageDefinitions(currentDependencies, options) {
  * @param {Object} [options={}]
  * @returns {string} The updated package data, as utf8 text
  */
-function upgradePackageData(pkgData, oldDependencies, newDependencies, newVersions, options = {}) {
+async function upgradePackageData(pkgData, oldDependencies, newDependencies, newVersions, options = {}) {
     for (const dependency in newDependencies) {
         if (!options.minimal || !isSatisfied(newVersions[dependency], oldDependencies[dependency])) {
+            if (options.upgrade && options.interactive) {
+                const to = versionUtil.colorizeDiff(newDependencies[dependency] || '', oldDependencies[dependency]);
+                const q = {
```
raineorshine: Let's pass the arguments object inline instead of a separate `q` variable.

lib/versionmanager.js
```diff
@@ -241,12 +242,28 @@ function upgradePackageDefinitions(currentDependencies, options) {
...
+                    type: 'confirm',
+                    name: 'value',
+                    message: `Do you want to upgrade: ${dependency} ${oldDependencies[dependency]} → ${to}?`,
+                    initial: true
+                };
+                let response = await prompts(q);
+                if (!response.value) {
+                    delete newDependencies[dependency];
```
raineorshine: I am hesitant to mutate `oldDependencies` and `newDependencies` (mutations = bad). Is there anything downstream that depends on these being mutated here? If so, would it be possible to copy them, mutate the copy, and then pass them down instead?

lib/versionmanager.js
```diff
@@ -241,12 +242,28 @@ function upgradePackageDefinitions(currentDependencies, options) {
...
+                    delete oldDependencies[dependency];
+                    continue;
+                }
+            }
             const expression = `"${dependency}"\\s*:\\s*"${escapeRegexp(`${oldDependencies[dependency]}"`)}`;
             const regExp = new RegExp(expression, 'g');
             pkgData = pkgData.replace(regExp, `"${dependency}": "${newDependencies[dependency]}"`);
```
raineorshine: While we are at it, we could return `pkgData` instead of mutating it as well.

raineorshine:
One more small point: we will need the new `--interactive` option added to the README under Advanced Options.

lib/versionmanager.js
mkungla: mutation of oldDependencies was unnecessary, however  newDependencies as I understand is used by reference when printing updated dependency list. 

`upgraded = newDependencies`

https://github.com/tjunnone/npm-check-updates/blob/6ba3c1d1e78c29e4c38e771a681d9a2aa46d078c/lib/npm-check-updates.js#L201-L208

lib/versionmanager.js
mkungla: I Don't quite understand what you mean.

raineorshine: Nvm, I see now it's just overwriting the reference, not mutating the argument.

raineorshine:
Thanks!!

PR #511, opened 2019-03-15
Make sure pkgFile is set in non-tty environments

### LINKED ISSUE #474: Can't get status code when using --error-level
[X]  I have read the list of known issues before filing this issue.
[X] I have searched for similiar issues before filing this issue.
* node version: v10.13.0
* npm version: v6.4.1
* npm-check-updates version: 2.15.0 and 3.0.0-alpha.3

---------------------------

No idea if this is an ncu issue or not.

When I run ncu from the command line I can get the correct error code in `$?`. But when that same command line is run via teamcity the `$?` status code is always 0. 

The script snippet looks like
```
ncu --error-level 2
check_status=$?
echo "check_status=${check_status}"
```

Any clues would be much appreciated.

stoically:
Fixes #474

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -306,7 +306,7 @@ Searches as follows:
  --stdin
  --findUp
 */
-function findPackage(options) {
+async function findPackage(options) {
 
     let pkgData;
     let pkgFile;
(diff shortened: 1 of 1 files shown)


lib/npm-check-updates.js
```diff
@@ -409,7 +408,7 @@ async function run(options={}) {
 
     let pendingAnalysis = options.global ?
         analyzeGlobalPackages(options) :
-        findPackage(options).spread(_.partial(analyzeProjectDependencies, options));
+        (_.partial(analyzeProjectDependencies, options)).apply(null, await findPackage(options));
```
raineorshine: I believe this is equivalent to: `analyzeProjectDependencies(options, [pkgData, pkgFile])`, however `analyzeProjectDependencies` has the signature: `function analyzeProjectDependencies(options, pkgData, pkgFile)`.

It may not be explained well in the code, but pkgFile is optional. If pkgFile is undefined, it is because the content of the package.json is piped into ncu via stdin. The problem described in #474 is somehow related to isTTY being false (which would normally imply content on stdin) when the user is not piping anything over stdin.

raineorshine:
I did some deeper investigating and ended up fixing the issue in 4f6b4a3. Thanks for jumping in though!

lib/npm-check-updates.js
stoically: The `.apply` spreads the returned array from `findPackage` onto the _.partial'ized function and thus calls it with the correct signature.

In non-tty environments where the stdin is empty it falls back to using a package.json and so pkgFile should be set, because there's actually one used. The only reason pkgFile is not set in this case is because the [Promise.all return](https://github.com/tjunnone/npm-check-updates/blob/e1432917eec472bfbc20d0766bf1554ee2a01416/lib/npm-check-updates.js#L376) is called (with undefined pkgFile) before the [stdin promise](https://github.com/tjunnone/npm-check-updates/blob/e1432917eec472bfbc20d0766bf1554ee2a01416/lib/npm-check-updates.js#L359) resolves and thus [setting pkgFile](https://github.com/tjunnone/npm-check-updates/blob/e1432917eec472bfbc20d0766bf1554ee2a01416/lib/npm-check-updates.js#L364) too late in this case.

raineorshine: Ah yes. I neglected the apply.

Aha! That's the timing issue! I will put together a new PR for that and get your feedback.

PR #513, opened 2019-03-16
Fix timing issue in non-tty findPackage

raineorshine:
@stoically Would love a quick review!

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -171,7 +171,11 @@ function analyzeProjectDependencies(options, pkgData, pkgFile) {
     let pkg;
 
     try {
-        pkg = jph.parse(pkgData);
+        if (!pkgData) {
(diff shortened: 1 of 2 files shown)


lib/npm-check-updates.js
```diff
@@ -355,18 +361,18 @@ function findPackage(options) {
             console.log(stdinWarningMessage);
         }, stdinWarningTime);
 
-        // clear the warning timer once stdin returns and fallback to scanning pwd if no content from stdin
-        pkgData = getstdin().then(_pkgData => {
+        // get data from stdin
+        // clear the warning timer once stdin returns
+        const stdinData = getstdin().then(data => {
```
stoically: Just a minor suggestion: since `findPackage` is now an async function, it would be possible to await the `getstdin` here and with that the `stdinData.then` calls in line 374 & 375 would not be necessary. Besides that it looks good to me!

PR #525, opened 2019-03-27
Add NO_PROXY support

stoically:
Fixes https://github.com/tjunnone/npm-check-updates/issues/521#issuecomment-477090315

Also added some proxy tests.

### DIFF
--- a/lib/package-managers/npm.js
+++ b/lib/package-managers/npm.js
@@ -6,25 +6,8 @@ const Promise = require('bluebird');
 const versionUtil  = require('../version-util.js');
 const spawn        = require('spawn-please');
 const packageJson  = require('package-json');
-const ProxyAgent   = require('https-proxy-agent');
-const npmConf      = require('npm-conf')();
-const proxyUrl     = require('get-proxy')();
 
-let agent;
-if (proxyUrl) {
-    const proxy = new URL(proxyUrl);
-    agent = new ProxyAgent({
-        host: proxy.hostname,
-        port: proxy.port,
-        secureProxy: proxy.protocol === 'https:',
-        keepAlive: true,
-        maxSockets: npmConf.get('maxsockets') || 50, // https://docs.npmjs.com/misc/config#maxsockets
-        ca: npmConf.get('ca'),
-        key: npmConf.get('key'),
-        cert: npmConf.get('cert'),
-        rejectUnauthorized: npmConf.get('strict-ssl')
-    });
-}
+const {getProxyAgent} = require('./npm-proxy');
 
 /** Parse JSON and throw an informative error on failure.
  * @param result Data to be parsed
@@ -50,10 +33,18 @@ function view(packageName, field, currentVersion) {
     if (currentVersion && (!semver.validRange(currentVersion) || versionUtil.isWildCard(currentVersion))) {
         return Promise.resolve();
     }
-    const options = {agent, allVersions: true};
+    const options = {allVersions: true};
+
+    const scope = packageName.includes('/') && packageName.split('/')[0];
+    const proxyAgent = getProxyAgent(scope);
+    if (proxyAgent) {
+        options.agent = proxyAgent;
+    }
+
     if (field === 'time') {
         options.fullMetadata = true;
     }
+
     return packageJson(packageName, options).then(result => {
         if (field.startsWith('dist-tags.')) {
             const split = field.split('.');
--- a/lib/package-managers/npm-proxy.js
+++ b/lib/package-managers/npm-proxy.js
@@ -1,0 +1,88 @@
+const url          = require('url');
+const ProxyAgent   = require('https-proxy-agent');
+const npmConf      = require('npm-conf')();
+const proxyUrl     = require('get-proxy')();
+const registryUrl  = require('registry-url');
+
+const mainRegistryUrl = registryUrl();
+
+let proxyAgent;
+let noProxy = false;
+let noProxyEnv;
+let noProxyRegExps;
+
+const noProxyRegExpCheck = (parsedUrl) => {
+    for (const regExp of noProxyRegExps) {
+        if (regExp.test(parsedUrl.hostname) || regExp.test(parsedUrl.host)) {
+            return true;
+        }
+    }
+    return false;
+};
+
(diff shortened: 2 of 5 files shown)


lib/package-managers/npm.js
```diff
@@ -9,21 +10,37 @@ const packageJson  = require('package-json');
...
         key: npmConf.get('key'),
         cert: npmConf.get('cert'),
         rejectUnauthorized: npmConf.get('strict-ssl')
     });
+
+    noProxy = process.env.NO_PROXY || process.env.no_PROXY || process.env.no_proxy;
+    if (noProxy === '*') {
+        noProxy = true;
+    } else if (noProxy) {
+        noProxy = noProxy.split(/\s*,\s*/).map(np => {
```
raineorshine: Can you add a comment clarifying this somewhat arcane block of code? Also, I am a little confused about `noProxy`'s type. Is it a Boolean or something else? I would re-using the same variable for different typed values.

lib/package-managers/npm.js
```diff
@@ -50,7 +67,26 @@ function view(packageName, field, currentVersion) {
...
+    const options = {allVersions: true};
+
+    if (proxyAgent && noProxy !== true) {
+        options.agent = proxyAgent;
+
+        if (noProxy) {
+            const scope = packageName.includes('/') && packageName.split('/')[0];
+            const _registryUrl = registryUrl(scope);
+            const parsedRegistryUrl = URL ? new URL(_registryUrl) : url.parse(_registryUrl);
+            console.log(parsedRegistryUrl);
```
raineorshine: Are these console.log's for debugging? If not, they should respond to `--loglevel`, like [`print`](https://github.com/tjunnone/npm-check-updates/blob/master/lib/npm-check-updates.js#L46-L52).

lib/package-managers/npm.js
```diff
@@ -50,7 +67,26 @@ function view(packageName, field, currentVersion) {
     if (currentVersion && (!semver.validRange(currentVersion) || versionUtil.isWildCard(currentVersion))) {
         return Promise.resolve();
     }
-    const options = {agent, allVersions: true};
+    const options = {allVersions: true};
+
+    if (proxyAgent && noProxy !== true) {
```
raineorshine: Prefer `proxyAgent && !noProxy`

raineorshine:
Thank you!

Take a look at the travis build, too, when you get a chance. I'm not sure the new tests are working in the CI environment.

lib/package-managers/npm.js
stoically: Added comments and tried to clear things up a bit by renaming and not re-using. Let me know if there's still something unclear.

lib/package-managers/npm.js
stoically: Yep, you catched me in the middle of debugging the failing travis builds. Removed.

lib/package-managers/npm.js
stoically: That wouldn't have worked in this case, because the intention was to exactly check for `true`, because `noProxy` could have been an array too. A bit confusing; I've tried to address that.

stoically:
Thanks for the quick review. Yep, on travis overwriting the registry url with the env var didn't work for some reason - mocking `registry-url` now instead.

lib/package-managers/npm.js
```diff
@@ -9,21 +10,46 @@ const packageJson  = require('package-json');
...
         keepAlive: true,
-        maxSockets: npmConf.get('maxsockets') || 50, // https://docs.npmjs.com/misc/config#maxsockets
+        maxSockets: npmConf.get('maxsockets'),
         ca: npmConf.get('ca'),
         key: npmConf.get('key'),
         cert: npmConf.get('cert'),
         rejectUnauthorized: npmConf.get('strict-ssl')
     });
+
+    noProxyEnv = process.env.NO_PROXY || process.env.no_PROXY || process.env.no_proxy;
```
raineorshine: This block is still missing a high-level of explanation of *why* this code exists. Low-level comments are fine when they clarify a confusing statement, but not when they are a literal translation of the code (`Replace dots with escaped dots and asterisks with .* matchers `). Someone who is not familiar with proxies would still have no idea *why* this code is here. We need a short description of what the code does generally, and how it is applied specifically in ncu.

For example:

> The NO_PROXY environment variable specifies URLs that should be excluded from proxying. Generate regular expressions for each NO_PROXY url to be tested against the registry URL later when fetching package info.

stoically: You're right. Thanks for providing a useful comment for that chunk of code, I've added it accordingly and also linked the no_proxy documentation from curl for more details.

stoically:
I've realized that the code would run the no_proxy check against the main registry url for every package, although once would be enough. Refactored accordingly - also added a test for checking against scoped registry urls and refactored the other test assertions.

stoically:
Ok, so, since the proxy code was getting a bit more complex, I've moved it into its own file and adjusted the tests (no more mocking of package-json needed). Hope it's ok now. Let me know if you want to see anything changed. Thanks!

lib/package-managers/npm.js
```diff
@@ -6,25 +6,8 @@ const Promise = require('bluebird');
...
-        keepAlive: true,
-        maxSockets: npmConf.get('maxsockets') || 50, // https://docs.npmjs.com/misc/config#maxsockets
-        ca: npmConf.get('ca'),
-        key: npmConf.get('key'),
-        cert: npmConf.get('cert'),
-        rejectUnauthorized: npmConf.get('strict-ssl')
-    });
-}
+
+const {proxyAgent, noProxyCheck} = require('./npm-proxy');
```
raineorshine: Exporting two tightly-coupled variables seems like could be improved by a single export. If they are always used in tandem, the API should reflect this. I would suggest exporting a single `proxyAgent` function that returns the `agent` or `null` is NO_PROXY and optionally takes a `scope` as an argument.  `packageName.includes('/') && packageName.split('/')[0] ` would be in `npm.js`. What do you think?

stoically: Makes sense. I've addressed the proposed changes (which also make the tests easier to understand).

PR #526, opened 2019-03-29
Add proxy.auth support

nakagawa424:


lib/package-managers/npm-proxy.js
```diff
@@ -39,10 +39,11 @@ const noProxyCheck = (scope) => {
 };
 
 if (proxyUrl) {
-    const proxy = global.URL ? new URL(proxyUrl) : url.parse(proxyUrl);
```
stoically: Thanks for the PR. The legacy url API is [deprecated](https://nodejs.org/docs/latest/api/url.html#url_legacy_url_api), so we'd need a code-path handling both `auth` from legacy url, and `username`:`password` from WHATWG URL to support future versions.

lib/package-managers/npm-proxy.js
```diff
@@ -43,6 +43,7 @@ if (proxyUrl) {
     proxyAgent = new ProxyAgent({
         host: proxy.hostname,
         port: proxy.port,
+        auth: proxy.username ? proxy.username + (proxy.password ? ':' + proxy.password : '') : proxy.auth,
```
stoically: Although it might not happen in the wild, the `username` can be empty and only a password is set. So maybe check for `global.URL ?` instead.

PR #528, opened 2019-04-04
Add support for ssl npmrc configurations

stoically:
Fixes https://github.com/tjunnone/npm-check-updates/issues/519#issuecomment-479937779

Can be tested with `.npmrc`

```
registry=https://expired.badssl.com/
strict-ssl=false
```


raineorshine:
Published in `v3.1.7`

PR #539, opened 2019-05-03
Add `yarn add` message

rigwild:
It may be useful.


raineorshine:
Thanks for the suggestion. I would personally prefer sticking with instructions for npm. Any yarn user will be used to translating. There are other package manager frontends also, and I wouldn't want to add them.

I would be open to an option (and hence something that can be added to `.ncurc`) but even that seems a bet heavy.

PR #546, opened 2019-06-20
Replace package-json with pacote

stoically:
After realizing that npm supports yet another way to configure NO_PROXY ([in the npm config](https://docs.npmjs.com/misc/config#noproxy)), I looked into [pacote](https://github.com/zkat/pacote) (which is used by npm under the hood) instead of adding another proxy thing to ncu. It was almost too easy.

Pros:
- Maintenance burden of keeping in line with npm changes is gone (together with 300 LOC)
- Full npm config compatibility
- Full proxy support
- 1 instead of 5 direct dependencies
- And it's a bit faster too: ~8s instead of ~10s to check ~1500 dependencies

Cons:
- More packages in the dependency tree (about ~25-30)
- No `.domain.tld` / `*.domain.tld` wildcard NO_PROXY  support, but no one expects that in the npm world anyway, and one could open an issue over at [make-fetch-happen](https://github.com/zkat/make-fetch-happen) (dependency of pacote) if needed


raineorshine:
Beautiful! Thanks for all your hard work on this issue.

stoically:
Thanks for merging!

PR #550, opened 2019-06-29
Drop bluebird

proteria:
Node.js has great support for Promise for all versions supported by `npm-check-udpates`. This pull request removes the dependency for `bluebird`.

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -402,38 +402,40 @@ async function run(options={}) {
     print(options, 'Initializing...', 'verbose');
 
     if (options.packageManager === 'npm' && !options.prefix) {
+        // eslint-disable-next-line require-atomic-updates
         options.prefix = await packageManagers.npm.defaultPrefix(options);
     }
 
     let timeout;
-    let timeoutPromise = Promise.resolve();
-
+    let timeoutPromise = new Promise(r => r);
     if (options.timeout) {
         const timeoutMs = _.isString(options.timeout) ? parseInt(options.timeout, 10) : options.timeout;
         timeoutPromise = new Promise((resolve, reject) => {
             timeout = setTimeout(() => {
-                // must catch the error and reject expicitly since we are in a setTimeout
+                // must catch the error and reject explicitly since we are in a setTimeout
+                const error = `Exceeded global timeout of ${timeoutMs}ms`;
+                reject(error);
                 try {
-                    programError(options, chalk.red(`Exceeded global timeout of ${timeoutMs}ms`));
-                } catch (e) {
-                    reject(e);
-                }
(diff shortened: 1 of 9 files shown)


raineorshine:
Thank you!

It looks like [finally is not supported](https://node.green/#ES2018-features-Promise-prototype-finally) on node v8, which explains the [build failures](https://travis-ci.org/tjunnone/npm-check-updates/jobs/552025002). We would have to find a solution to that before we can merge.

raineorshine:
Great, thanks for your persistence. Any thoughts about the remaining errors on node 8? It still seems related to the `finally` statement. The test tries to delete temporary files as part of its cleanup process in the finally block.

raineorshine:
@proteria Any ideas? Thanks!

proteria:
The errors relating to deleting temporary files were caused by a race condition and fixed.

raineorshine:
Thanks! Do you have time to look into the build error in node `v8`, or shall I?

proteria:
That error only occasionally reproduces. I have no idea why it only happens on Node 8.

stoically:
[`getAnalysis`](https://github.com/tjunnone/npm-check-updates/pull/550/files#diff-1415f5139deb67a37325f1bde416bcbdR437) sometimes wins the race with node8. My guess is that older node/v8's handled putting promises / timeouts in the event queue differently. A possible way to solve this would be to instantly resolve the setTimeout, something along the lines of https://github.com/stoically/npm-check-updates/commit/f2e5af8700ad96c7c40538b07bef0df5759a2e40

lib/npm-check-updates.js
```diff
@@ -406,34 +406,35 @@ async function run(options={}) {
     }
 
     let timeout;
-    let timeoutPromise = Promise.resolve();
-
+    let timeoutPromise = new Promise(r => r);
```
stoically: I'm curious, is there a reason you prefer that compared to `Promise.resolve()`?

raineorshine:
@proteria This does the trick: https://github.com/tjunnone/npm-check-updates/commit/bc36bb0e64fd659b41b22c05ee0552b04d994c3a

I will merge the PR and then cherry-pick bc36bb

stoically:
I doubt that helps - if anything passing in the actual package.json data makes `getAnalysis` even faster, so more likely to win the race against the `setTimeout`.

raineorshine:
@stoically Actually, I tested it on node v8 and it works. The package.json is large enough that it reliably takes more than 1ms to resolve, not to mention the fs read.

stoically:
Ah, got it, you make `getAnalysis` slower by passing in more stuff to process. Didn't thought about doing it that way, nice.

PR #552, opened 2019-07-02
Fix missing dependency.

GongT:
This is a phantom dependency used by `lib/versionmanager.js`

### DIFF
--- a/lib/package-managers/npm.js
+++ b/lib/package-managers/npm.js
@@ -25,7 +25,7 @@ function parseJson(result, data) {
 
 /**
  * @param packageName   Name of the package
- * @param field         Field such as "versions" or "dist-tags.latest" are parsed from the package-json result (https://github.com/sindresorhus/package-json)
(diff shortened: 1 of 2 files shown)


stoically:
Good catch. However, package-json was replace by pacote in #546 - so if you don't mind, would change this PR to remove [this](https://github.com/tjunnone/npm-check-updates/blob/80871a8a5c3cdaeaa6ca9d7dfa99f7de9ffd94ab/lib/versionmanager.js#L8) and [that](https://github.com/tjunnone/npm-check-updates/blob/80871a8a5c3cdaeaa6ca9d7dfa99f7de9ffd94ab/lib/versionmanager.js#L386) instead? Thanks!

GongT:
Ok, removed now.

I confirmed the error message is like "404 Not Found - GET https://registry.npmjs.org/debug1111 - Not found".

raineorshine:
Thank you @GongT!

PR #569, opened 2019-07-17
Clarify filter argument

raineorshine:
Make it clearer that the filter and exclude arguments can take space-delimited lists and that regexes must be enclosed by `/`.


raineorshine:
Published in `v3.1.22`

PR #577, opened 2019-09-07
Don't say "latest version" and "ncu -u" when nothing should be upgraded

Primajin:
Closes #575

Before:
Bildschirmfoto 2019-09-07 um 14 54 39

After:
Bildschirmfoto 2019-09-07 um 14 54 26

Also removes some duplicated code and updates some JSDoc.

PR #579, opened 2019-09-10
Fix global dependencies message

Primajin:
Fixes regression introduced in #577 

Please test thoroughly 😉 

global ncu vs. local (this PR):
Bildschirmfoto 2019-09-10 um 20 05 24


raineorshine:
Thanks! Published in `v3.1.23`

PR #584, opened 2019-10-03
check the internet connection

fscherwi:


raineorshine:
Great addition!

Can you ensure that this works when an alternate registry is specified by the user via `--registry`?

Also please confirm that `is-reachable` works through proxy servers. We've had a lot of complications due to proxy incompatibilities since we switched to a different package retrieval mechanism.

Thank you!

fscherwi:
I have added support for bower repository checks.

raineorshine:
Thank you, although you need to specifically add support for a custom registry via the [`--registry`](https://github.com/tjunnone/npm-check-updates/search?q=registry&unscoped_q=registry) option.

fscherwi:
I have added support for custom registries and an option to disable the online check.

raineorshine:
Hey! I just realized there is a bit of a problem with this. Given that `is-reachable` (necessarily) does a TCP handshake, it is going to add a delay to npm-check-updates which is unnecessary for 99% of users (who are connected to the internet/registry). It doesn't make sense to slow down 99% of users for the 1% that are not connected.

We may want to take a different tack. When there is no internet, ncu reports "All dependencies match the latest package versions" which is clearly misleading. Since ncu ignores individual package 404's, then presumably all of the dependencies are returning 404 under the hood when there is no internet. What if we detect when all dependencies return 404 and *then* check if there is an internet connection?

raineorshine:
Thanks for your effort on this! I think you are heading in the right direction. I am not willing to use globals unfortunately. Global mutability is widely discouraged and considered a bad practice. If you would like to continue working on this issue, I suggest architecting it such that the number of 404's are piped back through the typical return mechanism.

lib/versionmanager.js
```diff
@@ -11,6 +11,9 @@ const prompts = require('prompts');
 // keep order for setPrecision
 const DEFAULT_WILDCARD = '^';
 
+// track if all packages return error 404
+let allNotFoundErrors;
```
raineorshine: This is still effectively global (to the file). To be clear, `let` (more broadly, mutability) *is a bad practice*. I recommend reading a little more about functional programming if this point is confusing for you.

KingDarBoja:
Hey @fscherwi, any chance on fixing the latest feedback changes? 💯

raineorshine:
Closing due to inactivity and unresolved request https://github.com/raineorshine/npm-check-updates/pull/584#issuecomment-539650255.

PR #585, opened 2019-10-04
- npm: Add recommended `bugs` field (used on npmjs.com)

brettz9:
Btw, re: `package.json`, I don't know your frequency of updating your own deps, so I didn't update those.

This field is useful on npmjs.com for pointing directly to the issues page. (and it quiets my `package.json` linter!)

Also, the snyk tests failed (and were failing before my changes) due to the need for one patch; not sure how you wanted to resolve.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "url": "https://github.com/tjunnone/npm-check-updates.git"
   },
   "homepage": "https://github.com/tjunnone/npm-check-updates",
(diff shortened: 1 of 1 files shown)


raineorshine:
Thanks!

Any snyk failures will be addressed in the next release, and personally I just use semver version ranges to control dependencies. There are no major version updates at this time.

PR #586, opened 2019-10-04
Expose ncurc retrieval to programmatic API

brettz9:
I needed this with some programmatic usage, and thought it best for your other consumers not to have to duplicate this or resort to the CLI to get ncurc retrieval logic.

One breaking change I did not make (though I made it easier to implement if you want to go that route), but which I think would be useful--allow the rcLoader current working directory to be based by default on the directory of `packageFile`, since they would presumably usually be in the same (project root) directory.

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -445,9 +445,8 @@ function getNcurc({configFileName, configFilePath, packageFile} = {}) {
     const rcFile = rcLoader('ncurc', {
         configFileName: configFileName || '.ncurc',
         defaultExtension: ['.json', '.yml', '.js'],
-        cwd: packageFile ?
-            path.dirname(packageFile) :
-            configFilePath || undefined
+        cwd: configFilePath ||
+            (packageFile ? path.dirname(packageFile) : undefined)
     });
     return rcFile && rcFile.config;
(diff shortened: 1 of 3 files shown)


lib/npm-check-updates.js
```diff
@@ -433,6 +434,25 @@ async function run(options={}) {
...
+ * @param {string} [cfg.packageFile]
+ * @returns {PlainObject|undefined}
+ */
+function getNcurc({configFileName, configFilePath, packageFile} = {}) {
+    const rcFile = rcLoader('ncurc', {
+        configFileName: configFileName || '.ncurc',
+        defaultExtension: ['.json', '.yml', '.js'],
+        cwd: packageFile ?
+            path.dirname(packageFile) :
+            configFilePath || undefined
```
raineorshine: `configFilePath` should have precedence over `packageFile` if present.

brettz9: Makes sense, thanks. Changed.

brettz9:
I've applied the changes, including the breaking change for the CLI.

raineorshine:
Published in `v3.1.24`.

I published as a patch as the default path was clearly a bug.

PR #588, opened 2019-10-06
Jsdoc

brettz9:
- Use standard `jsdoc` tags and casing, denoting optional items (and add some types/return values)
- Minor refactoring: Remove `options` check in `npm.js`' `defaultPrefix` function as function does not later allow a non-object anyways
- Minor refactoring: Avoid else after throw

There is still one non-standard jsdoc-style tag you have in place (`@sideeffects`), but I didn't change that as there is no standard equivalent (there is a `@modifies` tag, but that is for side effects on objects) and JSDoc does allow custom tags.

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -304,8 +304,14 @@ function initOptions(options) {
     });
 }
 
+/**
+* @typedef {Array} PkgInfo
+* @property 0 pkgFile
+* @property 1 pkgData
+*/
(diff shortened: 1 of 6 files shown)


brettz9:
There could be more precision with types, and typedefs, but I figure they can be added later.

brettz9:
Greater specificity could be added for some of the types, but I figure that could be added as time goes on.

PR #589, opened 2019-10-06
Document other CLI-accessible commands

brettz9:
I did a review of the options which were neither set at runtime nor documented, and came up with the three in this PR.

Please let me know if there should be corrections in the summary.


raineorshine:
Published in `v3.1.24`

PR #592, opened 2019-10-07
update Travis CI build script

fscherwi:
Only test active node versions.

https://nodejs.org/en/about/releases/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,9 @@
-sudo: false
-
 language: node_js
 
 node_js:
   - 8
   - 10
-  - 11
+  - 12
 
 notifications:
   email: false


raineorshine:
Any pragmatic reason for this?

We would need to do a major version change to drop support for a version of node that is currently supported.

fscherwi:
I haven't said to drop support for node 11. I have only changed the Travis CI Node versions to actively maintained versions. Node 11 reached End-of-life on 2019-06-30. But if you want to test against node 11 I can add it back to Travis, but then you also have to test against node 9 which is also in your supported version range, but no more maintained.

raineorshine:
Thank you, I appreciate the suggestion. I don't think it makes sense to maintain support ostensibly while disabling the tests, as they go together in my mind.

PR #595, opened 2019-10-22
Suggest to use npx as an alternative to install the package globally

Zhouzi:
This PR documents an alternative way of using `npm-check-updates` without having to install it globally. It relies on npm's npx tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,14 @@ Installation
 npm install -g npm-check-updates
 ```
 
+Alternatively, you can use this package without installing it globally:
+
+```sh
+npx npm-check-updates
+```
+
+Note: `npx` comes with npm 5+.
+
 Usage
 --------------
 Show any new dependencies for the project in the current directory:


raineorshine:
Thank you for the suggested update. I would prefer to exclude general educational material about `npx` in the `npm-check-updates` README. I trust that if people know about `npx` they will know how to use it with `npm-check-updates`. I think a reasonable argument could be made either way, that is just my preference.

Zhouzi:
I have personally first learned about `npx` by reading eslint-config-airbnb's README. I also believe that `npx` should be favored over installing global dependencies.

But your points are totally valid and it's no big deal so I'm fine closing this PR ;)

adriengibrat:
Come'on, very few people know about the `-p` option of npx.

And as the bin cli is not named like the npm package it's tricky!

For the record, the alternative to install it globally is to run:

```bash
npx -p npm-check-updates ncu
```

stoically:
Just `npx npm-check-updates` works as well - since there are two bins

Issue #599, opened 2019-10-28
Feature: "--engines-node" parameter


webdevium:
- [x] I have read the list of [known issues](https://github.com/tjunnone/npm-check-updates#known-issues) before filing this issue.
- [x] I have searched for [similiar issues](https://github.com/tjunnone/npm-check-updates/issues) before filing this issue.

* node version: v10.17.0
* npm version: 6.11.3
* npm-check-updates version: 3.1.25

---------------------------

**Given:** abstract package.json
```js
{
	"name": "awesome/package",
	"engines": {
    	"node": ">=6.0.0"
  	},
	"dependencies": {
		"del": "3.0.0"
	}
}
```

### Current behavior

**When:** I run `ncu`

**Then:** I see

```sh
del		3.0.0 → 5.1.0
```

### What I want

**Then:** I run `ncu --engines-node`

**Should be:** I'll see

```sh
del		3.0.0 → 4.1.1
```

> because v4.1.1 is the latest version with **"engines.node": ">=6"**
> proof of:
> - [v4.1.1](https://github.com/sindresorhus/del/blob/v4.1.1/package.json#L13) 
> - [v5.0.0](https://github.com/sindresorhus/del/blob/v5.0.0/package.json#L13)

Some specific cases:
- if **engines.node** not exists in my `package.json`, I'll see 
```sh
del		3.0.0 → 5.1.0
```
- if **engines.node** not exists in new version of **del** package (in metadata for each version greater than currently installed), I'll see
```sh
All dependencies match the latest package versions :)
```

raineorshine:
Great suggestion! Definitely open to PR's.

webdevium:
Done!

PR #602, opened 2019-10-29
Added --engines-node functionality

webdevium:
#599

### DIFF
--- a/bin/ncu
+++ b/bin/ncu
@@ -23,6 +23,7 @@ program
     .option('--cwd <path>', 'Used as current working directory for `spawn` in npm listing')
     .option('--dep <dep>', 'check only a specific section(s) of dependencies: prod|dev|peer|optional|bundle (comma-delimited)')
     .option('-e, --error-level <n>', 'set the error-level. 1: exits with error code 0 if no errors occur. 2: exits with error code 0 if no packages need updating (useful for continuous integration). Default is 1.', cint.partialAt(parseInt, 1, 10), 1)
+    .option('--engines-node', 'upgrade to version which satisfies engines.node range')
     .option('-f, --filter <matches>', 'include only package names matching the given string, comma-or-space-delimited list, or /regex/')
     .option('-g, --global', 'check global packages instead of in the current project')
     // program.json is set to true in programInit if any options that begin with 'json' are true
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -188,6 +188,10 @@ function analyzeProjectDependencies(options, pkgData, pkgFile) {
 
     print(options, `Fetching ${vm.getVersionTarget(options)} versions...`, 'verbose');
 
+    if (options.enginesNode) {
+        options.enginesNode = _.get(pkg, 'engines.node');
+    }
+
     return vm.upgradePackageDefinitions(current, options).then(async ([upgraded, latest]) => {
         const {newPkgData, selectedNewDependencies} = await vm.upgradePackageData(pkgData, current, upgraded, latest, options);
 
--- a/lib/package-managers/npm.js
+++ b/lib/package-managers/npm.js
@@ -6,6 +6,8 @@ const versionUtil  = require('../version-util.js');
 const spawn        = require('spawn-please');
 const pacote = require('pacote');
 
+const TIME_FIELDS = ['modified', 'created'];
+
 // needed until pacote supports full npm config compatibility
 // See: https://github.com/zkat/pacote/issues/156
 const npmConfig = {};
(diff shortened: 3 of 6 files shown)


lib/npm-check-updates.js
```diff
@@ -188,6 +188,11 @@ function analyzeProjectDependencies(options, pkgData, pkgFile) {
 
     print(options, `Fetching ${vm.getVersionTarget(options)} versions...`, 'verbose');
 
+    if (options.enginesNode) {
+        options.enginesNode = _.get(pkg, 'engines.node');
+        options.versionTarget = 'latest';
```
raineorshine: Can you explain forcing versionTarget to `latest`? Do other version targets not work with `--engines-node`?

lib/package-managers/npm.js
```diff
@@ -159,15 +161,35 @@ module.exports = {
...
+                    const _version = semver.minVersion(options.enginesNode);
+                    if (!_version || !_version.version) {
+                        return versions;
+                    }
+                    return Object.keys(versions).filter((version) => {
+                        let versionEnginesNode = _.get(versions[version], 'engines.node');
+                        if (versionEnginesNode && semver.satisfies(_version.version, versionEnginesNode)) {
+                            return true;
+                        }
+                        return false;
```
raineorshine: Prefer `return versionEnginesNode && semver.satisfies(_version.version, versionEnginesNode)`

lib/package-managers/npm.js
```diff
@@ -159,15 +161,35 @@ module.exports = {
...
+     * @param {{}} options
      * @returns {Promise}
      */
-    latest(packageName, currentVersion, pre) {
+    latest(packageName, currentVersion, options) {
+
+        if (options.enginesNode) {
+            return view(packageName, 'versions-object', currentVersion)
+                .then((versions) => {
+                    const _version = semver.minVersion(options.enginesNode);
```
raineorshine: Prefer `const minVersion`

lib/package-managers/npm.js
```diff
@@ -159,15 +161,35 @@ module.exports = {
...
+                            return true;
+                        }
+                        return false;
+                    });
+                })
+                .then(filterOutPrereleaseVersions)
+                .then(_.last);
+        }
+
         return view(packageName, 'dist-tags.latest', currentVersion)
```
raineorshine: Let's put this section in an `else if` to make it clear that it is exclusive with the `enginesNode` case.

lib/npm-check-updates.js
webdevium: Because my idea was: 
if I enable engines matching, I want to get exactly latest version.

I can try to implement engines matching to all targets, but I think this will be crazy changes...

lib/package-managers/npm.js
```diff
@@ -63,6 +63,8 @@ function view(packageName, field, currentVersion) {
             }
         } else if (field === 'versions') {
             return Object.keys(result[field]);
+        } else if (field === 'versions-object') {
+            return result['versions'];
```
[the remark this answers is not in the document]
webdevium: Maybe, eslint rule for this? :)

lib/npm-check-updates.js
raineorshine: I can't justifiably add this feature only for `latest` when it should work for all version targets. This will require some extra code, but I trust you will be able to abstract out redundant code appropriately. If you are unsure the best approach architecturally, feel free to ask questions or request feedback on an initial approach. Thank you!

webdevium: @raineorshine How about this?
- minimum extra code
- works with all targets

lib/package-managers/npm.js
```diff
@@ -85,6 +88,49 @@ function isPre(version) {
...
+    return minVersion.version;
+}
+
+/**
+ *
+ * @param {{}} version
+ * @param {{}} minVersion
+ * @returns {boolean}
+ */
+function isSatisfies(version, minVersion) {
```
raineorshine: Since this function is small and only used once, I would recommend inlining it at the call point.

lib/package-managers/npm.js
```diff
@@ -159,78 +205,92 @@ module.exports = {
...
-            });
+    latest(packageName, currentVersion, options) {
+        return view(packageName, 'versions', currentVersion)
+            .then((versions) => {
+                return filterSatisfiesMinVersion(versions, options.enginesNode);
+            })
+            .then((versions) => {
+                return filterOutPrereleaseVersions(versions, options.pre);
+            })
+            .then(_.last);
```
raineorshine: The greatest version number is not necessarily the latest version. Therefore the `latest` dist-tag must be checked explicitly. I'm afraid there does not appear to be a unit test in place for this situation.

lib/package-managers/npm.js
```diff
@@ -85,6 +88,49 @@ function isPre(version) {
     return versionUtil.getPrecision(version) === 'release';
 }
 
+/**
+ * @param {{}}     versions    Object with all versions
+ * @param {String} enginesNode Package engines.node range
+ * @returns {Array} An array of versions which satisfies engines.node range
+ */
+function filterSatisfiesMinVersion(versions, enginesNode) {
```
raineorshine: Let's call this function `doesSatisfyEnginesNode` to better capture the purpose.

bin/ncu
```diff
@@ -23,6 +23,7 @@ program
     .option('--cwd <path>', 'Used as current working directory for `spawn` in npm listing')
     .option('--dep <dep>', 'check only a specific section(s) of dependencies: prod|dev|peer|optional|bundle (comma-delimited)')
     .option('-e, --error-level <n>', 'set the error-level. 1: exits with error code 0 if no errors occur. 2: exits with error code 0 if no packages need updating (useful for continuous integration). Default is 1.', cint.partialAt(parseInt, 1, 10), 1)
+    .option('--engines-node', 'upgrade to latest version which satisfies engines.node range')
```
raineorshine: You can omit 'latest' here since we are implementing `--engines-node` to be version target agnostic.

lib/package-managers/npm.js
```diff
@@ -56,25 +58,26 @@ function view(packageName, field, currentVersion) {
...
-        } else if (field === 'versions') {
-            return Object.keys(result[field]);
+        if (field === 'versions') {
+            return result.versions;
         } else {
-            return result[field];
+            return {
+                [field]: result[field],
+                versions: result.versions,
+            };
```
raineorshine: This is problematic since it changes the semantics of the `view` function in an unconventional way. Previously `view` returned the value of the field requested, which was simple and intuitive. Always returning `versions` is counterintuitive, even though it accomplishes what you want. i.e. It's a bit too much of a hack.

I would recommend the following approach: Define two functions: `viewOne` and `viewMany`. `viewMany` will be the same as the existing `view` but take an array of fields as the second argument and return an object containing each of those fields with their values. This is a more general solution than the one presented. Then `viewOne` simply wraps `viewMany` and returns only the field value, preserving the current behavior (and requiring no change on the caller side).

lib/package-managers/npm.js
```diff
@@ -85,6 +88,49 @@ function isPre(version) {
...
+    return _.keys(versions).filter((version) => {
+        return isSatisfies(versions[version], minVersion);
+    });
+}
+
+/**
+ * @param {String} enginesNode
+ * @return {Boolean|String} Minimum version of node
+ */
+function parseMinVersion(enginesNode) {
```
raineorshine: Inline, as described previously.

lib/package-managers/npm.js
```diff
@@ -159,78 +205,92 @@ module.exports = {
...
-                }
-            });
+    latest(packageName, currentVersion, options) {
+        return view(packageName, 'versions', currentVersion)
+            .then((versions) => {
+                return filterSatisfiesMinVersion(versions, options.enginesNode);
+            })
+            .then((versions) => {
+                return filterOutPrereleaseVersions(versions, options.pre);
+            })
```
raineorshine: Since `filterSatisfiesMinVersion` is not asynchronous, it does not need a separate `then` statement and can be combined with the one for `filterOutPrereleaseVersions`.

(The same can be done for `_.last` for that matter)

lib/package-managers/npm.js
```diff
@@ -159,78 +205,92 @@ module.exports = {
...
-     * @param {boolean} pre
+     * @param {{}} options
      * @returns {Promise}
      */
-    greatest(packageName, currentVersion, pre) {
+    greatest(packageName, currentVersion, options) {
         return view(packageName, 'versions', currentVersion)
+            .then((versions) => {
+                return filterSatisfiesMinVersion(versions, options.enginesNode);
+            })
```
raineorshine: Can be combined with following `then`.

lib/package-managers/npm.js
```diff
@@ -159,78 +205,92 @@ module.exports = {
...
-    greatestMajor(packageName, currentVersion, pre) {
-        return view(packageName, 'versions', currentVersion).then(versions => {
-            const resultVersions = pre ? versions : filterOutPrereleaseVersions(versions);
-            return versionUtil.findGreatestByLevel(resultVersions, currentVersion, 'major');
-        });
+    greatestMajor(packageName, currentVersion, options) {
+        return view(packageName, 'versions', currentVersion)
+            .then((versions) => {
+                return filterSatisfiesMinVersion(versions, options.enginesNode);
+            })
```
raineorshine: Can be combine with following `then`.

lib/package-managers/npm.js
```diff
@@ -159,78 +205,92 @@ module.exports = {
...
-            const resultVersions = pre ? versions : filterOutPrereleaseVersions(versions);
-            return versionUtil.findGreatestByLevel(resultVersions, currentVersion, 'minor');
-        });
+    greatestMinor(packageName, currentVersion, options) {
+        return view(packageName, 'versions', currentVersion)
+            .then((versions) => {
+                return filterSatisfiesMinVersion(versions, options.enginesNode);
+            })
+            .then(versions => {
+                const resultVersions = options.pre ? versions : filterOutPrereleaseVersions(versions, options.pre);
```
raineorshine: Since `filterOutPrerelease versions` was modified to handle the `pre` check, I think you can simply this line as such:

```
const resultVersions = filterOutPrereleaseVersions(versions, options.pre)
```

webdevium:
@raineorshine Thanks for recommendations.
I tried to do as good as I can do. I apologize if at some points I misunderstood you.

raineorshine:
Thank you so much for your persistence and willingness to work with the feedback! You've done more than most. Thanks for contributing to npm-check-updates!

raineorshine:
@webdevium I did some further cleanup and conformance to the functional programming style in c417828caca04dc22d5254602490a9f03ac289d9 if you are curious.

raineorshine:
Published in `v3.2.0`

webdevium:
@raineorshine Thanks for your patience!

PR #603, opened 2019-10-30
Avoid variable suggesting a Promise is returned from `printUpgrades`

brettz9:
- Refactoring: Avoid variable suggesting a Promise is returned from `printUpgrades`
- Docs: Add missing `returns` to jsdocs
- npm: Update devDeps (chokidar-cli, eslint, mocha, snyk)

(I noticed that `pacote` had a major version upgrade as well, but though it had no higher `engines`, using it as is, caused some tests to fail.)

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -100,8 +100,10 @@ function createDependencyTable() {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


raineorshine:
Thank you!

PR #607, opened 2019-11-05
Switched from pacote to package-json

webdevium:
The main reason: due to the fact that we use only one function of `pacote`, it is too big for `ncu`.
`package-json` just knows how to do only what we need :)
Accordingly, it has fewer dependencies.


stoically:
Thanks for the suggestion, but [we already had package-json](https://github.com/tjunnone/npm-check-updates/pull/497) and it was and interesting endeavor to add proxy support that covers all use-cases, and having to try to stay in line with all possible npm config stuff, which pacote (and its dependencies) naturally cover.

Issue #608, opened 2019-11-06
Cannot read property 'versions' of undefined


HitomiTenshi:
- [x] I have read the list of [known issues](https://github.com/tjunnone/npm-check-updates#known-issues) before filing this issue.
- [x] I have searched for [similiar issues](https://github.com/tjunnone/npm-check-updates/issues) before filing this issue.

* node version: 13.0.1
* npm version: 6.13.0
* npm-check-updates version: 3.2.0

---------------------------

Running `ncu -n` results in the following error when scanning one of my projects:

```
C:\Users\fegit1\AppData\Roaming\npm\node_modules\npm-check-updates\lib\npm-check-updates.js:392
    throw err;
    ^

TypeError: Cannot read property 'versions' of undefined
    at C:\Users\fegit1\AppData\Roaming\npm\node_modules\npm-check-updates\lib\package-managers\npm.js:226:64
    at async Promise.all (index 22)
    at async getAnalysis (C:\Users\fegit1\AppData\Roaming\npm\node_modules\npm-check-updates\lib\npm-check-updates.js:438:30)
    at async Object.run (C:\Users\fegit1\AppData\Roaming\npm\node_modules\npm-check-updates\lib\npm-check-updates.js:444:12)
```

Running `ncu` without the `-n` argument runs fine.

Doesn't happen on all of my projects, only some.

The previous version `v3.1.26` does not have this issue.

raineorshine:
Thank you for reporting! Can you please narrow it down to which dependency(ies) is causing the issue? Once you find it, please post the dependency name and version specified in your package.json so I can reproduce.

HitomiTenshi:
@raineorshine I just found the issue. It's a git repo dependency.

Run:  
`npm init` (yes to all)  
`npm install github:HitomiTenshi/pouchdb-browser --save`  
`ncu -n`

raineorshine:
@webdevium It looks like there is a regression from your PR. Could you look into this?

webdevium:
@raineorshine My PR, my regression, my fix

raineorshine:
Thanks!

Published in `v3.2.1`

@HitomiTenshi Did this fix the issue?

HitomiTenshi:
@raineorshine Yes it did, thanks!

PR #610, opened 2019-11-08
Added tests for invalid and wildCard ranges with --newest option

webdevium:
Fixed successful work in case of incorrect ranges.
For omitting errors like #608

### DIFF
--- a/test/test-ncu.js
+++ b/test/test-ncu.js
@@ -226,6 +226,47 @@ describe('npm-check-updates', function () {
             });
         });
 
+        it('should work with --newest option with any invalid or wildcard range', () => {
+            return Promise.all([
+                ncu.run({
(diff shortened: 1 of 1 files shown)


raineorshine:
Thank you!

webdevium:
@raineorshine Have you thought about improving branch coverage?

raineorshine:
You mean like unit tests for other branches? Sorry I'm not familiar. Can you explain a bit?

webdevium:
No, in **master** branch.
If you look at test coverage, you will see that not all conditions are covered in `if/else/ternary operator` statements. It is called branch coverage.

raineorshine:
Ah! That kind of branch :). Yes, we could really use better coverage!

PR #614, opened 2019-11-17
Fix the good displayed log in interactive mode

Tchoupinax:
Hello,

When using `ncu` in interactive mode i noticed it did not display the tip to launch `npm install` at the end like it does with `ncu -u`. I suggest this PR to allow to display this tip when the interactive mode installs at least one package.

Classic behaviour (for example)
```bash
あ→ ncu -u        
Upgrading /home/user/Documents/xxx/yyy/package.json
[====================] 36/36 100%

 cookie-universal-nuxt  ^2.0.18  →  ^2.0.19 
 eslint-plugin-nuxt       0.4.3  →    0.5.0

Run npm install to install new versions.
```

**Actual** interactive bahviour
```bash
あ→ ncu -i
Checking /home/user/Documents/xxx/yyy/package.json
[====================] 36/36 100%
✔ Do you want to upgrade: cookie-universal-nuxt ^2.0.18 → ^2.0.19? … yes
✔ Do you want to upgrade: eslint-plugin-nuxt 0.4.3 → 0.5.0? … no

 cookie-universal-nuxt  ^2.0.18  →  ^2.0.19 

```

**Expected** interactive bahviour (what the PR does)
```bash
あ→ ~/Downloads/npm-check-updates/bin/ncu -i
Checking /home/user/Documents/xxx/yyy/package.json
[====================] 36/36 100%
✔ Do you want to upgrade: cookie-universal-nuxt ^2.0.18 → ^2.0.19? … yes
✔ Do you want to upgrade: eslint-plugin-nuxt 0.4.3 → 0.5.0? … no

 cookie-universal-nuxt  ^2.0.18  →  ^2.0.19 

Run npm install to install new versions.  (<== this log looks better)
```

Be free to comment or to make another suggestion :)

Thank you for your work ;)

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -235,7 +235,7 @@ function analyzeProjectDependencies(options, pkgData, pkgFile) {
             // if there is a package file, write the new package data
             // otherwise, suggest ncu -u
             if (pkgFile) {
-                if (options.upgrade) {
+                if (options.upgrade || options.interactive) {
                     // short-circuit return in order to wait for write operation, but still return the same output
                     return writePackageFile(pkgFile, newPkgData)
                         .then(() => {


raineorshine:
Hi there and thanks for the suggestion!

`-i` and `-u` are currently independent options. Thus, if you run `-i` on its own it will *not* upgrade your package.json because you did not specify `-u`. You would have to run `ncu -i -u`. 

Perhaps this behavior is confusing and `-u` should be implied. That said, if `--json` or `--jsonDeps` is specified, `-u` should not be implied, as there is a valid use case for generating the json interactively.

Tchoupinax:
Hello,

Yeah indeed i think this behaviour is confusing and option `-u` should be implied because when you are in interactive mode, it is asking you if you want to upgrade or not and i expect that choosen package are updating at the end in package.json ^_^

So, the good way to patch it is to implicity add the upgrade option ? :)

Thank you for your answer

raineorshine:
Fixed in https://github.com/tjunnone/npm-check-updates/commit/757826486c038b6b94ff0023a2c2e8d6d489902c

Published in v3.2.2

Thanks!

PR #619, opened 2020-01-03
[Feat] Postinstall message about repo owner

### LINKED ISSUE #465: Issue a warning when repository owner is changed
- [x] I have read *most of* the list of [known issues](https://github.com/tjunnone/npm-check-updates#known-issues) before filing this issue.
- [x] I have searched for [similiar issues](https://github.com/tjunnone/npm-check-updates/issues) before filing this issue. (took a quick look, though)

---------------------------

This is an enhancement request: Due to a security vulnerability message of a dependency package, I took a look at the [issue](https://github.com/dominictarr/event-stream/issues/116) and understand the injection has been performed by a different person, not by the creator of the library. 

Original author drops maintaining the library, then someone wants it, he hands it over. The new guy implements a malicious code, publishes it. Currently even the original author has no publish rights on npm. 

What I could think of is that: npm-check-updates might warn us when owner is changed, like in the way: 

```
************************************ WARNING ******************************************

This repository's owner has changed. You are advised to check on the new owner. 
Please type the new owner's mail address to continue: ...

************************************ WARNING ******************************************
```

KingDarBoja:
Fixes #465 and also some missing lint errors that were causing the travis build to fail.

The only thing that I don't understand is why the `chalk` color isn't being applied on the postinstall after using the one from npm while locally it works.

Chalk doesnt work

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -192,22 +192,22 @@ function analyzeProjectDependencies(options, pkgData, pkgFile) {
         options.enginesNode = _.get(pkg, 'engines.node');
     }
 
-    print(options, '\nOptions (partial):', 'verbose')
+    print(options, '\nOptions (partial):', 'verbose');
     print(options, {
         registry: options.registry ? options.registry : null,
         pre: options.pre,
         packageManager: options.packageManager,
         json: options.json,
         enginesNode: options.enginesNode
-    }, 'verbose')
+    }, 'verbose');
 
     return vm.upgradePackageDefinitions(current, options).then(async ([upgraded, latest]) => {
 
-        print(options, '\nFetched:', 'verbose')
-        print(options, latest, 'verbose')
+        print(options, '\nFetched:', 'verbose');
+        print(options, latest, 'verbose');
 
-        print(options, '\nUpgraded:', 'verbose')
-        print(options, upgraded, 'verbose')
+        print(options, '\nUpgraded:', 'verbose');
+        print(options, upgraded, 'verbose');
 
         const {newPkgData, selectedNewDependencies} = await vm.upgradePackageData(pkgData, current, upgraded, latest, options);
 
--- a/lib/postinstall.js
+++ b/lib/postinstall.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const chalk = require('chalk');
+const util = require('util');
+const exec = util.promisify(require('child_process').exec);
 
 const stdOutCols = process.stdout.columns || 80;
 const warnWord = ' WARNING ';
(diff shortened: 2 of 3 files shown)


raineorshine:
Thanks! Could you explain the connection between `postinstall` and there being a new owner? I'm not sure I understand how they are connected. There is nothing in this PR which checks `npm owner ls PACKAGE`.

KingDarBoja:
> Thanks! Could you explain the connection between `postinstall` and there being a new owner? I'm not sure I understand how they are connected. There is nothing in this PR which checks `npm owner ls PACKAGE`.

You are right, the reason with this PR was addresses on the linked issue as someone asked to provide some message if the owner of this repo changed.

I would have to add the extra check if `npm owner ls PACKAGE` is different that the original one in order to trigger the postinstall message. It will let some room for any other custom message if package owner/maintainers want to display important things like breaking changes and so on, just saying.

stoically:
I don't think it makes much sense to work with stdout from (post)install scripts anymore, since it's planned to get removed soonish: https://github.com/npm/rfcs/pull/77#issuecomment-566644469

KingDarBoja:
I managed to retrieve the list of this package owner as example by using a process child, not sure if it is the ideal way to do it.

Of course we can apply the same logic for every package that is being updated. However, it have a big question about how the message should be displayed as I was thinking on displaying the entire list of packages whose owner changed on the postinstall script but noticed it was a **DUMB** idea as the author of the original issue was talking about **every** package being update and not only this one 🤦‍♂ 

I can provide such functionality and delete the postinstall script if someone helps me to clarify the author needs.

Cheers!

raineorshine:
I think the original issue was referring to dependency owners. If a dependency owner changes, there is a security risk. npm-check-updates is not a runtime dependency so does not have the same risk.

I would suggest the following output if an owner changed:

```
 htmlparser2              ^3.10.1  →   ^4.0.0
 moment                   ^2.22.2  →  ^2.24.0   *owner changed*
 react-click-n-hold        ^1.0.6  →   ^1.0.7
```

I have some concern about the performance implications of a separate network request for each package. For this reason I think it should be opt-in.

stoically:
Doesn't the owner ls endpoint just give the current owners? Is there a historical endpoint? As otherwise that wouldn't reliably work, I think

raineorshine:
@stoically Indeed, we would need a reliable historical owner endpoint. Or it could be cached locally but that seems fragile.

KingDarBoja:
I don't think npm has such historical owner endpoint...

stoically:
Querying npm for a specific package version gives the maintainer information from that very version. So querying all old versions first and comparing that to the new versions might actually work.

KingDarBoja:
Sounds good to me, I will make the changes, test and share a screenshot if the result is the expected one. After that, I will close this PR and open another in order to keep clean the commit history.

stoically:
Setting `npmConfig.fullMetadata = true` might help.

KingDarBoja:
Thanks mate, that was really helpfull. 🚀

PR #621, opened 2020-01-16
Feat: repo owner changed option

### LINKED ISSUE #465: Issue a warning when repository owner is changed

KingDarBoja:
Fixes #465. Thanks @stoically for helping me with the packument stuff.

Unfortunately, I don't have any screenshot at the moment to showcase the new owner changed column, as I have no idea which npm package got author changed.

To test it out, I had to tweak this library and use verdaccio for local publish in order to check the behaviour of the packument author data between two versions.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "npm-check-updates",
-  "version": "4.0.1",
+  "version": "4.1.0",
```
stoically: Changing the package version is up to the package maintainers to decide and should never be done in a PR unless requested.

package.json
```diff
@@ -31,7 +31,8 @@
   "scripts": {
     "lint": "eslint bin/* lib test",
     "watch": "chokidar \"lib/**/*.js\" -c \"npm run test\"",
-    "test": "npm run lint && mocha && mocha test/individual && if [ ! \"$TRAVIS\" ]; then snyk test; fi"
+    "test": "npm run lint && mocha && mocha test/individual && if [ ! \"$TRAVIS\" ]; then snyk test; fi",
+    "test:dev": "mocha && mocha test/individual"
```
stoically: Please don't add scripts you used while testing to your PR. It's an unrelated change.

lib/package-managers/pacotetest.js
```diff
@@ -0,0 +1,77 @@
+const _ = require('lodash');
```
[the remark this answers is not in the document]
KingDarBoja: Didn't noticed this one got pushed into the commit, sorry about that.

lib/npm-check-updates.js
```diff
@@ -97,20 +99,54 @@ function createDependencyTable() {
...
+        mappedArgs = _.cloneDeep(newArgs);
+    } else {
+        mappedArgs = _.cloneDeep(args);
+    }
+    const rows = Object.keys(mappedArgs.to).map(dep => {
+        const from = mappedArgs.from[dep] || '';
+        const to = versionUtil.colorizeDiff(mappedArgs.from[dep], mappedArgs.to[dep] || '');
+        let owner = '';
+        if (options.ownerChanged) {
+            owner = mappedArgs.owner && mappedArgs.owner[dep] ? '*owner changed*' : '';
```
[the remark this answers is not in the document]
KingDarBoja: How could that be displayed on the table? 🤔

lib/package-managers/npm.js
```diff
@@ -44,16 +44,40 @@ function parseJson(result, data) {
...
+}
+
 /**
  * Returns the value of one of the properties retrieved by npm view.
  * @param {string} packageName   Name of the package
  * @param {string} field         Field such as "versions" or "dist-tags.latest" are parsed from the pacote result (https://www.npmjs.com/package/pacote#packument)
  * @param {string} currentVersion
  * @returns {Promise}            Promised result
  */
-function viewOne(packageName, field, currentVersion) {
```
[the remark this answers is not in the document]
KingDarBoja: Done!. I admit I got cooky as I am used to Typescript stuff as enforcing types, provide everything on their onw function and then export at the module definition, etc.

lib/package-managers/pacotetest.js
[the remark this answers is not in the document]
KingDarBoja: Done! It is gone now 🚀

lib/npm-check-updates.js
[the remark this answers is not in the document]
raineorshine: I somewhat prefer `*owner changed*` as the 

Listing the actual changed owners would make sense in `--loglevel verbose` if we were to expand results beyond a single line.

lib/npm-check-updates.js
```diff
@@ -77,8 +77,8 @@ function getPackageFileName(options) {
 }
 
 function createDependencyTable() {
-    return new Table({
-        colAligns: ['left', 'right', 'right', 'right'],
+    const baseTable = new Table({
```
raineorshine: Avoid single-use variables (revert)

lib/npm-check-updates.js
```diff
@@ -97,20 +97,54 @@ function createDependencyTable() {
             middle: ''
         }
     });
+    return baseTable;
+}
+
+/**
```
raineorshine: Add descriptive text.

lib/npm-check-updates.js
```diff
@@ -97,20 +97,54 @@ function createDependencyTable() {
...
+async function getOwnerPerDependency(args, options) {
+    const packageManager = vm.getPackageManager(options.packageManager);
+    args.owner = {};
+    await Promise.all(
+        Object.keys(args.to).map(async (dep) => {
+            const from = args.from[dep] || null;
+            const to = args.to[dep] || null;
+            const ownerChanged = await packageManager.packageAuthorChanged(dep, from, to);
+            // eslint-disable-next-line require-atomic-updates
+            args.owner[dep] = ownerChanged;
```
raineorshine: Avoid mutation. Instead return a new object from each iteration of `map`.

e.g.
```js
return Object.assign({}, args, {
  owner: Object.assign({}, args.owner, {
    [dep]: ownerChanged
  }
})
```

lib/npm-check-updates.js
```diff
@@ -97,20 +97,54 @@ function createDependencyTable() {
...
     });
+    return baseTable;
+}
+
+/**
+ * @param {object} args
+ * @param {object} options
+ * @returns {Promise<object>}
+ */
+async function getOwnerPerDependency(args, options) {
```
raineorshine: This would be cleaner taking `from` and `to` and returning `owner`, rather than taking the whole `args` object when it doesn't need other fields.

lib/npm-check-updates.js
```diff
@@ -97,20 +97,54 @@ function createDependencyTable() {
...
-function toDependencyTable(args) {
+async function toDependencyTable(args, options) {
     const table = createDependencyTable();
-    const rows = Object.keys(args.to).map(dep => {
-        const from = args.from[dep] || '';
-        const to = versionUtil.colorizeDiff(args.from[dep], args.to[dep] || '');
-        return [dep, from, '→', to];
+    let mappedArgs = [];
+    if (options.ownerChanged) {
+        const newArgs = await getOwnerPerDependency(args, options);
```
raineorshine: `toDependencyTable` is simply a formatting function, so this is not an appropriate place to call `getOwnerPerDependency`. Rather, this call should probably be done in `upgradePackageDefinitions` and passed to `printUpgrades`.

lib/package-managers/npm.js
```diff
@@ -44,6 +44,30 @@ function parseJson(result, data) {
...
+ * Check if package author changed between current and next version.
+ * @param {string} packageName Name of the package
+ * @param {string} currentVersion
+ * @param {string} latestVersion
+ * @returns {Promise<boolean>} A promise that fullfills with boolean value.
+ */
+async function packageAuthorChanged(packageName, currentVersion, latestVersion) {
+    npmConfig.fullMetadata = true;
+
+    const current = versionUtil.isWildCard(currentVersion.charAt(0)) ? currentVersion.substr(1) : currentVersion;
```
raineorshine: What if `^5.0.0` is specified and the current version is `4.9.0`? It won't find anything at `result.versions['5.0.0']`.

`currentVersion` is any semver string, while `latestVersion` is a concrete version. I think you want to get the latest concrete version within `currentVersion`.

lib/npm-check-updates.js
```diff
@@ -97,20 +97,54 @@ function createDependencyTable() {
...
+        mappedArgs = _.cloneDeep(newArgs);
+    } else {
+        mappedArgs = _.cloneDeep(args);
+    }
+    const rows = Object.keys(mappedArgs.to).map(dep => {
+        const from = mappedArgs.from[dep] || '';
+        const to = versionUtil.colorizeDiff(mappedArgs.from[dep], mappedArgs.to[dep] || '');
+        let owner = '';
+        if (options.ownerChanged) {
+            owner = mappedArgs.owner && mappedArgs.owner[dep] ? '*owner changed*' : '';
```
raineorshine: The `if` statement could be simplified since you are already using a `ternary`:

```js
const owner = options.ownerChanged && mappedArgs.owner && mappedArgs.owner[dep] ? '*owner changed*' : '';
```

lib/package-managers/npm.js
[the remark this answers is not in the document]
stoically: No worries - refactoring in itself isn't a bad thing, but should be done in separate PR with appropriate reasoning; for huge refactors it might even make sense to open an issue first. Mixing feature and refactoring in one PR makes reviewing and tracing history harder.

lib/npm-check-updates.js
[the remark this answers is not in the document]
stoically: Agreed, that might make the whole thing noisy. Maybe additionally a "some owners changed while checking for updates, execute with --loglevel verbose to see the full list" at the end of the listing in case something changed?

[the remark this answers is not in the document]
stoically: On the other hand, since currently getting the owner changed information is already opt-in, wouldn't that be enough intent to actually get the owner information as well?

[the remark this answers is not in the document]
raineorshine: What about: `New owner: OWNER_NAME`? That generally still fit on one line.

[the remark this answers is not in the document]
stoically: The current owners (and hence the new owners) are pretty easy to look up on npmjs.com tho, whereas it's harder to get information about past owners. Also there can be multiple owners. Maybe "No longer owner(s): []"?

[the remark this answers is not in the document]
raineorshine: Yes, but is looking up past owners useful? The security risk is that the new owner is untrustworthy.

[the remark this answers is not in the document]
stoically: That's true. Well, personally I'd be fine if it just would print additional lines, one for old and one for new owner(s), so all information is available if you request it using `-o` anyway. But having just the new owner(s) in the same line would be equally fine for me. Might make even sense in the light of https://github.com/tjunnone/npm-check-updates/issues/570#issuecomment-573380990, if potentially additional information gets added like license change or changelog.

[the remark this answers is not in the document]
raineorshine: Another idea (thanks for the discussion) is `old → new` with `new` colored red.

e.g. (fabricated)

```
 htmlparser2              ^3.10.1  →   ^4.0.0  feedic → foo
 moment                   ^2.22.2  →  ^2.24.0
 react-click-n-hold        ^1.0.6  →   ^1.0.7
```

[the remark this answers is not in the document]
raineorshine: Another option that I think would be reasonable would be to add something like `--full` which would be expected to break the single-line formatting and would allow room for additional information to be added in the future, as suggested.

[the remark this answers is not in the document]
stoically: That could get interesting if multiple owners get added and removed (long line). Also there might be a problem with accessibility (color blind) if owners are only added or removed (no arrow, or arrow pointing to / originating from nothing).

But

```
htmlparser2              ^3.10.1  →   ^4.0.0

                       Owner added:  foo, bar, baz
                           removed:  baz, bar, foo

              License changed from:  MIT
                                to:  GPL

moment                   ^2.22.2  →  ^2.24.0
react-click-n-hold        ^1.0.6  →   ^1.0.7
```

isn't exactly pretty :grimacing:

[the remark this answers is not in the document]
stoically: > Another option that I think would be reasonable would be to add something like --full

So, leaving `*owner changed*` as is and showing more information with the `--full` flag?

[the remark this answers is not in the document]
raineorshine: > isn't exactly pretty 😬

That's not so bad for the full view.

> So, leaving *owner changed* as is and showing more information with the --full flag?

That's what I'm leaning towards.

lib/package-managers/npm.js
KingDarBoja: Sorry, I didn't understood what you mean.

raineorshine: Yeah, doesn't make sense to me either now that I read it :).

I think you are mixing up semver ranges with semver versions. If `currentVersion` specifies a version range, say, `^5.0.0`, then `current` will be set to `5.0.0`. However, there is not necessarily an existing version that corresponds to `5.0.0`. The latest may be `4.9.0`. So later  `result.versions[current]` may return `undefined`.

I don't think you can index into `result.versions` with a version range. Versions can only satisfy a version range.

KingDarBoja: Ah, I see what you mean, I should try to fetch the latest version by using the semver utils instead of a fixed string which, in this case, would not exist as version.

Not sure how to tackle that thing if something like that happens.

raineorshine: I can't offer a solution as I am working on other things, but I'm happy to review anything that you come up with.

KingDarBoja: I got lazy these months but will look at it this weekend 😆

KingDarBoja: I made a slighty refactor for this code block and tested on replit: [packageAuthorChanged test](https://repl.it/repls/SiennaWiseLivecd). I think this should cover those cases based on the semver ranges for both `current` and `latest`.

raineorshine: Thanks! I will review.

lib/package-managers/npm.js
```diff
@@ -44,6 +44,30 @@ function parseJson(result, data) {
...
+    npmConfig.fullMetadata = true;
+
+    const result = await pacote.packument(packageName, npmConfig);
+    if (result.versions) {
+        let pkgVersions = Object.keys(result.versions);
+        const current = semver.maxSatisfying(pkgVersions, currentVersion);
+        const latest = semver.maxSatisfying(pkgVersions, latestVersion);
+        if (current && latest) {
+            const currentAuthor = result.versions[current].author;
+            const latestAuthor = result.versions[latest].author;
```
raineorshine, line 64: This appears so simply return the `author` field in the `package.json`. If a malicious user gained control of the repo, they probably would not update the `author` field.

What about the `_npmUser` field? Would that be more secure?

lib/package-managers/npm.js
```diff
@@ -44,6 +44,30 @@ function parseJson(result, data) {
...
+ * @param {string} latestVersion
+ * @returns {Promise<boolean>} A promise that fullfills with boolean value.
+ */
+async function packageAuthorChanged(packageName, currentVersion, latestVersion) {
+    npmConfig.fullMetadata = true;
+
+    const result = await pacote.packument(packageName, npmConfig);
+    if (result.versions) {
+        let pkgVersions = Object.keys(result.versions);
+        const current = semver.maxSatisfying(pkgVersions, currentVersion);
```
raineorshine, line 60: I think you want to use `currentVersion` as-is rather than getting `maxSatisfying`. If the owner changed on a patch version in between `currentVersion` and `current`, it would not detect the owner change.

bin/ncu
```diff
@@ -27,6 +27,7 @@ program
     .option('--engines-node', 'upgrade to version which satisfies engines.node range')
     .option('-f, --filter <matches>', 'include only package names matching the given string, comma-or-space-delimited list, or /regex/')
     .option('-g, --global', 'check global packages instead of in the current project')
+    .option('-o, --ownerChanged', 'check if package owner changed and show as additional column')
```
raineorshine, line 30: Please alphabetize and add to README.

lib/npm-check-updates.js
```diff
@@ -103,14 +103,17 @@ function createDependencyTable() {
  * @param {object} args
  * @param {object} args.from
  * @param {object} args.to
+ * @param {object} args.owner
+ * @param {object} options
  * @returns {Table}
  */
-function toDependencyTable(args) {
+function toDependencyTable(args, options) {
```
raineorshine, line 110: I believe `args.owner[dep]` will be `undefined` if `options.ownerChanged` is `false`, so there is no need to check it again here. Thus you can avoid having to pass the options object to `toDependencyTable`.

lib/package-managers/npm.js
KingDarBoja: You got a point, better be safe and check npm user instead

lib/package-managers/npm.js
```diff
@@ -48,6 +48,31 @@ function parseJson(result, data) {
...
+ * @returns A promise that fullfills with boolean value.
+ */
+async function packageAuthorChanged(packageName, currentVersion, latestVersion) {
+  npmConfig.fullMetadata = true
+
+  const result = await pacote.packument(packageName, npmConfig)
+  if (result.versions) {
+    const pkgVersions = Object.keys(result.versions)
+    const current = currentVersion
+    const latest = semver.maxSatisfying(pkgVersions, latestVersion)
```
raineorshine, line 66: Since `latestVersion` is not a range, I don't think `maxSatisfying` is a NOOP here.

KingDarBoja: What do you mean? The `maxSatisfying` takes as first argument an array of the available versions and the second argument either a `Range` or a `string` specifying the version to satisfy.

KingDarBoja:
Tested against mocha with current version being `^7.2.0` and latest being `8.0.1` and checked all version list `_npmUser` to see what is going on. Noticed that Mocha hasn't `_npmUser` on any version until now, so the owner in theory never changed but since they do not specify it, probably gonna return *unknown* as requested.

lib/package-managers/npm.js
KingDarBoja: Thinking about the bug after removing the `maxSatisfying`, I tested the counterpart function (`minSatisfying`) and I think it does what I wanted for both **current** and **latest** version:

> Return the lowest version in the list that satisfies the range, or null if none of them do.

So if Mocha **current** is `^7.0.1`, the minimum version will be `7.0.1` whereas if **latest** is `8.0.0` or `^8.0.0`, it will return `8.0.0`. For cases where the version doesn't exist, i.e. **current** being `8.2.1`/ `^8.2.1` and **latest** being `9.0.0`, it will fail as `minSatisfying` will be unable to find a version to match these.

KingDarBoja: One image is better than 1000 words.

ncu-owner-check

lib/package-managers/npm.js
raineorshine: Sorry, I meant I *think* `maxSatisfying` is a NOOP.

Here is the definition of `maxSatisfying`:

> `maxSatisfying(versions, range)`: Return the highest version in the list that satisfies the range, or null if none of them do.

Since `latestVersion` is a concrete version as opposed to a range, won't `semver.maxSatisfying(pkgVersions, latestVersion)` always return `latestVersion`?

raineorshine:
I'm sort of following. Do you need assistance or do you have a solution? I can try to understand better if you are in need of support.

lib/package-managers/npm.js
KingDarBoja: Tested and you are right, it is the same version. Also there is a big downside: if the latest provided version doesn't exist (which I guess should never happen), it returns `null` because can't find a satisfying version.

raineorshine: I think it's safe to assume `latestVersion` will always exist.

raineorshine:
That should work. Do you think we should print something like `*unknown*` in those cases where no comparison can be made? Otherwise it might be possible to have a false positive.

KingDarBoja:
> That should work. Do you think we should print something like `*unknown*` in those cases where no comparison can be made? Otherwise it might be possible to have a false positive.

Of course, that's why I am returning null, to check if it is null before displaying at the table. If it is null, then show *unknown* :) that's atleast what I have in mind.

raineorshine:
Great, that sounds good.

Note that you will also need to check `result.versions[current] && result.versions[latest]`.

KingDarBoja:
> Great, that sounds good.
> 
> Note that you will also need to check `result.versions[current] && result.versions[latest]` otherwise we get the `Cannot read property '_npmUser' of undefined` error.

Too bad we can't use the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here 😢

raineorshine:
Thanks! I will review this week.

raineorshine:
Okay, I believe it's ready to merge!

I just made some small changes in 7f08648:
- Only print ownerChanged status when `--ownerChanged` is specified
- Fix issue with `getOwnerPerDependency` that was not returning the correct object
- Minor refactor, renames, convert to functional style

lib/npm-check-updates.js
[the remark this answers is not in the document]
raineorshine: @stoically I'm going to get this PR merged as-is, but I'm open to enhanced functionality in a separate PR!

lib/package-managers/npm.js
raineorshine: I can confirm that `currentVersion` is a range, not an exact version as I suggested in my initial comment.

I changed it to use `minSatisfying` on `currentVersion` and `maxSatisfying` on `upgradedVersion`. This ensures that the broadest change is covered in case the owner changed anywhere between the two.

KingDarBoja:
Finally! After few months and some lazyness from me, this will be merged 🎉

PR #625, opened 2020-02-05
Npm updates

brettz9:
This updates npm deps. and devDeps, including a couple vulnerabilities, one in `update-notifier`, and another in `tree-kill`.

I also switched to lint all files by default (minus the `node_modules` referenced in `.eslintignore`) if that's ok.


raineorshine:
Thanks, looks good to me!

Issue #634, opened 2020-03-05
 FetchError: network timeout


retog:
- [X] I have read the list of [known issues](https://github.com/tjunnone/npm-check-updates#known-issues) before filing this issue.
- [X] I have searched for [similiar issues](https://github.com/tjunnone/npm-check-updates/issues) before filing this issue.

* node version: 12.15.0
* npm version: 6.13.4
* npm-check-updates version: 4.0.3

---------------------------
 Even with a huge global timeout ` npm-check-updates` fails with a network timeout. The timeout isn't always at the same package and happens even though the reource can be retrieved without issue with curl.

```
npm-check-updates --timeout 600000
Checking /mnt/c/Users/me/Documents/plazi/synospecies/package.json
[==================--] 46/50 92%/usr/local/lib/node_modules/npm-check-updates/lib/npm-check-updates.js:412
    throw err;
    ^

FetchError: network timeout at: https://registry.npmjs.org/mustache
    at Timeout.<anonymous> (/usr/local/lib/node_modules/npm-check-updates/node_modules/minipass-fetch/lib/index.js:89:18)
    at listOnTimeout (internal/timers.js:531:17)
    at processTimers (internal/timers.js:475:7) {
  code: 'FETCH_ERROR',
  errno: 'FETCH_ERROR',
  type: 'request-timeout'
}
```

raineorshine:
Looks like pacote is timing out:

```hs
$ npm ls minipass-fetch
npm-check-updates@4.0.3 /Users/raine/projects/npm-check-updates
└─┬ pacote@11.1.0
  ├── minipass-fetch@1.2.1 
  └─┬ npm-registry-fetch@8.0.0
    ├─┬ make-fetch-happen@8.0.2
    │ └── minipass-fetch@1.2.1  deduped
    └── minipass-fetch@1.2.1  deduped
```

Maybe npm-check-updates is not setting the timeout on pacote properly?

raineorshine:
Published in `v4.0.4`.

retog:
Thanks @raineorshine for acting so quickly. Unfortunately I'm still getting the errror:

```
$ npm-check-updates -v
4.0.4
$ npm-check-updates --timeout 600000
Checking /mnt/c/Users/me/Documents/plazi/synospecies/package.json
[==================--] 46/50 92%/usr/local/lib/node_modules/npm-check-updates/lib/npm-check-updates.js:412
    throw err;
    ^

[...]

PR #640, opened 2020-03-24
fixed bug: timeout wasn't passed to pacote, also check requests are now batched using pMap

abnud1:
#634

### DIFF
--- a/lib/package-managers/npm.js
+++ b/lib/package-managers/npm.js
@@ -70,7 +70,7 @@ function viewMany(packageName, fields, currentVersion, {timeout} = {}) {
 
     npmConfig.fullMetadata = _.includes(fields, 'time');
 
-    return pacote.packument(packageName, Object.assign({}, npmConfig, timeout)).then(result =>
+    return pacote.packument(packageName, Object.assign({}, npmConfig, {timeout})).then(result =>
         fields.reduce((accum, field) => Object.assign(
             accum,
             {
--- a/lib/versionmanager.js
+++ b/lib/versionmanager.js
@@ -256,7 +256,7 @@ function upgradePackageDefinitions(currentDependencies, options) {
         loglevel: options.loglevel,
         enginesNode: options.enginesNode,
         timeout: options.timeout,
-        batchSize: options.requestsBatchSize
+        concurrency: options.concurrency
     }).then(latestVersions => {
 
         const upgradedDependencies = upgradeDependencies(currentDependencies, latestVersions, {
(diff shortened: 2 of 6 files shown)


lib/versionmanager.js
```diff
@@ -456,7 +457,7 @@ function queryVersions(packageMap, options = {}) {
         });
     }
 
-    return Promise.all(packageList.map(getPackageVersionProtected))
+    return pMap(packageList,getPackageVersionProtected,{concurrency: os.cpus().length})
```
raineorshine: I believe the CPU is not the bottleneck on HTTP requests.

> A key feature of node compared to some other systems is that the cost of holding open an idle connection is very low.  I've written tests that open up 64K connections, and once the connections are established, the CPU is obvious idle while no work is taking place, but the resident memory is under 200MB.

https://www.quora.com/How-many-connections-can-a-node-js-server-handle

Instead, I would suggest providing a reasonable default of `10`, and allow that to be overridden by a cli option.

lib/package-managers/npm.js
```diff
@@ -70,7 +70,7 @@ function viewMany(packageName, fields, currentVersion, {timeout} = {}) {
 
     npmConfig.fullMetadata = _.includes(fields, 'time');
 
-    return pacote.packument(packageName, Object.assign({}, npmConfig, timeout)).then(result =>
+    return pacote.packument(packageName, Object.assign({}, npmConfig, {timeout})).then(result =>
```
raineorshine: Aha! That was the issue. Thanks.

lib/versionmanager.js
abnud1: I searched the internet But I didn't find any good recommendation on how to decide the requests concurrency, the only thing I found is this old post:
https://www.physicsforums.com/threads/idm-how-to-increase-downloading-thread-number.875706/post-5501364

Which says that it should be at most 8(multi-threaded download managers almost have the same problem as ours).

I'm thinking of setting it to 8 without providing a cli option, I don't think any user is interested in changing this number, in fact I didn't know ncu issues requests in parallel until I looked at code, before that I thought it issues one request at a time.

raineorshine: Yes, `8` should be fine. We are fetching from a known source, the `npm` registry, so it is less of a mystery how many connections they support (many), but I still think it's a good idea to batch requests.

raineorshine: I do ask that this be added as a cli option. I am not willing to hard-code it into the library.

abnud1: finished, I made 8 the default and added cli option called `--requests-batch-size`, feel free to change its name and description.

raineorshine:
Thank you!

raineorshine:
Published `v4.1.0`

PR #642, opened 2020-03-25
fix: declare namespace and remove async keyword

adelkahomolova:
Building my project is failing because of issues in d.ts file. See the print screen. This PR should fix it.

Remove async - The fact that it returns a Promise is what makes it of async type.
Add declare keyword - Top-level declarations in .d.ts files must start with either a 'declare' or 'export' modifier. Export can't be used because the ncu is exported.

Screenshot 2020-03-25 at 09 21 19

npm-check-updates: 4.0.5
Node: 12.13.0
YARN: 1.21.1
NPM: 6.12.0
TypeScript: 3.8.3

### DIFF
--- a/lib/npm-check-updates.d.ts
+++ b/lib/npm-check-updates.d.ts
@@ -1,4 +1,4 @@
-namespace ncu {
+declare namespace ncu {
   interface RunOptions {
     /**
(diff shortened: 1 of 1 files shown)


raineorshine:
Thank you!

raineorshine:
Published in `v4.0.6`

Issue #146, opened 2015-10-05
ncu -g doesn't work on homebrew or windows


cebor:
`ncu -g` always says that my deps are up to date, even when they are not!

node: 4.1.1 (brew)
os: osx 10.11

```bash
$ ncu -g

All global packages are up-to-date :)
```

On my ubuntu machine all works fine!

metaraine:
Thanks for reporting. Will you post your global packages? (`npm -g ls --depth=0 `)

cebor:
```bash
➜  ~  npm -g ls --depth=0
/usr/local/lib
├── bower@1.5.3
├── generator-angular@0.12.1
├── generator-gulp@0.2.0
├── generator-karma@1.0.0
├── grunt-cli@0.1.13
├── npm@2.14.8
├── npm-check-updates@2.3.0
├── strongloop@5.0.1
└── yo@1.4.8

 ➜  ~  ncu -g

All global packages are up-to-date :)
```

wickedsp1d3r:
I have the same problem.

osahner:
same here.

```
npm -g list --dept=0
/usr/local/lib
...
├── clean-css@3.4.5
...
├── npm@3.3.9
├── npm-check-updates@2.3.3
...
```

```
npm show clean-css version
3.4.6
```

```
ncu -g

All global packages are up-to-date :)

```



```
uname -a
Darwin ***.local 15.0.0 Darwin Kernel Version 15.0.0: Sat Sep 19 15:53:46 PDT 2015; root:xnu-3247.10.11~1/RELEASE_X86_64 x86_64

brew --version
0.9.5 (git revision 91b03; last commit 2015-10-21)
```

metaraine:
Unfortunately I cannot reproduce on my machine:

```
$ ncu -g

 clean-css   3.4.5  →   3.4.6 
 npm        2.14.4  →   3.3.9 

$ uname -a
Darwin ***.local 13.1.0 Darwin Kernel Version 13.1.0: Wed Apr  2 23:52:02 PDT 2014; root:xnu-2422.92.1~2/RELEASE_X86_64 x86_64

```

cebor:
i have this problem, since upgraded to node 4.2 from 0.12.

daleee:
Similar issues here:
```$ npm -g ls --depth=0
/usr/local/lib
├── babel@5.8.23
├── bower@1.5.3
├── browserify@11.2.0
├── eslint@1.6.0
[...]

PR #643, opened 2020-03-31
Fixes an issue which occurs if prefix is different than the npm default one

nhristov:
Trims since `npm config set prefix` appends a new line at the end of .npmrc This should probably be done for homebrew, as well, but I'm using it with Windows and am unaware of how it behaves under macOS, therefore, I don't know if the issue even occurs there. It most probably does, but if anybody can confirm it before I submit the second pull request, that be great.

To reproduce set npm prefix to a different folder and install a global package. After that run `ncu -g` and if you get `No dependencies.` then it does.

### DIFF
--- a/lib/package-managers/npm.js
+++ b/lib/package-managers/npm.js
@@ -152,7 +152,7 @@ function defaultPrefix(options) {
             // Workaround: get prefix on windows for global packages
             // Only needed when using npm api directly
             process.platform === 'win32' && options.global && !process.env.prefix ?
-                `${process.env.AppData}\\npm` :
+                prefix ? prefix.trim() : `${process.env.AppData}\\npm` :
                 null;
     });
 }


raineorshine:
Thanks! We just need to be sure that this doesn't cause a regression, especially related to #146.

raineorshine:
> To reproduce set npm prefix to a different folder and install a global package. After that run `ncu -g` and if you get `No dependencies.` then it does.

On OSX I change the npm prefix, install a global package, and ncu `v4.1.0` correctly checks for new updates for the installed package.

nhristov:
No, it doesn't. I took a look at the previous issues before opening this PR and I've also performed a test without modifying prefix. The change I made just checks if prefix is available to the context of spawn and uses that and if it's not it falls back to the previous behavior with `${process.env.AppData}\\npm`

raineorshine:
Published in `v4.1.1`

PR #645, opened 2020-04-07
Fix `reject` type

garyking:
Minor fix, for the `reject` type.


raineorshine:
Published in `v4.1.2`.

PR #654, opened 2020-05-15
Update repo URL

XhmikosR:
Sorry for not splitting these, but checking the [non-whitespace diff](https://github.com/raineorshine/npm-check-updates/pull/654/files?w=1) and the rich diffs should be enough 🙂 If not please let me know.

EDIT: I decided to split the patches so that's it's cleaner


raineorshine:
Thanks so much

PR #656, opened 2020-05-15
README: fix badges

XhmikosR:
* use the npm link directly since it's faster
* switch to travis-ci.com since that's what's currently being used

Requires #654 to be merged first.


XhmikosR:
@raineorshine if we get this and #657 merged then I'll be able to open the GH Actions CI PR 🙂

PR #659, opened 2020-05-15
Delete the unused appveyor.yml

XhmikosR:
BTW we could try to use GitHub Actions CI and test both Linux and Windows.

I could give it a go later if you are interested.


raineorshine:
> BTW we could try to use GitHub Actions CI and test both Linux and Windows.
> 
> I could give it a go later if you are interested.

Sure! I've had some trouble reproducing bugs on Windows so that would be helpful.

PR #665, opened 2020-05-15
Lint tweaks.

XhmikosR:
* fix indentation in lib/npm-check-updates.d.ts
* use Object.spread when possible
* use `Number.parseInt`
* add `prefer-const`
* simplify string.slice by replacing `length - 1` with `-1` for its second param
* throw a `TypeError` when the type isn't the expected one
* use `path.join` to join paths
* use `should.equal` for booleans too
* remove unneeded ESLint suppression

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,14 @@
 {
+    "env": {
+        "node": true,
+        "mocha": true,
+        "es6": true
+    },
+    "extends": [
+        "eslint:recommended"
+    ],
     "parserOptions": {
-        "ecmaVersion": 8
+        "ecmaVersion": 2018
     },
     "rules": {
         "indent": [
@@ -101,6 +109,8 @@
             2,
             "always"
         ],
+        "prefer-const": 2,
+        "prefer-spread": 2,
         "no-throw-literal": 2,
         "no-eval": 2,
         "no-unneeded-ternary": 2,
@@ -117,11 +127,5 @@
         "no-trailing-spaces": 2,
         "camelcase": 2,
         "no-undef": 2
-    },
-    "env": {
-        "node": true,
-        "mocha": true,
-        "es6": true
-    },
-    "extends": ["eslint:recommended"]
+    }
 }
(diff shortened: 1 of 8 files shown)


XhmikosR:
BTW most of these were found with xo. Not sure if you want to use that, if so let me know. I have a WIP branch adapted to match the current style mostly.

Also, it seems this `await` is redundant? https://github.com/raineorshine/npm-check-updates/pull/665/files#diff-1415f5139deb67a37325f1bde416bcbdR468

raineorshine:
> BTW most of these were found with xo. Not sure if you want to use that, if so let me know. I have a WIP branch adapted to match the current style mostly.

The current code style was inherited from the original project owner many years ago. I was thinking it's about time to update it. I actually use an [extended version of Standard style](https://github.com/cybersemics/em/blob/dev/.eslintrc.json). I will probably copy the relevant rules over from that.

> Also, it seems this `await` is redundant? [#665 (files)](https://github.com/raineorshine/npm-check-updates/pull/665/files#diff-1415f5139deb67a37325f1bde416bcbdR468)

Do you have a line number? I didn't see `await` in the diff.

> EDIT: it seems the `bin` files are not linted because they don't have an extension. We could explicitly specify them or we could rename them to .js. I prefer the latter, but your call. 🙂

Sure, rename works.

Thanks again for your involvement and help.

XhmikosR:
> The current code style was inherited from the original project owner many years ago. I was thinking it's about time to update it. I actually use an [extended version of Standard style](https://github.com/cybersemics/em/blob/dev/.eslintrc.json). I will probably copy the relevant rules over from that.

Never used it but this is quite objective, so as long as it's better :)

> Do you have a line number? I didn't see `await` in the diff.

https://github.com/raineorshine/npm-check-updates/blob/9e48048ed7216d970610f38560763c6afc00502a/lib/npm-check-updates.js#L467

>Sure, rename works.

I'll do the rename in another PR tomorrow.

>Thanks again for your involvement and help.

NP, I like giving back to projects I use 🙂

PR #666, opened 2020-05-16
Assorted tweaks

XhmikosR:
* Split mocha test script so that one can run it separately
* Change the env variable for tests to `NCU_TESTS`
* Add js extension in bin files
* Update .gitignore to ignore test-generated files
* test: move timeout to the parent `describe()` (BTW it seems the timeout for individual tests could be decreased AFAICT)
* Tighten mocha options
* Remove the unused lib/raw-promisify.js
* bin/ncu.js: simplify version command

@raineorshine About the env var, I assume you only wanted the conditional for ncu's own tests, right? Because `NODE_ENV` is a common env var and could be set by others accidentally.


bin/ncu.js
```diff
@@ -65,7 +65,7 @@ const {configFileName, configFilePath, packageFile} = program;
 // NOTE: Do not load .ncurc from project directory when tests are running
 // Can be overridden if configFilePath is set explicitly
 let rcArguments = [];
-if (process.env.NODE_ENV !== 'test' || configFilePath) {
+if (!process.env.NCU_TESTS || configFilePath) {
```
XhmikosR, line 68: I could change these to `=== true` if you prefer that :)

test/test-ncu.js
```diff
@@ -8,7 +8,7 @@ const tmp             = require('tmp');
 chai.use(require('chai-as-promised'));
 chai.use(require('chai-string'));
 
-process.env.NODE_ENV = 'test';
+process.env.NCU_TESTS = true;
```
raineorshine, line 11: Actually, I think `NODE_ENV` is what we want here. Jest itself sets `NODE_ENV = 'test'` so there's some good precedent for using that variable to identify a test environment.

raineorshine:
> @raineorshine About the env var, I assume you only wanted the conditional for ncu's own tests, right? Because `NODE_ENV` is a common env var and could be set by others accidentally.

I see. So if someone is running `ncu` as part of a CI setup for example, `NODE_ENV` may be set to `test` but in that case `ncu` should not definitely not ignore their `.ncurc` file.

PR #669, opened 2020-05-18
Replace some lodash methods with native

XhmikosR:
Some notes:

* Since some lodash methods work with undefined/null while the native methods don't, I had to adapt `doesSatisfyEnginesNode()` because apparently `versions` could be undefined, at least in some of the tests
* ~~I'm not 100% sure that `cint.partialAt` is 100% the same as `_.partialRight` so please confirm since you are the author of cint 🙂~~ I dropped that patch for this PR
* The plan isn't to get rid of lodash or cint for that matter; the plan is to use native methods when possible without sacrificing FP
* Could you maybe have a look and see if we can replace cint with lodash in the codebase? I know you are the author of cint but I assume it's not actively maintained even though it works, and less dependencies is always better as long as one does not reinvent the wheel.

BTW have you thought about adding coverage?

### DIFF
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -313,10 +313,8 @@ function printUpgrades(options, {current, upgraded, numUpgraded, total}) {
 
 /** Initializes and consolidates options from the cli. */
 function initOptions(options) {
-
-    const json = _(options)
-        .keys()
-        .filter(_.partial(_.startsWith, _, 'json', 0))
+    const json = Object.keys(options)
+        .filter(option => option.startsWith('json'))
         .some(_.propertyOf(options));
 
     return {
--- a/lib/npm-check-updates.js
+++ b/lib/npm-check-updates.js
@@ -420,12 +420,13 @@ process.on('unhandledRejection', err => {
 async function run(options = {}) {
     // if not executed on the command-line (i.e. executed as a node module), set some defaults
     if (!options.cli) {
-        options = _.defaults({}, options, {
+        const defaultOptions = {
             jsonUpgraded: true,
             loglevel: 'silent',
             packageManager: 'npm',
             args: []
-        });
+        };
+        options = {...defaultOptions, ...options};
     }
 
     options = initOptions(options);
--- a/lib/package-managers/npm.js
+++ b/lib/package-managers/npm.js
@@ -70,7 +70,7 @@ function viewMany(packageName, fields, currentVersion, {timeout} = {}) {
         return Promise.resolve({});
     }
 
-    npmConfig.fullMetadata = _.includes(fields, 'time');
+    npmConfig.fullMetadata = fields.includes('time');
 
     return pacote.packument(packageName, {...npmConfig, timeout}).then(result =>
         fields.reduce((accum, field) => Object.assign(
@@ -219,7 +219,7 @@ module.exports = {
             .then(result => {
                 const versions = doesSatisfyEnginesNode(result.versions, options.enginesNode);
                 return _.keys(result.time).reduce((accum, key) =>
-                    accum.concat(_.includes(TIME_FIELDS, key) || _.includes(versions, key) ? key : []), []
+                    accum.concat(TIME_FIELDS.includes(key) || versions.includes(key) ? key : []), []
                 );
             })
             .then(_.partialRight(_.pullAll, TIME_FIELDS))
--- a/lib/package-managers/npm.js
+++ b/lib/package-managers/npm.js
@@ -99,14 +99,17 @@ function filterOutPrereleaseVersions(versions, pre) {
  * @returns {Array} An array of versions which satisfies engines.node range
  */
 function doesSatisfyEnginesNode(versions, enginesNode) {
+    if (!versions) {
+        return [];
+    }
     if (!enginesNode) {
-        return _.keys(versions);
+        return Object.keys(versions);
     }
     const minVersion = _.get(semver.minVersion(enginesNode), 'version');
     if (!minVersion) {
-        return _.keys(versions);
+        return Object.keys(versions);
     }
-    return _.keys(versions).filter(version => {
+    return Object.keys(versions).filter(version => {
         const versionEnginesNode = _.get(versions[version], 'engines.node');
         return versionEnginesNode && semver.satisfies(minVersion, versionEnginesNode);
     });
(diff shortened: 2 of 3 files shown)


lib/package-managers/npm.js
```diff
@@ -216,8 +217,8 @@ module.exports = {
         return viewMany(packageName, ['time', 'versions'], currentVersion, {timeout: options.timeout})
             .then(result => {
                 const versions = doesSatisfyEnginesNode(result.versions, options.enginesNode);
-                return _.keys(result.time).reduce((accum, key) =>
-                    accum.concat(_.includes(TIME_FIELDS, key) || _.includes(versions, key) ? key : []), []
+                return Object.keys(result.time || {}).reduce((accum, key) =>
```
XhmikosR, line 220: I was getting `result.time` undefined in some of the current tests hence the fallback.

lib/npm-check-updates.js
```diff
@@ -420,12 +418,14 @@ process.on('unhandledRejection', err => {
 async function run(options = {}) {
     // if not executed on the command-line (i.e. executed as a node module), set some defaults
     if (!options.cli) {
-        options = _.defaults({}, options, {
+        const defaultOptions = {
+            ...options,
```
raineorshine, line 422: The `defaultOptions` should not include `options` since it is merged in the next statement.

XhmikosR: Oops, typo, will fix ASAP.

XhmikosR:
>     * Could you maybe have a look and see if we can replace cint with lodash in the codebase? I know you are the author of cint but I assume it's not actively maintained even though it works, and less dependencies is always better as long as one does not reinvent the wheel.

I think only this is left. There aren't many instances and I think we can probably try to use lodash. cint isn't a big dependency but the less the better, as long as we don't reinvent the wheel :)

raineorshine:
> * The plan isn't to get rid of lodash or cint for that matter; the plan is to use native methods when possible without sacrificing FP

Yup, sounds good.

> * Could you maybe have a look and see if we can replace cint with lodash in the codebase? I know you are the author of cint but I assume it's not actively maintained even though it works, and less dependencies is always better as long as one does not reinvent the wheel.

That's not a bad idea, although I have an argument against it.

Here are the `cint` functions that are used in `npm-check-updates`:

- `cint.aritize`
- `cint.filterObject`
- `cint.index`
- `cint.keyValue`
- `cint.mapObject`
- `cint.not`
- `cint.partialAt`
- `cint.toArray`
- `cint.toObject`

Some of them may be easy to replace, while others are cleaner than the lodash equivalent. If we wrapped the lodash expressions in functions with equivalent interfaces, we would still have the question of where to put them. I would not want to inline them, as I value modularity more than minimizing dependencies. `cint` has 0 dependencies itself, which is about as light as it gets, and it has working unit tests. All things considered, it seems like a reasonable import, even if it is old and calcified :).

> 
> BTW have you thought about adding coverage?

I haven't, but I am open to suggestions.

XhmikosR:
I think as a first step we should try to use native methods where possible. Then we should have a clear picture and if cint has lodash equivalents, we can go the other way around, maybe?

> I haven't, but I am open to suggestions.

I could try adding nyc and you could add the repo on coverage? I can make a PR if you want.

XhmikosR:
BTW it seems lodash is also a dependency of another dependency (node-alias). We should find a way to do this without node-alias 🙂

raineorshine:
> I think as a first step we should try to use native methods where possible. 

Yes, Javascript certainly has much more available than it did when I wrote `cint` :).

> Then we should have a clear picture and if cint has lodash equivalents, we can go the other way around, maybe?

Yes, but only if there are equivalents. If the lodash expression is more complex, I would prefer the simpler interface.

> > I haven't, but I am open to suggestions.
> 
> I could try adding nyc and you could add the repo on coverage? I can make a PR if you want.

Sure, would love to. Can you explain what you mean by "add the repo on coverage"?

raineorshine:
> BTW it seems lodash is also a dependency of another dependency (node-alias). We should find a way to do this without node-alias 🙂

It looks like `node-alias` is just [55 lines of code](https://github.com/justinhelmer/node-alias/blob/master/index.js) with three references to lodash which are easily replaced. I might suggest a PR, and if the developer has moved on we can fork.

XhmikosR:
> Yes, but only if there are equivalents. If the lodash expression is more complex, I would prefer the simpler interface.

Agreed. It seems some methods currently used are interchangeable, so it would be nice to at least streamline this :) For example, we use `_.last` and `cint.index`. But anyway, you should know this stuff better since you wrote cint, so I'll let you handle this step.

>Sure, would love to. Can you explain what you mean by "add the repo on coverage"?

Typo, I meant coveralls :) I'll make a new PR and ping you

> It looks like `node-alias` is just [55 lines of code](https://github.com/justinhelmer/node-alias/blob/master/index.js) with three references to lodash which are easily replaced. I might suggest a PR, and if the developer has moved on we can fork.

Can't we just require ncu directly? Not sure if we need another package is all I'm saying :)

raineorshine:
> > Yes, but only if there are equivalents. If the lodash expression is more complex, I would prefer the simpler interface.
> 
> Agreed. It seems some methods currently used are interchangeable, so it would be nice to at least streamline this :) For example, we use `_.last` and `cint.index`. But anyway, you should know this stuff better since you wrote cint, so I'll let you handle this step.

Yes, I agree. I'll see what I can do.

> 
> > Sure, would love to. Can you explain what you mean by "add the repo on coverage"?
> 
> Typo, I meant coveralls :) I'll make a new PR and ping you

Got it, that makes sense. Yes, please do!

> 
> > It looks like `node-alias` is just [55 lines of code](https://github.com/justinhelmer/node-alias/blob/master/index.js) [](/justinhelmer/node-alias/blob/HEAD@{2020-05-18T13:52:22Z}/index.js) with three references to lodash which are easily replaced. I might suggest a PR, and if the developer has moved on we can fork.
> 
> Can't we just require ncu directly? Not sure if we need another package is all I'm saying :)

Can you explain what you mean by requiring `ncu` directly? Forgive me as I am not seeing the connection yet.

XhmikosR:
> Can you explain what you mean by requiring `ncu` directly? Forgive me as I am not seeing the connection yet.

Can't we place this

```js
require('node-alias')('ncu', __dirname, {message: false});
```

with this

```js
require('./ncu');
```

Or if there are path issues we could use `path.join(__dirname, './ncu')`?

raineorshine:
No, the point is that `node-alias` notifies the user with a message that tells them there is a shorter command.

XhmikosR:
But its `message` option is set to false?

raineorshine:
Oh! Good point :)

XhmikosR:
@raineorshine do you still want to keep `node-alias`? If so maybe you should add the message, otherwise remove the dep?

raineorshine:
Fixed in 9bc0b9dde215e7998b9a27d1455c80c5b845d6a7

PR #675, opened 2020-05-18
Remove bower support.

### LINKED ISSUE #622: Bower is deprecated
Since [bower is deprecated](https://www.npmjs.com/package/bower) since about a year, it might make sense to drop support for it in ncu. That would clean up the code base and make maintenance easier.

XhmikosR:
Not sure if we should remove more code. I mean, the current package managers do use package.json as far as I know.

Fixes #622


raineorshine:
That's all we need! Thanks

XhmikosR:
I guess the next version will need to be a major version bump?

raineorshine:
Yup

PR #677, opened 2020-05-18
Replace cint with lodash and native functions

raineorshine:
I found replacements for everything but the following [cint](https://github.com/raineorshine/cint/) functions:

- `cint.filterObject`
- `cint.mapObject`
- `cint.toArray`
- `cint.toObject`

@XhmikosR I don't see equivalent functions in `lodash`, so I think I will have to keep `cint` as a dependency. I prefer that over inlining the cint functions. It is better than before at least.


XhmikosR:
BTW @raineorshine that's why I suggested adapting the Coveralls threshold 😛

0.5% or half should be better.

XhmikosR:
@raineorshine you edited my message hence I wasn't notified :/

Just use 0.5% as the failure threshold and see how it goes.

About NOT requiring coveralls, this should be done in the repo branch protection settings.

raineorshine:
> @raineorshine you edited my message hence I wasn't notified :/

Oops!

> Just use 0.5% as the failure threshold and see how it goes. **EDIT**: or even bigger like 1, 2%.
> 
> About NOT requiring coveralls, this should be done in the repo branch protection settings.

I see. Done!

XhmikosR:
@raineorshine I have a couple of patches that won't apply to current master, but I could update them if you are interested: https://github.com/raineorshine/npm-check-updates/compare/master...XhmikosR:xo Ignore all patches minus the "Replace..." ones 🙂

raineorshine:
> @raineorshine I have a couple of patches that won't apply to current master, but I could update them if you are interested: [master...XhmikosR:xo](https://github.com/raineorshine/npm-check-updates/compare/master...XhmikosR:xo) Ignore all patches minus the "Replace..." ones 🙂

Thanks, but I think it's okay without them. `_.last` works fine and the `reverse` technique is a bit worse performance-wise.

XhmikosR:
Yup, that's why I hadn't made PRs. All good!

PR #684, opened 2020-06-24
Create dependabot.yml

eegeeZA:
This basic config will (ironically) keep the internal NPM dependencies up-to-date. I have left out Docker as it's configured to always use the latest Node.

https://help.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates#enabling-github-dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,4 +3,4 @@ updates:
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
-      interval: "daily"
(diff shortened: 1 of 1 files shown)


raineorshine:
Thanks for the proposal! Can you provide a justification for this? ncu's dependencies get updated whenever a new feature or bug is fixed. I haven't had any complaints about this, so it seems to be working for people. I'm not sure I see the value in getting notifications all the time about updates.

eegeeZA:
There may be times where bug fixes or development is unnecessary for a few months. During those times, it would be good to have an automated process create notifications for security updates and new features. The frequency of these checks can also be configured to weekly or monthly to suit your preferences, The pull requests can also be ignored as newer PRs will close the old ones.

stoically:
I'd really wish if there were a middle ground between "spamming PRs" (imho it's still spam if it happens once a month) and "you see exactly which dep upgrade might break things" (because CI runs for every update).

raineorshine:
I'd be okay with monthly.

Issue #707, opened 2020-07-29
Disable Strict-SSL not working


Lalem001:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [x] I am using the latest version of `npm-check-updates`
- [x] I am using `node >= 10.17`

---------------------------

## Steps to Reproduce
- Private Registry setup with self signed cert.
  - Yes I have seen other issues about this.
  - This issue is not about working with such certs, but ignoring them as follows.
- Set `strict-ssl = false` to npm's config (per [npm config docs](https://docs.npmjs.com/using-npm/config#strict-ssl) snake case is required)
- Attempt NCU

## Current Behavior
- NCU fails with `self signed certificate in certificate chain` as mentioned in other issues
- I have manually run the code listed in [npm.js#L15-24](https://github.com/raineorshine/npm-check-updates/blob/master/lib/package-managers/npm.js#L15-L24)
  - Parsed config passed to `pacote`, and subsequently `npm-registry-fetch`, has `strict-ssl: false`
  - `npm-registry-fetch` is expecting `strictSSL` not `strict-ssl`
  - Setting `strictSSL = true` in npm config, while invalid, stops the self signed error

## Expected Behavior
- NCU should correct npm config keys to match what `pacote` and `npm-registry-fetch` are expecting.
  - At least `strict-ssl` key converted to `strictSSL`

@raineorshine I can make a PR for this if you wish

Lalem001:
Just found that this is technically a regression of #528

raineorshine:
Thank you for identifying the problem!

> @raineorshine I can make a PR for this if you wish

This would be SO helpful! I have been unable to reproduce, and I know it is affecting many people.

raineorshine:
Fixed and published in `v7.0.3-alpha.1`.

PR #713, opened 2020-08-03
Convert npm config to pacote config

Lalem001:
Here is my first pass at resolving #707. 

Handles the following cases:
- Keys that are converted to camelCase
- `strict-ssl` -> `strictSSL`
- Parses `cafile` and sets result as `npmConfig.ca`
  - npm does this internally when it references the same config
  - `cafile` is not handled in pacote/npm-registry-fetch
  - caveat: if `ca` is specified AFTER `cafile`, `ca` will overwrite the parsed result (npm parses `cafile` last)

I did what I could to cross reference [npm config](https://docs.npmjs.com/misc/config), [pacote docs](https://www.npmjs.com/package/pacote), and [npm-registry-fetch](https://www.npmjs.com/package/npm-registry-fetch).  

> Note: Pacote does reference [cacache](https://github.com/npm/cacache) in its opts docs. I am not sure if it is relevant here.

### DIFF
--- a/lib/package-managers/npm.js
+++ b/lib/package-managers/npm.js
@@ -2,6 +2,7 @@
 
 const _ = require('lodash')
 const cint = require('cint')
+const fs = require('fs')
 const semver = require('semver')
 const spawn = require('spawn-please')
 const pacote = require('pacote')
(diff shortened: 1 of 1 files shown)


lib/package-managers/npm.js
```diff
@@ -10,16 +11,49 @@ const versionUtil = require('../version-util')
...
+    return { ca: output }
+  },
+  'fetch-retries': 'fetchRetries',
+  'fetch-retry-factor': 'fetchRetryFactor',
+  'fetch-retry-mintimeout': 'fetchRetryMintimeout',
+  'local-address': 'localAddress',
+  'prefer-offline': 'preferOffline',
+  'prefer-online': 'preferOnline',
+  'strict-ssl': 'strictSSL',
+  'user-agent': 'userAgent',
```
raineorshine, line 33: These appear to be a regular transformation to PascalCase. Can you use https://github.com/sindresorhus/camelcase to generalize this? Are there any false positives?

`strict-ssl` might be an exception since `SSL` is capitalized, but an exception for a single key is better than making every key explicit.

Lalem001: You have lodash in this file, and it has a `camelCase` method. I was thinking that using a precomputed value was more straight forward than having `_.camelCase` listed a bunch of times. And, yes, strict-ssl would be an exception.

raineorshine: Yes, we could use lodash.

The idea would be not to have multiple places where `_.camelCase` is used, but to abstract it so that `_.camelCase` only appears once in the code, inside [libnpmconfig.read().forEach](https://github.com/raineorshine/npm-check-updates/pull/713/files#diff-036fe149039aa46377f04ef192bfe426R39) where it will automatically be called on every key. Are there any false positives that would prevent us from doing that?

Lalem001: No false positives that I can see, but there is a false negative (maxsockets -> maxSockets) that I missed in the list before. I will adjust the PR.

Lalem001: Bah, I just remembered that there are extra values that `libnpmconfig` adds to the returned config. I will need to make sure those are not affected by simply camel casing all special case keys.

Lalem001:
For reference to the key that cannot be camel cased, [Using private packages in a CI/CD workflow](https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#create-and-check-in-a-project-specific-npmrc-file)

> //registry.npmjs.org/:_authToken=${NPM_TOKEN}

raineorshine:
Thank you!

Lalem001:
> Thank you!

Glad I could help ^_^

PR #723, opened 2020-08-18
Normalize CLI files

LitoMore:
### Changes

- Drop `npm-check-updates.js`
- Rename `ncu.js` to `cli.js`
- Update `bin` section in `package.json`

The `npm-check-updates.js` and `ncu.js` are the same, we can keep only one file in our source code.

BTW, `cli.js` is a better name for the package bin file, people can know this is an executable file from its name.


raineorshine:
Agreed! Thanks!

PR #736, opened 2020-09-22
make rimraf a production dependency

faustbrian:
Installation with yarn berry fails because rimraf is referenced in build files but it isn't included as a dependency in production builds.

```
Error: npm-check-updates tried to access rimraf, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: rimraf (via "rimraf")
Required by: npm-check-updates@npm:9.0.1 (via /Users/brian/.yarn/berry/cache/npm-check-updates-npm-9.0.1-4c479406dd-2.zip/node_modules/npm-check-updates/lib/)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "prompts": "^2.3.2",
     "rc-config-loader": "^3.0.0",
     "remote-git-tags": "^3.0.0",
+    "rimraf": "^3.0.2",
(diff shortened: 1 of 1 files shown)


raineorshine:
Thanks!

raineorshine:
Published in `v9.0.2`.

PR #760, opened 2020-11-09
Update test.yml

XhmikosR:
* skip dependabot branches
* switch to Node.js 14.x for coveralls

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+    branches-ignore:
+      - "dependabot/**"
   pull_request:
     branches:
       - "**"
@@ -42,6 +44,6 @@ jobs:
 
       - name: Run Coveralls
         uses: coverallsapp/github-action@master
-        if: startsWith(matrix.os, 'ubuntu') && matrix.node == 12
+        if: startsWith(matrix.os, 'ubuntu') && matrix.node == 14
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,8 +4,7 @@ on:
   push:
     branches:
       - master
-    branches-ignore:
-      - "dependabot/**"
+      - "!dependabot/**"
   pull_request:
     branches:
       - "**"


XhmikosR:
BTW @raineorshine I'm getting two failures on Windows 10 with the native Windows cmd (CI does not use this on Windows hence why it passes there)

```
  198 passing (1m)
  1 pending
  2 failing

  1) doctor
       upgrade dependencies when tests pass:
     AssertionError: expected 'npm install\nnpm run test\n' to include '✓ Tests pass'
      at Context.<anonymous> (test\test-doctor.js:70:19)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)

  2) yarn
       "No lockfile" error should be thrown on list command when there is no lockfile:

      AssertionError: expected promise to be rejected with an error including 'No lockfile in this directory. Run `yarn install` to generate one.' but got 'Unexpected token E in JSON at position 0'
      + expected - actual

      -Unexpected token E in JSON at position 0
      +No lockfile in this directory. Run `yarn install` to generate one.
```

The first one is due to the Unicode character which unfortunately Windows does not display properly.

The second one, could it be because I don't have yarn installed?

That being said, we could change CI to use the native cmd on Windows if needed, or just ignore all this since it seems to be trivial, your call :)

XhmikosR:
Also, perhaps we should add Node.js 15 to catch any issues early?

.github/workflows/test.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+    branches-ignore:
+      - "dependabot/**"
```
raineorshine, line 8: > * skip dependabot branches

Could you explain this? Does this cause the tests to not be run on dependabot PR's?

XhmikosR: It only skip the branch builds since dependabot always opens a PR so it's duplicate runs :)

raineorshine:
> The first one is due to the Unicode character which unfortunately Windows does not display properly.

Perhaps just matching against "Tests pass" would fix it?

> The second one, could it be because I don't have yarn installed?

Yes, I see now that we need `yarn` as a `devDependency` to run the [yarn tests](https://github.com/raineorshine/npm-check-updates/blob/master/test/package-managers/yarn/index.js), and we probably need a way to mock yarn with the local version since by default it uses the system yarn. Is this something you'd be able to help with?

> That being said, we could change CI to use the native cmd on Windows if needed, or just ignore all this since it seems to be trivial, your call :)

Good question. What does the CI use on Windows instead of the native cmd? If we switch to the native cmd, will the other one still work, or do we need to run on both?

XhmikosR:
> > The first one is due to the Unicode character which unfortunately Windows does not display properly.
> 
> Perhaps just matching against "Tests pass" would fix it?

Yup, that would work.



> > The second one, could it be because I don't have yarn installed?
> 
> Yes, I see now that we need `yarn` as a `devDependency` to run the [yarn tests](https://github.com/raineorshine/npm-check-updates/blob/master/test/package-managers/yarn/index.js), and we probably need a way to mock yarn with the local version since by default it uses the system yarn. Is this something you'd be able to help with?

Unfortunately, I don't have a lot of time but it's not a huge deal :) I just wasn't aware and expected tests to still be green on Windows.

> > That being said, we could change CI to use the native cmd on Windows if needed, or just ignore all this since it seems to be trivial, your call :)
> 
> Good question. What does the CI use on Windows instead of the native cmd? If we switch to the native cmd, will the other one still work, or do we need to run on both?

CI is using powershell by default on Windows: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell
So, it depends on what you want to test. I personally still use the default cmd a lot hence why I hit the issue, but I guess most developers on Windows do run some kind of emulation. So, I'd say just leave it as is until it comes up again :)

raineorshine:
> > > The second one, could it be because I don't have yarn installed?
> > 
> > 
> > Yes, I see now that we need `yarn` as a `devDependency` to run the [yarn tests](https://github.com/raineorshine/npm-check-updates/blob/master/test/package-managers/yarn/index.js?rgh-link-date=2020-11-09T15%3A34%3A37Z), and we probably need a way to mock yarn with the local version since by default it uses the system yarn. Is this something you'd be able to help with?
> 
> Unfortunately, I don't have a lot of time but it's not a huge deal :) I just wasn't aware and expected tests to still be green on Windows.

Thanks for bringing it to my attention! I think it would be just as much an issue on Linux if yarn wasn't installed.

> > > That being said, we could change CI to use the native cmd on Windows if needed, or just ignore all this since it seems to be trivial, your call :)
> > 
> > 
> > Good question. What does the CI use on Windows instead of the native cmd? If we switch to the native cmd, will the other one still work, or do we need to run on both?
> 
> CI is using powershell by default on Windows: [docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell)
> So, it depends on what you want to test. I personally still use the default cmd a lot hence why I hit the issue, but I guess most developers on Windows do run some kind of emulation. So, I'd say just leave it as is until it comes up again :)

Thanks! That's helpful. I'm still curious about my last question, but I'm okay leaving it as it is :).

.github/workflows/test.yml
XhmikosR: BTW something must be wrong with this change because CI doesn't seem to be firing. I'll try to figure out why, later, and ping you.

raineorshine: Okay, thanks! Should I merge or wait until you figure that out?

XhmikosR: Yup :) https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#excluding-branches-and-tags

I'll flip the check and see how it goes

XhmikosR: @raineorshine should be good to go now. Feel free to change the test to exclude the Unicode character :)

raineorshine: Not sure if that was "yup" to merge or "yup" to wait, so I guess I'll wait! Thanks

XhmikosR: It's ready now :)

PR #761, opened 2020-11-09
Regenerate package-lock.json

XhmikosR:
This had a weird entry...

```json
    "jsonlines": {
      "version": "0.1.1",
      "resolved": "https://registry.npm.taobao.org/jsonlines/download/jsonlines-0.1.1.tgz",
      "integrity": "sha1-T80kbcXQ44aRkHxEqwAveC0dlMw="
    },
```

Note that we can use [lockfile-lint](https://github.com/lirantal/lockfile-lint) to catch such issues.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,19 +14,19 @@
       }
     },
     "@babel/core": {
-      "version": "7.10.5",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.10.5.tgz",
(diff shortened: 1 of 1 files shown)


raineorshine:
@XhmikosR I did some more digging and found a good discussion here: https://github.com/npm/cli/issues/1431

The `taobao` registry was added in 78faa87e4e0d859b04146cead6f6143e4a1bf51c, likely due to the China-based developer's cli automatically switching to a closer registry.

XhmikosR:
Yup, I figured it out too, but still, I'd rather err on the safe side and not allow arbitrary sources; the npm ecosystem is so fragile as is 🙂

raineorshine:
Agreed. Glad to solve the mystery!

PR #762, opened 2020-11-09
Add lockfile-lint

XhmikosR:
I'm going to rebase after #761 

BTW we could probably split the actions later and move lint to a separate action which would run only once.


raineorshine:
Looks good!

PR #763, opened 2020-11-09
yarn tests: Fallback to local yarn install

raineorshine:
By default, npm-check-updates uses the system version of `npm` and `yarn`. However if a developer does not have `yarn` installed, the tests will fail. We still want system `yarn` to be used by default since that more closely matches the user's setup, but we should fallback to a `devDependency` when it is not available.

Fixes the issue raised in: https://github.com/raineorshine/npm-check-updates/pull/760#issuecomment-723771376

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2914,6 +2914,15 @@
             "p-locate": "^4.1.0"
           }
         },
+        "p-limit": {
+          "version": "2.3.0",
+          "resolved": "https://registry.npmjs.org/p-limit/-/p-limit-2.3.0.tgz",
+          "integrity": "sha512-//88mFWSJx8lxCzwdAABTJL2MyWB12+eIY7MDL2SqLmAkeKU9qxRvWuSyTjm3FUmpBEMuFfckAIqEaVGUDxb6w==",
+          "dev": true,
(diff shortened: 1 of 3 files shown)


XhmikosR:
Looks good!

I still get a failure but it's not from yarn:

```
 1) doctor
       upgrade dependencies when tests pass:
     AssertionError: expected 'npm install\nnpm run test\n' to include 'Tests pass'
      at Context.<anonymous> (test\test-doctor.js:70:19)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

XhmikosR (approved):

raineorshine:
> I still get a failure but it's not from yarn:
> 
> ```
>  1) doctor
>        upgrade dependencies when tests pass:
>      AssertionError: expected 'npm install\nnpm run test\n' to include 'Tests pass'
>       at Context.<anonymous> (test\test-doctor.js:70:19)
>       at processTicksAndRejections (internal/process/task_queues.js:93:5)
> ```

Huh, maybe it wasn't the unicode character that was causing it to fail. It appears to be existing prematurely after `npm run test`.

PR #768, opened 2020-11-14
Add repository links option to new --format command

### LINKED ISSUE #570: Feature: Show changelog URL for each new detected versions
> I could be nice for `npm-check-updates` in the interactive mode to display the changelog URL for each updated version, to ease user decision process 🤔

_Originally posted by @yvele in https://github.com/tjunnone/npm-check-updates/issues/555#issuecomment-508802010_

> Yes, I agree this would be a nice feature!

_Originally posted by @raineorshine in https://github.com/tjunnone/npm-check-updates/issues/555#issuecomment-508826303_

> Something like
> 
> ```
> $ ncu
> Checking package.json
> [====================] 3/3 100%
> 
>  express           4.12.x  →   4.13.x   https://github.com/expressjs/express/releases
>  multer            ^0.1.8  →   ^1.0.1   https://github.com/expressjs/multer/blob/master/CHANGELOG.md
>  react-bootstrap  ^0.22.6  →  ^0.24.0   https://github.com/react-bootstrap/react-bootstrap/blob/master/CHANGELOG.md
> 
> Run ncu -u to upgrade package.json
> ```
> 
> This is an example only to help you understand what I was suggesting. It's not a "real" example, maybe we want the full changelog in interactive mode.. This issue is not really the place to discuss that 😉

_Originally posted by @yvele in https://github.com/tjunnone/npm-check-updates/issues/555#issuecomment-508864031_

Edit: For information, `npm` is showing the changelog URL like this

image

But because `ncu` may show multiple times this kind of informations, we must find a way to fit the changelog URL in a one liner.

Also, is there a standard way to get a package changelog URL? Sometimes it's the GitHub release notes page, sometimes a dedicated changelog page, etc.

CreativeTechGuy:
### Main changes
* Added option to show repository links in the output with the new `--format repositoryLink` arg
  * This will infer the link from the data in the `package.json` for that package which is already installed on disk. If the link cannot be determined, the column will be left blank

### Minor changes
* Deprecates the `-o` and `--ownerChanged` CLI arguments in favor of the new `--format` arg
* Adds the new `--format` arg which can take a single string or a comma-delimited list of strings indicating which additional fields should be shown in the output. This enables future expansions for more custom/advanced data to be shown and enabled one at a time.
* Added `@deprecated` annotation to TypeScript types for deprecated fields
* Added new mechanism to warn users when using deprecated args with a custom message

### Developer experience changes
* Update build command to work on Windows
* Refactor `cli-args` data structure to make it easier to use this data elsewhere
* Refactored default args handling to use the `default` field from `cli-args`
* Adds `ncu` command to run this package against itself

Closes #570

### DIFF
--- a/lib/cli-options.js
+++ b/lib/cli-options.js
@@ -1,143 +1,184 @@
 // store CLI options separately from bin file so that they can be used to build type definitions
 const cliOptions = [
   {
+    optionName: 'concurrency',
     name: '--concurrency <n>',
     description: 'Max number of concurrent HTTP requests to registry.',
     parse: s => parseInt(s, 10),
     default: 8
   },
   {
+    optionName: 'configFilePath',
     name: '--configFilePath <path>',
     description: 'Directory of .ncurc config file (default: directory of `packageFile`).'
   },
   {
+    optionName: 'configFileName',
     name: '--configFileName <filename>',
     description: 'Config file name (default: .ncurc.{json,yml,js})'
   },
   {
+    optionName: 'cwd',
     name: '--cwd <path>',
     description: 'Working directory in which npm will be executed.'
   },
   {
+    optionName: 'dep',
     name: '--dep <dep>',
     description: 'Check one or more sections of dependencies only: prod, dev, peer, optional, bundle (comma-delimited).'
   },
   {
+    optionName: 'deprecated',
     name: '--deprecated',
     description: 'Include deprecated packages.'
   },
   {
+    optionName: 'doctor',
     name: '--doctor',
     description: 'Iteratively installs upgrades and runs tests to identify breaking upgrades. Run "ncu --doctor" for detailed help. Add "-u" to execute.',
   },
   {
+    optionName: 'enginesNode',
     name: '--enginesNode',
     description: 'Include only packages that satisfy engines.node as specified in the package file.'
   },
   {
+    optionName: 'errorLevel',
     name: '-e, --errorLevel <n>',
     description: 'Set the error level. 1: exits with error code 0 if no errors occur. 2: exits with error code 0 if no packages need updating (useful for continuous integration).',
     parse: s => parseInt(s, 10),
     default: 1
   },
   {
+    optionName: 'filter',
     name: '-f, --filter <matches>',
     description: 'Include only package names matching the given string, comma-or-space-delimited list, or /regex/.',
     type: 'string | string[] | RegExp'
   },
   {
+    optionName: 'global',
     name: '-g, --global',
     description: 'Check global packages instead of in the current project.'
   },
   {
+    optionName: 'greatest',
     name: '--greatest',
     description: 'DEPRECATED. Renamed to "--target greatest".'
   },
   {
+    optionName: 'interactive',
     name: '-i, --interactive',
     description: 'Enable interactive prompts for each dependency; implies -u unless one of the json options are set,'
   },
   {
     // program.json is set to true in programInit if any options that begin with 'json' are true
+    optionName: 'jsonAll',
     name: '-j, --jsonAll',
     description: 'Output new package file instead of human-readable message.'
   },
   {
+    optionName: 'jsonDeps',
     name: '--jsonDeps',
     description: 'Like `jsonAll` but only lists `dependencies`, `devDependencies`, `optionalDependencies`, etc of the new package data.'
   },
   {
+    optionName: 'jsonUpgraded',
     name: '--jsonUpgraded',
     description: 'Output upgraded dependencies in json.'
   },
   {
+    optionName: 'loglevel',
     name: '-l, --loglevel <n>',
     description: 'Amount to log: silent, error, minimal, warn, info, verbose, silly.',
     default: 'warn'
   },
   {
+    optionName: 'minimal',
     name: '-m, --minimal',
     description: 'Do not upgrade newer versions that are already satisfied by the version range according to semver.'
   },
   {
+    optionName: 'newest',
     name: '-n, --newest',
     description: 'DEPRECATED. Renamed to "--target newest".'
   },
   {
+    optionName: 'packageManager',
     name: '-p, --packageManager <name>',
     description: 'npm, yarn',
     default: 'npm'
   },
   {
+    optionName: 'ownerChanged',
     name: '-o, --ownerChanged',
-    description: 'Check if the package owner changed between current and upgraded version.',
+    description: 'DEPRECATED. Renamed to "--output ownerChanged".',
   },
   {
+    optionName: 'output',
+    name: '--output <value>',
+    description: 'Enable additional output data, string or comma-delimited list: ownerChanged, repositoryLink. ownerChanged: shows if the package owner changed between versions. repositoryLink: infers and displays links to source code repository.',
+    parse: value => typeof value === 'string' ? value.split(',') : value,
+    default: [],
+    type: 'string[]'
+  },
+  {
+    optionName: 'packageData',
     name: '--packageData <string>',
     description: 'Package file data (you can also use stdin).'
   },
   {
+    optionName: 'packageFile',
     name: '--packageFile <path>',
     description: 'Package file location (default: ./package.json).'
   },
   {
+    optionName: 'pre',
     name: '--pre <n>',
     description: 'Include -alpha, -beta, -rc. (default: 0; default with --newest and --greatest: 1).',
     type: 'boolean'
   },
   {
+    optionName: 'prefix',
     name: '--prefix <path>',
     description: 'Current working directory of npm.'
   },
   {
+    optionName: 'registry',
     name: '-r, --registry <url>',
     description: 'Third-party npm registry.'
   },
   {
+    optionName: 'removeRange',
     name: '--removeRange',
     description: 'Remove version ranges from the final package version.'
   },
   {
+    optionName: 'semverLevel',
     name: '--semverLevel <value>',
     description: 'DEPRECATED. Renamed to --target.'
   },
   {
+    optionName: 'silent',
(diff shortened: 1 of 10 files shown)


test/repo-url.test.js
```diff
@@ -0,0 +1,49 @@
...
+    })
+    it('return url constructed from git protocol', () => {
+      repoUrl.getRepoUrl('package-name', { repository: { url: 'git://github.com/user/repo.git' } }).should.equal('https://github.com/user/repo/tree/master')
+    })
+    it('return url constructed from http protocol', () => {
+      repoUrl.getRepoUrl('package-name', { repository: { url: 'http://github.com/user/repo.git' } }).should.equal('https://github.com/user/repo/tree/master')
+    })
+    it('return url with directory path', () => {
+      repoUrl.getRepoUrl('package-name', { repository: { url: 'http://github.com/user/repo.git', directory: 'packages/specific-package' } }).should.equal('https://github.com/user/repo/tree/master/packages/specific-package')
+    })
```
raineorshine, line 46: We should omit `/tree/master` whenever possible for a cleaner output.

lib/index.js
```diff
@@ -30,6 +31,15 @@ const packageFileNames = {
...
 const stdinWarningMessage = `Hmmmmm... this is taking a long time. Your console is telling me to wait for input \non stdin, but maybe that is not what you want.\nTry ${chalk.cyan('winpty ncu.cmd')}, or specify a package file explicitly with ${chalk.cyan('--packageFile package.json')}. \nSee https://github.com/raineorshine/npm-check-updates/issues/136#issuecomment-155721102`
 
+// In the format of [argName, deprecationMessage]
+const deprecatedArgs = [
+  ['greatest', '--greatest has been deprecated. Renamed to "--target greatest"'],
+  ['newest', '--newest has been deprecated. Renamed to "--target newest".'],
+  ['o', '-o has been deprecated. Renamed to "--output ownerChanged".'],
+  ['ownerChanged', '--ownerChanged has been deprecated. Renamed to "--output ownerChanged".'],
+  ['semverLevel', '--semverLevel has been deprecated. Renamed to --target']
+]
```
raineorshine, line 41: Thanks for adding these! Let's put them directly in `cli-options` so we don't have to maintain parallel data structures. You can add a `deprecated?: string` property to each option and then filter them later to print them out.

lib/index.js
```diff
@@ -166,7 +176,7 @@ async function analyzeProjectDependencies(options, pkgData, pkgFile) {
   const filteredUpgraded = options.minimal ? cint.filterObject(selectedNewDependencies, dep => !isSatisfied(dep)) : selectedNewDependencies
   const numUpgraded = Object.keys(filteredUpgraded).length
 
-  const ownersChangedDeps = options.ownerChanged
+  const ownersChangedDeps = options.ownerChanged || options.output.includes('ownerChanged')
```
raineorshine, line 179: Options normalization should go in `initOptions`.

lib/index.js
```diff
@@ -251,13 +261,11 @@ function initOptions(options) {
 
   return {
     ...options,
-    concurrency: options.concurrency || 8,
-    errorLevel: options.errorLevel != null ? options.errorLevel : 1,
```
raineorshine, line 255: Great, thanks for cleaning this up.

lib/repo-url.js
```diff
@@ -0,0 +1,66 @@
...
+const path = require('path')
+const { URL } = require('url')
+const hostedGitInfo = require('hosted-git-info')
+
+/**
+ * @param packageName A package name as listed in package.json's dependencies list
+ * @param packageJson Optional param to specify a object representation of a package.json file instead of loading from node_modules
+ * @returns A valid url to the root of the package's source or null if a url could not be determined
+ */
+function getRepoUrl(packageName, packageJson = undefined) {
```
raineorshine, line 13: `packageJson = undefined` can be simplified to `packageJson`

lib/repo-url.js
```diff
@@ -0,0 +1,66 @@
...
+  if (packageJson === undefined) {
```
raineorshine, line 14: I prefer `if (!packageJson)` since it also catches `null` and we don't have to worry about `false` or `0`.

lib/repo-url.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+      }
+      // The package.json file cannot be found
+      if (packageJson === undefined) {
+        return null
+      }
+    }
+    catch (e) {
+      return null
+    }
```
raineorshine, line 31: This logic should be factored out into a `getPackageRepo` function that returns `packageJson.repository` or `null`.

test/repo-url.test.js
```diff
@@ -0,0 +1,49 @@
+'use strict'
+
+const chai = require('chai')
+const repoUrl = require('../lib/repo-url')
+
+const should = chai.should()
+
+describe('repo-url', () => {
```
raineorshine, line 8: Thanks! Tests look good.

raineorshine (changes requested):
Thanks!

Please change `--output` to `--format`. Traditionally `--output` refers to files that are written to disk. `--format` fits our needs and has precedent in cli applications like `git`.

lib/cli-options.js
```diff
@@ -1,143 +1,184 @@
 // store CLI options separately from bin file so that they can be used to build type definitions
 const cliOptions = [
   {
+    optionName: 'concurrency',
     name: '--concurrency <n>',
```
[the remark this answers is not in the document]
CreativeTechGuy: Totally makes sense. I'll update to `long`, `short` and `arg`

test/repo-url.test.js
CreativeTechGuy: I was debating this. It seems like GitHub (and many others) are planning on renaming the default branch to something other than master [see here](https://github.com/github/renaming). While this does make the URLs less clean, I figured it'd be better to always include it than try to keep up with what the standard branch name is for each repository.

I was going for the least maintenance approach, but maybe I can run a test string through the `hosted-git-info` function to see what default branch name it spits out and use that to determine what string to remove when applicable.

Something like this maybe?

```
const commonBranchPath = hostedGitInfo.fromUrl('user/repo').browse('').match(/(\/tree\/[a-z]+)/)[0]
```

lib/index.js
CreativeTechGuy: Definitely!

lib/repo-url.js
CreativeTechGuy: I took this approach to make it explicit that it was optional. Having a default value (even a redundant one) will result in the IDE tooltip for this function to be:

```
getRepoUrl(packageName: any, packageJson?: any): any
```

rather than

```
getRepoUrl(packageName: any, packageJson: any): any
```

If you'd like it to be removed and rely on the JSDoc to explain that it's optional, I'll make the change.

lib/repo-url.js
CreativeTechGuy: Ha there's my TypeScript experience coming in. I try to not coerce values of different types and be explicit. But happy to change it if that's the standard here.

test/repo-url.test.js
raineorshine: Unless I am mistaken, we can follow the same pattern as the `repository` value itself: if it specifies a branch, use it; if it does not specify a branch, don't add anything. 

e.g. the repository for `npm-check-updates` is specified as: 

```json
  "repository": {
    "type": "git",
    "url": "git+https://github.com/raineorshine/npm-check-updates.git"
  },
```

This should generate `https://github.com/raineorshine/npm-check-updates`. There's no need to compare a branch at all. `hosted-git-info` is creating it. You may need to detect this manually if they don't have a way to detect a branch-less url.

lib/repo-url.js
raineorshine: Interesting! Well that's preferable until I get around to converting to Typescript :)

lib/repo-url.js
raineorshine: If it were in Typescript, the type checker would prevent `null` from being passed. But since we're in vanilla JS, we should handle `null` to guard against a runtime error. 

One of these days I'll get around to converting to Typescript...

test/repo-url.test.js
CreativeTechGuy: Yeah the library will always create the longest-form URL. So I'll detect and remove it.

raineorshine:
Thank you! It was a pleasure working with you!

P.S. I shortened the option to `--format repo`

raineorshine (approved):

raineorshine:
Published in `v10.1.0`

CreativeTechGuy:
Thanks for letting me contribute. As a long-time user I'm glad I was able to help make this great tool even better! And thanks for catching those final refactors!

cutesquirrel:
Hi, this feature is really great.
However, I have a strange issue : only the global packages are found (due to the `require.resolve.paths` that gives a list of many paths, but never the local project `node_modules` path)
Here is a debug output of the `require.resolve.paths` :
As you see, it doesn't look for the local project node_modules.
I gonna do some other tests, but if you have an idea, I'm interested !
thanks

Screenshot 2021-02-08 at 10 42 28

cutesquirrel:
Digging a bit deeper into the code, I realize that ncu was installed globally, so the resolve is done globally.

I tried to uninstall the global ncu and install it locally.
The only way to execute it is then : `node node_modules/npm-check-updates/bin/cli --format repo`
and it works ! I have much more repo URLs : 

Screenshot 2021-02-08 at 10 59 08

I tried to append the local module path, it seems to work well : `nodeModulePaths.unshift(path.join(process.cwd(), 'node_modules'))` even if the NCU is installed globally.

PR #772, opened 2020-12-10
Print correct install message when using yarn

PieterBoeren:
When using yarn, the message after running "ncu -u" should be "Run 'yarn install' ..." instead of "Run 'npm install' ...", this PR adds the correct message.


raineorshine:
Thank you!

raineorshine:
Published in `v10.2.4`

PR #785, opened 2021-01-12
Allow glob pattern as arg for packageFile cli option. closes #522

### LINKED ISSUE #522: Implement --deep flag to update deps in monorepo
Similar to  #154, but would it be idea to add `--deep` or similar flag to check updates in monorepo to update multiple packages found in repository?

use case would be doing something like bellow, but more elegantly:

```
const glob = require('glob')
const childProcess = require('child_process')

const check = (pkgJsonPath) => {
  try {
    return childProcess.execSync(`./node_modules/.bin/ncu --error-level 2 --packageFileDir --packageFile ${pkgJsonPath}`).toString()
  } catch (error) {
    console.log('exec error: ' + error.message)
    process.exit(error.status)
  }
}
glob('./packages/**/package.json', {}, (er, files) => {
  files.forEach((file) => {
    if (file.includes('node_modules')) {
      return
    }
    console.log(`command to update: ./node_modules/.bin/ncu --packageFileDir --packageFile ${file} -u -i`)
    console.log(check(file))
  })
})
```

* node version: *
* npm version: *
* npm-check-updates version: *

---------------------------

mkungla:
Allow `glob` as arg for `--packageFile` and implement --deep flag as alias of
`--packageFile '{,*[!node_modules]/**/}package.json'`

closes #522


package.json
```diff
@@ -63,6 +63,7 @@
     "find-up": "5.0.0",
     "fp-and-or": "^0.1.3",
     "get-stdin": "^8.0.0",
+    "glob": "^7.1.6",
```
kachkaev, line 66: WDYT of using [globby](https://www.npmjs.com/package/globby) instead?

raineorshine (changes requested):
Thanks for the contribution!

It looks good. A few additional considerations:

- **Testing** - We should have at least once test for this new behavior. I would suggest something like this in `bin.test.js`:

    ```js
    it('--deep', async () => {
      const output = await spawn('node', [path.join(__dirname, '../bin/cli.js'), '--jsonUpgraded', '--deep'], { cwd: path.join(__dirname, 'ncu/deep') })
      ...
    })
    ```

- **Performance** - It's expensive to run `analyzeProjectDependencies` multiple times as-is since it will duplicate `npm view` for dependencies that show up in multiple package files. It would be more efficient to determine all of the versions up front and pass that to `analyzeProjectDependencies` to bypass the typical lookups. This of course adds additional work. In lieu of this, I think a warning message and invitation to contribute would be adequate. I think users would expect dependencies to get de-duped, so a warning here is appropriate. What do you think?

- **Breaking** - I believe this is a breaking change since `packageFile` is now interpreted as a glob pattern, and it's possible that a value that someone used before will not be interpreted the same way. No big deal, just wanted to make that explicit here.

mkungla:
> * **Testing** - We should have at least once test for this new behavior. I would suggest something like this in `bin.test.js`:
>   ```js
>   it('--deep', async () => {
>     const output = await spawn('node', [path.join(__dirname, '../bin/cli.js'), '--jsonUpgraded', '--deep'], { cwd: path.join(__dirname, 'ncu/deep') })
>     ...
>   })
>   ```

Added some tests...

> * **Performance** - It's expensive to run `analyzeProjectDependencies` multiple times as-is since it will duplicate `npm view` for dependencies that show up in multiple package files. It would be more efficient to determine all of the versions up front and pass that to `analyzeProjectDependencies` to bypass the typical lookups. This of course adds additional work. In lieu of this, I think a warning message and invitation to contribute would be adequate. I think users would expect dependencies to get de-duped, so a warning here is appropriate. What do you think?

Not sure what would be best option here... should `versionmanager` have some caching   

> * **Breaking** - I believe this is a breaking change since `packageFile` is now interpreted as a glob pattern, and it's possible that a value that someone used before will not be interpreted the same way. No big deal, just wanted to make that explicit here.

It's should not have breaking changes or what somebody could have passed to `--packageFile` which would not fall to

https://github.com/raineorshine/npm-check-updates/blob/a53e9224d2c9d603f13e71dd55da386edb12b8bc/lib/index.js#L451-L457

lib/index.js
```diff
@@ -250,6 +254,10 @@ function initOptions(options) {
     ...options.ownerChanged ? ['ownerChanged'] : []
   ]
 
+  if (options.deep) {
+    options.packageFile = deepPattern
+  }
```
raineorshine, line 259: We want to avoid unnecessary mutations; let's set `packageFile` in the merged return object below.

lib/index.js
```diff
@@ -415,17 +423,40 @@ async function run(options = {}) {
...
+      const deepJsonOut = {}
+      analysis = await pkgs.reduce(async (previousPromise, packageFile) => {
+        const packages = await previousPromise
+        if (packageFile.length > 0 && packageFile !== defaultPackageFilename) {
+          options.packageFile = packageFile // set packageFile only if needed
+        }
+        const [pkgData, pkgFile] = await findPackage(options)
+        return {
+          ...packages,
+          [pkgFile]: await analyzeProjectDependencies(options, pkgData, pkgFile, deepJsonOut)
```
raineorshine, line 444: Since the `reduce` returns the aggregate output, we should be able to pass that directly to `printJson` rather than using `deepJsonOut`, unless I'm missing something. Either way, I consider passing a mutable object to a subcall to modify to be a bad practice. Let's try to find an immutable solution.

raineorshine (changes requested):
Tests look really great, thank you for taking the time! Let's rename `monorepo` to `deep` to create a closer association between the tests and what feature is being tested.

raineorshine:
> Not sure what would be best option here... should `versionmanager` have some caching

Yes, we would probably want to cache the results from [queryVersions](https://github.com/raineorshine/npm-check-updates/blob/master/lib/versionmanager.js#L444). Given the additional work, I'm okay with a warning message instead. If we do take it on, we should merge this first and do that work in a separate PR, since this PR has a good scope right now.

> It's should not have breaking changes or what somebody could have passed to `--packageFile` which would not fall to

`npm-check-updates` is particularly conservative since it is used in many CI setups. I can't definitively conclude that no user was utilizing a `--packageFile` that would now be interpreted as a glob and relying on unspecified behavior. Making this a breaking change allows developers to evaluate for themselves if there is any risk in upgrading.

lib/index.js
mkungla: Only reason currently is to display well formatted `json` output.

```
{
  "package.json": {
    ...
  },
  "pkgs/pkg1/package.json": {
    ...
  }
  ...
}

```


https://github.com/raineorshine/npm-check-updates/blob/59cae824de738767e3ffcba02151be5f9825bb25/lib/logging.js#L38-L42

mkungla:
updates also make use of `--prefix` flag e.g. to check everything in `~/projects` directory regardless of initial `pwd`

`ncu --deep --prefix '~/projects`

lib/index.js
raineorshine: Yes, that is the intended functionality. I'm not speaking of the functionality, but rather the implementation. I consider mutations a bad practice. Unfortunately passing a value by reference and mutating it is unacceptable for the coding standards of this project. I understand and respect that approach for other situations, but just not here. You'll have to find a way to achieve the same functionality but follow a more functional approach that respects the value of immutability.

raineorshine (changes requested):
> updates also make use of `--prefix` flag e.g. to check everything in `~/projects` directory regardless of initial `pwd`
> 
> `ncu --deep --prefix '~/projects`

Great idea! However the `--prefix` option is for overriding the `npm prefix`. `--cwd` is used to change where the current working directory.

mkungla:
> > updates also make use of `--prefix` flag e.g. to check everything in `~/projects` directory regardless of initial `pwd`
> > `ncu --deep --prefix '~/projects`
> 
> Great idea! However the `--prefix` option is for overriding the `npm prefix`. `--cwd` is used to change where the current working directory.

Ok that was a BUG I created before introducing the feature :disappointed: . It's fixed now

`ncu --deep --cwd '~/projects` works as expected now

lib/index.js
mkungla: my bad missing that solution to follow that practice was super simple

mkungla:
Since `--deep` is alias of default `--packageFile glob` then should it ignore other directories  like `test|tests`  etc.

raineorshine:
Thanks for being responsive to feedback and helping to expand the functionality of `npm-check-updates`! This will benefit many other users.

raineorshine:
> > Not sure what would be best option here... should `versionmanager` have some caching
> 
> Yes, we would probably want to cache the results from [queryVersions](https://github.com/raineorshine/npm-check-updates/blob/master/lib/versionmanager.js?rgh-link-date=2021-01-15T16%3A26%3A03Z#L444). Given the additional work, I'm okay with a warning message instead. If we do take it on, we should merge this first and do that work in a separate PR, since this PR has a good scope right now.

@mkungla This was WAY easier than I thought it would be! https://github.com/raineorshine/npm-check-updates/commit/0f81e6db47503688e0cdc56b4bf15aab36fd426d

mkungla:
> @mkungla This was WAY easier than I thought it would be! [0f81e6d](https://github.com/raineorshine/npm-check-updates/commit/0f81e6db47503688e0cdc56b4bf15aab36fd426d)

@raineorshine  that's nice and to follow up

1. Docs in `README` need also update.
2. `suppressErrors: false,` should be removed so that errors like `Error: EACCES: permission denied, scandir` would not make it hang forever.

https://github.com/raineorshine/npm-check-updates/blob/c1e2f3d39824e33e73ece1910debdc84422ba1cb/lib/index.js#L425-L428

raineorshine:
> 1. Docs in `README` need also update.

Yes, I forgot to run `npm run build` which regenerates the options in the README.

> 2\. `suppressErrors: false,` should be removed so that errors like `Error: EACCES: permission denied, scandir` would not make it hang forever.

Great, done.

PR #794, opened 2021-01-21
Minor README.md tweaks

XhmikosR:
* switch to shields.io for the badges; it's more consistent and should be faster
* add missing code block language
* remove `$` from one-line command where there's no output

FYI https://dlaa.me/markdownlint/ (there is also an npm package)


raineorshine:
> FYI [dlaa.me/markdownlint](https://dlaa.me/markdownlint/) (there is also an npm package)

Cool! That'll be helpful.

PR #795, opened 2021-01-21
package.json: fix `lint:markdown` script

XhmikosR:
Single quotes don't work on all platforms

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "build": "node bin/build.js",
     "lint": "npm run lint:src && npm run lint:lockfile && npm run lint:markdown",
     "lint:lockfile": "lockfile-lint --allowed-hosts npm --allowed-schemes https: --empty-hostname false --type npm --path package-lock.json",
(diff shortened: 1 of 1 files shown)


raineorshine:
Thanks :)

XhmikosR:
Might be worth looking at https://github.com/DavidAnson/markdownlint-cli2 later too :)

raineorshine:
> Might be worth looking at [DavidAnson/markdownlint-cli2](https://github.com/DavidAnson/markdownlint-cli2) later too :)

I looked briefly. It had fewer stars and less recent updates so I want with the main cli.

XhmikosR:
TBH this one is the newer cli made by the creator of markdowlint. He maintains both packages, but cli2 is a lot faster. Your call, I personally have switched to cli2 and it works like a charm, but it has a few fundamental differences.

raineorshine:
Thanks, I'll keep that in mind.

PR #796, opened 2021-01-21
package.json: be more explicit with files section

XhmikosR:
This is nitpicking, still build.js doesn't need to be in the npm package.

Another way or along with this would be to move build.js to a script folder or something, but I'll leave that to you to decide @raineorshine


raineorshine:
Good idea. How about 69fa856?

XhmikosR:
That's better, but I'd still keep the wildcards to be more explicit. I've seen random .DS_Store files included in various packages for example. But it's your call in the end :)

raineorshine:
I'm fine with the simpler syntax for now. Thanks!

Plus, have you ever had too many .DS_Store files??? 😂

XhmikosR:
> 
> 
> I'm fine with the simpler syntax for now. Thanks!
> 
> Plus, have you ever had too many .DS_Store files??? 😂

Well, it's 2 issues IMHO

1. size (I know, negligible, but still :) )
2. any file can be accidentally included with the folder alone

The downside is that if new filetypes are needed, you will need to add them manually.

raineorshine:
I see your point. I lean towards over-including rather than under-including. Over-including may result in unnecessary size, but at least it won't break the release.

PR #801, opened 2021-01-22
README.md: fix syntax highlighting

XhmikosR:


raineorshine:
After doing a closer comparison, I prefer the `shell` output. `console` uses `blue` for its main color instead of black, which I find to be uglier/harder to read for large blocks of text. I can live with a few miscolorings, and some of them like numbers and punctuation end up looking better anyway.

XhmikosR:
But those are not shell scripts they are shell outputs :) Anyway, your call.

raineorshine:
Technically, sure, but the whole point is to have nice syntax highlighting. I'll label it `fortran` if I have to :).

PR #805, opened 2021-01-26
fix `**/package.json` cause comment loose effect

joebnb:
fix `**/package.json` cause comment loose effect

image

```
 /**
 * Run recursively in current working directory. Alias of (--packageFile '**/package.json').
 */
    deep?: boolean;
```
should add \ to avoid or use inline comment


raineorshine:
Thanks!

PR #808, opened 2021-01-28
update docs with npx

MkMan:
Update README.md to include npx usage instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,16 @@ You may also want to consider [npm-check](https://github.com/dylang/npm-check). 
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


raineorshine:
Thank you for the suggestion. I would prefer to keep the README limited to ncu-specific instructions. `npx` is a tool that users can learn about elsewhere.

MkMan:
> Thank you for the suggestion. I would prefer to keep the README limited to ncu-specific instructions. `npx` is a tool that users can learn about elsewhere.

Fair enough 🙂  Awesome tool by the way, thank you 👏

PR #821, opened 2021-02-08
Repo Url: Use local node_modules

cutesquirrel:
…. But it's precisely the most important path. Added the local path manually in the list

See https://github.com/raineorshine/npm-check-updates/pull/768#issuecomment-775033067


raineorshine:
Good catch, and thanks for digging into the code to find the fix!

I’ll publish a patch as soon as I get my computer back, in a few days.

PR #852, opened 2021-03-12
feat: try SSH keys before HTTPS for github packages without protocol

### LINKED ISSUE #850: Private repo with a semver tag hangs
- [x] I am using `node >= 10.17`

---------------------------

## Steps to Reproduce

1. Add a private GitHub repo with a semver tag, e.g.
    ```
    "dependencies": {
        "@newredo/my-private-repo": "NewRedo/my-private-repo#semver:^1.0.0"
    }
    ```
2. Run `ncu`.

## Current Behavior

Hangs forever.

## Expected Behavior

Either tells you it can't access the private repo or checks it.

## Comments

Removing the semver tag works fine -- it only hangs when the semver tag is present.

midgleyc:
Fixes #850.

If no protocol is specified for a github URL, try the SSH url followed by the HTTPS url. If a protocol is specified, only try that one: assume that people have put in the url correctly.

We try SSH first because HTTPS might prompt for credentials for private repos, while SSH fails silently.

The graphical glitch where the password prompt pops up and disappears if the progress bar updates persists: I'm okay with that because I can't see how you could manage to install a package without either an SSH key or having signed in to GitHub, in which case you're hopefully expecting it.

### DIFF
--- a/lib/package-managers/gitTags.js
+++ b/lib/package-managers/gitTags.js
@@ -15,10 +15,11 @@ const getSortedVersions = async (name, declaration, options) => {
   const protocolKnown = protocol != null
   if (protocolKnown) {
     tagsPromise = tagsPromise.then(() => remoteGitTags(`${protocol ? protocol.replace('git+', '') : 'https:'}//${auth ? auth + '@' : ''}${host}/${path}`))
-  } else {
+  }
+  else {
     // try ssh first, then https on failure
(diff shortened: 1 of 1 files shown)


raineorshine:
Thanks so much!

raineorshine:
Published to `v11.3.0`

PR #854, opened 2021-03-28
README: Add async to example

827652549:
if you not , you can see 

```
SyntaxError: await is only valid in async function
```

by the way ，i add two options(packageFile and upgrade) so that easier to understand it


raineorshine:
Thank you!

PR #869, opened 2021-04-01
Respect peer dependencies

### LINKED ISSUE #376: Feature request: detect when dependencies should not be updated because other packages are peer dependent with the old version
Whenever I use ncu to check and update my dependencies, I feel like most of the effort is spent manually picking through a bunch of package.json files to see if the peer dependencies match up properly across all the updates.

Inevitably some library updated to a new major version and all the plugins haven't caught up yet, so I have to manually update everything and manually figure out what the latest updates are that ARE compatible with each other.

Obviously, this is not a tiny body of work, but it seems like you have all the information to be able to KNOW when a dependency should not be updated or know the most up to date version you can safely update to.

korelstar:
The purpose of this PR is to fix #376. 

@raineorshine Please have a look if this goes in the correct direction. Open questions are, e.g.:
- should the check be optional, what is default?
- which warning should be raised if detection does not work? (the tests are currently failing due to this warning)

I've exported the new `getPeerDependencies` for the purpose that other tools can use it too, e.g. https://github.com/th0r/npm-upgrade

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -142,14 +142,14 @@ async function analyzeProjectDependencies(options, pkgData, pkgFile) {
     options.enginesNode = _.get(pkg, 'engines.node')
   }
 
+  if (options.checkPeer) {
+    options.peerDependencies = getPeerDependencies(current, options)
+  }
+
   print(options, '\nOptions:', 'verbose')
   print(options, sortOptions(options), 'verbose')
 
-  const peerDependencies = getPeerDependencies(current, options)
-  print(options, '\nPeerDependencies:', 'verbose')
-  print(options, peerDependencies, 'verbose')
-
-  const [upgraded, latest] = await vm.upgradePackageDefinitions(current, { ...options, peerDependencies })
+  const [upgraded, latest] = await vm.upgradePackageDefinitions(current, options)
 
   print(options, '\nFetched:', 'verbose')
   print(options, latest, 'verbose')
@@ -247,8 +247,7 @@ function getPeerDependencies(current, options) {
       if (acc[pkgName] === undefined) {
         acc[pkgName] = []
       }
-      // eslint-disable-next-line fp/no-mutating-methods
-      acc[pkgName].push(version)
+      acc[pkgName][acc[pkgName].length] = version
     })
     return acc
   }, {})
--- a/lib/index.js
+++ b/lib/index.js
@@ -145,7 +145,11 @@ async function analyzeProjectDependencies(options, pkgData, pkgFile) {
   print(options, '\nOptions:', 'verbose')
   print(options, sortOptions(options), 'verbose')
 
-  const [upgraded, latest] = await vm.upgradePackageDefinitions(current, options)
+  const peerDependencies = getPeerDependencies(current, options)
+  print(options, '\nPeerDependencies:', 'verbose')
+  print(options, peerDependencies, 'verbose')
+
+  const [upgraded, latest] = await vm.upgradePackageDefinitions(current, { ...options, peerDependencies })
 
   print(options, '\nFetched:', 'verbose')
   print(options, latest, 'verbose')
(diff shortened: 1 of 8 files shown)


lib/index.js
```diff
@@ -224,6 +228,32 @@ async function analyzeProjectDependencies(options, pkgData, pkgFile) {
...
+    catch (e) {
+      print(options, 'Could not read peer dependencies for package ' + pkgName + '. Is this package installed?', 'warn')
+      return {}
+    }
+  }).reduce((acc, peers) => {
+    Object.entries(peers).forEach(([pkgName, version]) => {
+      if (acc[pkgName] === undefined) {
+        acc[pkgName] = []
+      }
+      // eslint-disable-next-line fp/no-mutating-methods
```
raineorshine, line 250: Note: PR's that disable the linter will not be accepted in order to maintain the project's code style and quality requirements. You'll have to find a way to do this in a more functional style with no mutations.

raineorshine (changes requested):
> The purpose of this PR is to fix #376.
> 
> @raineorshine Please have a look if this goes in the correct direction.

Thanks for your contribution!

> * should the check be optional, what is default?

It should be opt-in via the `--peer` option. Default `ncu` behavior is intentionally kept very minimal so it is easy to reason about and there are no surprises.

> * which warning should be raised if detection does not work? (the tests are currently failing due to this warning)

Can you explain more what you mean by "does not work"? Do you mean an upgrade breaks a peer dependency (expected), or do you mean the detection algorithm doesn't work in some cases (unexpected)? Are there other cases besides when packages have not yet been installed that you are thinking of?

Some other thoughts:

- Should `ncu` ignore breaking peer dependencies completely when this option is provided, or should it suggest the latest version that is compatible? Not sure what would be expected/most useful.
- Unit test required

Thanks!

lib/index.js
korelstar: To be honest, I just copied that from your own code: https://github.com/raineorshine/npm-check-updates/search?q=no-mutating-methods

But not problem, I've fixed that now :-)

raineorshine: Fair enough! There is some legacy code from an older version that had different linter settings.

korelstar:
> It should be opt-in via the `--peer` option. Default `ncu` behavior is intentionally kept very minimal so it is easy to reason about and there are no surprises.

The `--peer` option can't be used, since there is already an attribute `peer` in the `options` variable which is `true` by default. It looks like it is filled by the `--dep` command line option. Therefore, I have chosen the option `--checkPeer` for now.

> > * which warning should be raised if detection does not work? (the tests are currently failing due to this warning)
> 
> Can you explain more what you mean by "does not work"? Do you mean an upgrade breaks a peer dependency (expected), or do you mean the detection algorithm doesn't work in some cases (unexpected)? Are there other cases besides when packages have not yet been installed that you are thinking of?

In fact, I think it's only relevant when a package is not installed locally. An upgrade should never break a peer dependency, since the purpose of this PR is to check this (see next comment).

> * Should `ncu` ignore breaking peer dependencies completely when this option is provided, or should it suggest the latest version that is compatible? Not sure what would be expected/most useful.

My goal is that `ncu` suggests the latest version that is compatible to the peer dependencies. I think this is what a user would expect and this is what I've implemented. I don't think a user would expect that peer dependency are ignored completely.

> * Unit test required

I've added a test, now. The test uses a package [`ncu-test-peer`](https://www.npmjs.com/package/ncu-test-peer) (which has a peer dependency to `ncu-test-return-version`) for this purpose. We can transfer the ownership to you, if you want.

raineorshine:
> > It should be opt-in via the `--peer` option. Default `ncu` behavior is intentionally kept very minimal so it is easy to reason about and there are no surprises.
> 
> The `--peer` option can't be used, since there is already an attribute `peer` in the `options` variable which is `true` by default. It looks like it is filled by the `--dep` command line option. Therefore, I have chosen the option `--checkPeer` for now.

Perfect! I will go ahead and remove those options since they are legacy and internal only.

> > * Should `ncu` ignore breaking peer dependencies completely when this option is provided, or should it suggest the latest version that is compatible? Not sure what would be expected/most useful.
> 
> My goal is that `ncu` suggests the latest version that is compatible to the peer dependencies. I think this is what a user would expect and this is what I've implemented. I don't think a user would expect that peer dependency are ignored completely.

That makes sense.

> > * Unit test required
> 
> I've added a test, now. The test uses a package [`ncu-test-peer`](https://www.npmjs.com/package/ncu-test-peer) (which has a peer dependency to `ncu-test-return-version`) for this purpose. We can transfer the ownership to you, if you want.

Yes! That would be great.

lib/index.js
```diff
@@ -224,6 +228,31 @@ async function analyzeProjectDependencies(options, pkgData, pkgFile) {
...
 
+/** Get peer dependencies from installed packages */
+function getPeerDependencies(current, options) {
+  const basePath = options.cwd || './'
+  return Object.keys(current).map(pkgName => {
+    const path = basePath + 'node_modules/' + pkgName + '/package.json'
+    try {
+      const pkgData = fs.readFileSync(path)
+      const pkg = jph.parse(pkgData)
+      return vm.getCurrentDependencies(pkg, { dep: 'peer', filter: options.filter, reject: options.reject })
```
raineorshine, line 239: We should probably pass the all of the users' options and just override `dep: 'peer'`:

```js
return vm.getCurrentDependencies(pkg, { ...options, dep: 'peer' })
```

raineorshine (changes requested):

lib/index.js
korelstar: Yes, that sounds good. I've fixed it.

korelstar:
> > > * Unit test required
> > 
> > 
> > I've added a test, now. The test uses a package [`ncu-test-peer`](https://www.npmjs.com/package/ncu-test-peer) (which has a peer dependency to `ncu-test-return-version`) for this purpose. We can transfer the ownership to you, if you want.
> 
> Yes! That would be great.

I've added you as maintainer. We can remove me now or after this PR is merged. Whatever you prefer.

raineorshine (approved):

raineorshine:
Published in `v11.4.0`

Issue #826, opened 2021-02-17
`ncu --deep` ignore .ncurc.yml settings


art-ws:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [x] I am using the latest version of `npm-check-updates`
- [x] I am using `node >= 10.17`

---------------------------

## Abstract

Follow up on [closed issue](https://github.com/raineorshine/npm-check-updates/issues/825)

## Steps to Reproduce
* Create [or clone](https://github.com/art-ws/temp) workspace with nested packages 
* Define `.ncurc.yml` for some projects
* Run `ncu -u --deep` for root

## Current Behavior
Tool update all package versions and ignore `.ncurc.yml` settings.

## Expected Behavior
Would be nice if tool can recognize existing `.ncurc.yml` settings file and follow it rules.

raineorshine:
Yes, good catch!

art-ws:
Fixed in v11.4.1

PR #871, opened 2021-04-07
fix: Fix ignore .ncurc.yml settings for --deep option

art-ws:
#826


raineorshine:
Thanks! Could we get a test?

art-ws:
> Thanks! Could we get a test?

Not sure we have a time for it:) Just forked and using a fixed version for a while. 
Although fix is too simple, but test implementation for such core feature better to implement by core team as you understand all details and bound cases much more better than random contributor.

raineorshine:
Cool. Shouldn’t be too hard. Thanks!

raineorshine:
Published in `v11.4.1`

PR #872, opened 2021-04-19
feat: detect yarn according to yarn.lock

Tchoupinax:
Hello,

I saw that PR were merged to detect yarn but It does not work for me. I read the code and I do not understand how yarn could be detected.

I add the following code set yarn as package manager if `yarn.lock` is detected without the presence of `package-lock.json`.
I use the simplest code I can :)

Thank you for your project, it is very useful!
Have a nice day.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -5,7 +5,6 @@
 const program = require('commander')
 const _ = require('lodash')
 const updateNotifier = require('update-notifier')
-const fs = require('fs')
 const ncu = require('../lib/')
 const pkg = require('../package.json')
 const cliOptions = require('../lib/cli-options')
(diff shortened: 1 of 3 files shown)


raineorshine:
Hi! Thanks for your contribution! Currently there is no automatic detection of yarn, as you discovered. It must be specified manually via `-p yarn`. I don't see any reason not to autodetect it when no `package-lock.json` is detected.

Tchoupinax:
Hi!

I made this behaviour because I consider `npm` like the default package manager and the presence of `package-lock.json` could override the presence of `yarn.lock`.
Thank you for merging the feature!

Have a nice day,

PR #874, opened 2021-04-25
update peer dependencies on upgrade and retry

korelstar:
This is a follow-up of #869 and improves the situation, where updated packages have different peer dependencies. Here is an example:

The package `ncu-test-peer-update` is published with two versions:
- `1.0.0` has peer dependency `"ncu-test-return-version": "1.0.x"`
- `1.1.0` has peer dependency `"ncu-test-return-version": "1.1.x"`

Our test app has the following dependencies:
```
    "ncu-test-peer-update": "1.0.0",
    "ncu-test-return-version": "1.0.0"
```

### Before (i.e. with `ncu` version 11.5.1)
A call of `ncu --peer` will propose the following updates:
```
 ncu-test-peer-update     1.0.0  →  1.1.0     
 ncu-test-return-version  1.0.0  →  1.0.1     
```
There is not update suggestion for `ncu-test-return-version` to `1.1.0`, since the installed version of `ncu-test-peer-update` requires version `1.0.x`. This means, I would have to run `npm install` afterwards and then again `ncu --peer` to get all updates.

### After (i.e. with this PR)
Now, `ncu` will get the new peer dependencies from the registry and check again for updates. I.e., the resulting update suggestion will contain all compatible updates.
```
 ncu-test-peer-update     1.0.0  →  1.1.0     
 ncu-test-return-version  1.0.0  →  1.1.0     
```

### Without `--peer` option
Just as comparison: without using the `--peer` option, `ncu` will suggest incompatible updates in both cases (with and without this PR):
```
 ncu-test-peer-update     1.0.0  →  1.1.0     
 ncu-test-return-version  1.0.0  →  2.0.0     
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -149,7 +149,10 @@ async function analyzeProjectDependencies(options, pkgData, pkgFile) {
   print(options, '\nOptions:', 'verbose')
   print(options, sortOptions(options), 'verbose')
 
-  const [upgraded, latest] = await vm.upgradePackageDefinitions(current, options)
+  const [upgraded, latest, upgradedPeerDependencies] = await vm.upgradePackageDefinitions(current, options)
+
+  print(options, '\nupgradedPeerDependencies:', 'verbose')
+  print(options, upgradedPeerDependencies, 'verbose')
 
   print(options, '\nFetched:', 'verbose')
   print(options, latest, 'verbose')
@@ -233,22 +236,18 @@ function getPeerDependencies(current, options) {
   const basePath = options.cwd || './'
   return Object.keys(current).map(pkgName => {
     const path = basePath + 'node_modules/' + pkgName + '/package.json'
+    let peers = {}
     try {
       const pkgData = fs.readFileSync(path)
       const pkg = jph.parse(pkgData)
-      return vm.getCurrentDependencies(pkg, { ...options, dep: 'peer' })
+      peers = vm.getCurrentDependencies(pkg, { ...options, dep: 'peer' })
     }
     catch (e) {
       print(options, 'Could not read peer dependencies for package ' + pkgName + '. Is this package installed?', 'warn')
-      return {}
     }
-  }).reduce((acc, peers) => {
-    Object.entries(peers).forEach(([pkgName, version]) => {
-      if (acc[pkgName] === undefined) {
-        acc[pkgName] = []
-      }
-      acc[pkgName][acc[pkgName].length] = version
-    })
+    return [pkgName, peers]
+  }).reduce((acc, [pkgName, peers]) => {
+    acc[pkgName] = peers
     return acc
   }, {})
 }
--- a/lib/index.js
+++ b/lib/index.js
@@ -234,7 +234,7 @@ async function analyzeProjectDependencies(options, pkgData, pkgFile) {
 /** Get peer dependencies from installed packages */
 function getPeerDependencies(current, options) {
   const basePath = options.cwd || './'
-  return Object.keys(current).map(pkgName => {
+  return Object.keys(current).reduce((accum, pkgName) => {
     const path = basePath + 'node_modules/' + pkgName + '/package.json'
     let peers = {}
     try {
@@ -245,10 +245,7 @@ function getPeerDependencies(current, options) {
     catch (e) {
       print(options, 'Could not read peer dependencies for package ' + pkgName + '. Is this package installed?', 'warn')
     }
-    return [pkgName, peers]
-  }).reduce((acc, [pkgName, peers]) => {
-    acc[pkgName] = peers
-    return acc
+    return { ...accum, [pkgName]: peers }
   }, {})
 }
 
--- a/lib/versionmanager.js
+++ b/lib/versionmanager.js
@@ -214,7 +214,7 @@ async function getOwnerPerDependency(fromVersion, toVersion, options) {
 }
 
 /**
- * Returns an 2-tuple of upgradedDependencies and their latest versions.
+ * Returns an 3-tuple of upgradedDependencies, their latest versions and the resulting peer dependencies.
  *
  * @param currentDependencies
  * @param options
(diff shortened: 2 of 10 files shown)


lib/versionmanager.js
```diff
@@ -441,6 +458,37 @@ async function queryVersions(packageMap, options = {}) {
...
+  const numItems = Object.keys(packageMap).length
+  let bar
+  if (!options.json && options.loglevel !== 'silent' && options.loglevel !== 'verbose' && numItems > 0) {
+    bar = new ProgressBar('[:bar] :current/:total :percent', { total: numItems, width: 20 })
+    bar.render()
+  }
+
+  const packageManager = getPackageManager(options.packageManager)
+  let result = {}
+  if (packageManager.getPeerDependencies) {
```
raineorshine, line 478: Rather than use an if statement here, it would be better to do the opposite check and short circuit the logic:

```js
if (!packageManager.getPeerDependencies) return {}
```

This eliminates the mutation and results in flatter code. You can put the conditional above the `ProgressBar` instantiation.

lib/versionmanager.js
```diff
@@ -441,6 +458,37 @@ async function queryVersions(packageMap, options = {}) {
...
+    result = await Object.entries(packageMap).reduce(async (prom, [pkg, version]) => {
+      const dep = await packageManager.getPeerDependencies(pkg, version)
+      if (bar) {
+        bar.tick()
+      }
+      const acc = await prom
+      acc[pkg] = dep
+      return acc
```
raineorshine, line 486: Here you can avoid the mutation by using:

```js
return {
  ...acc,
  pkg: dep,
}
```

lib/versionmanager.js
```diff
@@ -441,6 +458,37 @@ async function queryVersions(packageMap, options = {}) {
...
+  let bar
+  if (!options.json && options.loglevel !== 'silent' && options.loglevel !== 'verbose' && numItems > 0) {
+    bar = new ProgressBar('[:bar] :current/:total :percent', { total: numItems, width: 20 })
+    bar.render()
+  }
+
+  const packageManager = getPackageManager(options.packageManager)
+  let result = {}
+  if (packageManager.getPeerDependencies) {
+    result = await Object.entries(packageMap).reduce(async (prom, [pkg, version]) => {
```
raineorshine, line 479: Let's call the accumulator `promise` or `accumPromise` so it's more obvious to other developers.

lib/versionmanager.js
```diff
@@ -231,7 +231,24 @@ async function upgradePackageDefinitions(currentDependencies, options) {
     return !options.jsonUpgraded || !options.minimal || !isSatisfied(latestVersions[dep], currentDependencies[dep])
   })
 
-  return [filteredUpgradedDependencies, latestVersions]
+  let result = [filteredUpgradedDependencies, latestVersions, options.peerDependencies]
+  if (options.peer && !_.isEmpty(filteredUpgradedDependencies)) {
+    const upgradedPeerDependencies = await getPeerDependenciesFromRegistry(filteredUpgradedDependencies, options)
+    const peerDependencies = { ...options.peerDependencies, ...upgradedPeerDependencies }
+    result[2] = peerDependencies
```
raineorshine, line 238: If you could find a way to do this without mutations that would be great. They introduce ordering constraints into the code and I consider them a bad practice. For example, in this code it is less clear than it could be all of the places that `result` is set. With a more functional approach, the code structure itself would help to show the code paths that lead to different `result` values.

lib/index.js
```diff
@@ -233,22 +236,18 @@ function getPeerDependencies(current, options) {
...
-      if (acc[pkgName] === undefined) {
-        acc[pkgName] = []
-      }
-      acc[pkgName][acc[pkgName].length] = version
-    })
+    return [pkgName, peers]
+  }).reduce((acc, [pkgName, peers]) => {
+    acc[pkgName] = peers
     return acc
   }, {})
```
raineorshine, line 252: Here you can use [_.fromPairs](https://lodash.com/docs/4.17.15#fromPairs), or, even better, you can convert the above `map` into a `reduce` and avoid the additional loop.

raineorshine (changes requested):
Thanks for the additional contribution and excellent explanation! Makes perfect sense. I'd love to add your description to a markdown file that documents for users how the `--peer` feature works.

I just have a few requests for some simplifications and conforming to a more functional programming paradigm. Thanks and let me know if you have any questions!

lib/versionmanager.js
korelstar: Thanks for your feedback! I've fixed all issues except this one. I will have to think about, how I can improve this part. But I agree, that the readability can be improved :smiley:

korelstar: I added a fix for this (the `result[2] = peerDependencies` wasn't even necessary :roll_eyes: ). Is it better, now?

raineorshine: Yes, thank you!

raineorshine (approved):

raineorshine:
Thank you!

korelstar:
@raineorshine : Are you interested in a functionality for `ncu` to show a list of ignored incompatible updates? This could look like this:


#### Example 1:
```
New versions of active modules available:

  ncu-test-peer-update      1.0.0   →   1.1.0 
  ncu-test-return-version   1.0.0   →   1.1.0 

  ncu-test-return-version   1.0.0   →   2.0.0   reason: ncu-test-peer-update requires 1.1.x 
```

#### Example 2:
```
New versions of active modules available:

  @nextcloud/dialogs                ^3.1.1   →     ^3.1.2 
  @nextcloud/router                 ^1.2.0   →     ^2.0.0 
  @nextcloud/vue                    ^3.8.0   →     ^3.9.0 
  @nextcloud/webpack-vue-config     ^4.0.1   →     ^4.0.3 
  easymde                          ^2.14.0   →    ^2.15.0 
  markdown-it                      ^12.0.4   →    ^12.0.6 
  @babel/core                     ^7.13.14   →   ^7.13.16 
  @babel/preset-env               ^7.13.12   →   ^7.13.15 
  core-js                          ^3.10.0   →    ^3.11.0 
  eslint                           ^7.23.0   →    ^7.25.0 
  eslint-plugin-vue                 ^7.8.0   →     ^7.9.0 
  eslint-webpack-plugin             ^2.5.3   →     ^2.5.4 
  sass                             ^1.32.8   →   ^1.32.11 
  stylelint                       ^13.12.0   →   ^13.13.0 
  webpack                          ^5.28.0   →    ^5.35.1 

  css-loader               ^4.3.0   →    ^5.2.4   reason: @nextcloud/webpack-vue-config requires ^4.3.0                                    
  eslint-plugin-promise    ^4.3.1   →    ^5.1.0   reason: @nextcloud/eslint-config requires ^4.2.1, eslint-config-standard requires ^4.2.1 
  sass-loader             ^10.1.1   →   ^11.0.1   reason: @nextcloud/webpack-vue-config requires ^10.0.5                                   

Ignored updates:

                  From          To   Ignored versions   Reason        
  vue-fragment   1.5.1   →   1.5.2        1.5.2         buggy version 
```

Background: I've made a PR for [`npm-upgrade`](https://github.com/th0r/npm-upgrade/) (a tool which uses `npm-check-updates` for providing interactive updates) which generates such a list: https://github.com/th0r/npm-upgrade/pull/59

I could move that code from `npm-upgrade` to `npm-check-updates`, if you want.

raineorshine:
Looks nice!

- Is this intended to only occur with `--peer`?
- What's the difference between "Ignored incompatible updates (peer dependencies)" and "Ignored updates"?

korelstar:
>     * Is this intended to only occur with `--peer`?

Yes.

>     * What's the difference between "Ignored incompatible updates (peer dependencies)" and "Ignored updates"?

Oh, sorry. The section "Ignored updates" is a special thing of `npm-upgrade`. It would not be there in `ncu`.

raineorshine:
Yes, I like it!

PR #876, opened 2021-04-27
print summary of incompatible/ignored updates

korelstar:
see https://github.com/raineorshine/npm-check-updates/pull/874#issuecomment-826938074

Example:

```
 vue-fragment    1.5.1  →    1.5.2     
 webpack       ^5.35.1  →  ^5.36.0     

 css-loader              ^4.3.0  →   ^5.2.4  reason: @nextcloud/webpack-vue-config requires ^4.3.0                                    
 eslint-plugin-promise   ^4.3.1  →   ^5.1.0  reason: @nextcloud/eslint-config requires ^4.2.1, eslint-config-standard requires ^4.2.1 
 sass-loader            ^10.1.1  →  ^11.0.1  reason: @nextcloud/webpack-vue-config requires ^10.0.5                                   

```

### DIFF
--- a/lib/logging.js
+++ b/lib/logging.js
@@ -140,4 +140,17 @@ function printUpgrades(options, { current, upgraded, numUpgraded, total, ownersC
   }
 }
 
-module.exports = { print, printJson, printUpgrades, toDependencyTable }
+function printIgnoredUpdates(options, ignoredUpdates) {
+  print(options, `\nIgnored incompatible updates (peer dependencies):\n`)
+  const table = createDependencyTable()
+  const rows = Object.entries(ignoredUpdates).map(([pkgName, { from, to, reason }]) => {
+    const strReason = 'reason: ' + Object.entries(reason)
+      .map(([pkgReason, requirement]) => pkgReason + ' requires ' + requirement)
+      .join(', ')
+    return [pkgName, from, '→', colorizeDiff(from, to), strReason]
+  })
+  rows.forEach(row => table.push(row)) // eslint-disable-line fp/no-mutating-methods
+  print(options, table.toString())
+}
+
+module.exports = { print, printJson, printUpgrades, toDependencyTable, printIgnoredUpdates }
--- a/lib/versionmanager.js
+++ b/lib/versionmanager.js
@@ -328,6 +328,26 @@ function getCurrentDependencies(pkgData = {}, options = {}) {
   return allDependencies
 }
 
+async function getIgnoredUpgrades(current, upgraded, upgradedPeerDependencies, options = {}) {
+  const [upgradedLatestVersions, latestVersions] = await upgradePackageDefinitions(
+    current,
+    { ...options, peer: false, peerDependencies: undefined, loglevel: 'silent' }
+  )
+
+  return Object.entries(upgradedLatestVersions)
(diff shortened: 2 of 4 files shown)


lib/logging.js
```diff
@@ -140,4 +140,17 @@ function printUpgrades(options, { current, upgraded, numUpgraded, total, ownersC
...
+function printIgnoredUpdates(options, ignoredUpdates) {
+  print(options, `\nIgnored incompatible updates (peer dependencies):\n`)
+  const table = createDependencyTable()
+  const rows = Object.entries(ignoredUpdates).map(([pkgName, { from, to, reason }]) => {
+    const strReason = 'reason: ' + Object.entries(reason)
+      .map(([pkgReason, requirement]) => pkgReason + ' requires ' + requirement)
+      .join(', ')
+    return [pkgName, from, '→', colorizeDiff(from, to), strReason]
+  })
+  rows.forEach(row => table.push(row)) // eslint-disable-line fp/no-mutating-methods
```
korelstar, line 152: Sorry for this, but I don't know how to do it in another way.

The same approach is already in the existing code:
https://github.com/raineorshine/npm-check-updates/blob/v11.5.3/lib/logging.js#L93

raineorshine: Thanks for calling it out providing justification! No problem.

Actually, it annoyed me so I contributed a PR to their repo 😆
https://github.com/Automattic/cli-table/pull/139

lib/versionmanager.js
```diff
@@ -328,6 +328,26 @@ function getCurrentDependencies(pkgData = {}, options = {}) {
   return allDependencies
 }
 
+async function getIgnoredUpgrades(current, upgraded, upgradedPeerDependencies, options = {}) {
+  const [upgradedLatestVersions, latestVersions] = await upgradePackageDefinitions(
+    current,
+    { ...options, peer: false, peerDependencies: undefined, loglevel: 'silent' }
+  )
```
raineorshine, line 335: Doesn't `analyzeProjectDependencies` already call `upgradePackageDefinitions`? This would result in doubling all requests to `queryVersions` and `getPeerDependencies`. That seems like a poor design unless I'm missing something.

korelstar: Yes and no.

This is necessary, since the target functions (e.g. `latest`) directly filter out all other versions than the one desired. There is no chance to access the latest compatible version (with option `peerDependencies`) and the latest incompatible version (without option `peerDependecies`) at the same time. `analyzeProjectDependencies` calls `upgradePackageDefinitions` with peer dependencies and this code calls it without.

However, the requests aren't doubled since you cache all responses already:
https://github.com/raineorshine/npm-check-updates/blob/v11.5.4/lib/package-managers/npm.js#L151-L154

Therefore, the second call is executed from memory and hence runs very fast.

Btw., `getPeerDependencies` isn't called twice.

raineorshine: Excellent points! Thanks so much for walking me through that. Some day I would like to re-architect npm-check-updates to have a cleaner pipeline with distinct fetch, label, and filter phases. In the mean time, this will work fine.

I forgot about the caching! That was added for `--deep` but it's really handy here.

raineorshine:
Published in `v11.5.5`

PR #877, opened 2021-04-29
Add node v16 to test matrix and remove deprecated fs.rmDir

raineorshine:


XhmikosR:
Oops, sorry just opened #885. I don't think the solution is to switch to `fs.rm` since it's not available on Node.js < v14.14.0.

In #885 I just disabled the `throw-deprecation` mocha option.

raineorshine:
Yeah, I discovered the same issue :).

PR #885, opened 2021-05-01
Update CI config

XhmikosR:
* Add Node.js 16
* Switch to `actions/cache@v2`

**EDIT**: I added a patch so that we get a nice colored output on CI too. `strip-ansi` is already an indirect dependency so it doesn't cost us anything. If you agree with this, I can rebase to squash that patch with the first one.


raineorshine:
> **EDIT**: I added a patch so that we get a nice colored output on CI too. `strip-ansi` is already an indirect dependency so it doesn't cost us anything. If you agree with this, I can rebase to squash that patch with the first one.

Yup sounds good to me!

PR #888, opened 2021-05-02
Typescript!

raineorshine:
- [x] tsconfig
- [x] source maps
- [x] typescript-eslint
- [x] mocha integration
- [x] Split out getPeerDependencies
- [x] Split out version manager functions
- [x] Split out index functions
- [x] Convert npm to typescript
- [x] Convert yarn to typescript
- [x] Remove module.exports syntax
- [x] Remove vm from public API
- [ ] Branded version types (optional)
  - Version, VersionRaw, VersionSpec
  - Also Github urls, npm urls, etc
- [ ] Convert tests to typescript (optional)
  - A lot of tests!!! Can be done incrementally


raineorshine:
@XhmikosR Do you have experience with Typescript? Would love to have someone else check over my setup. I haven't gotten very far with the actual conversion but it's compiling and tests are passing at least.

package.json
```diff
@@ -30,21 +30,23 @@
       "raine"
     ]
   },
-  "main": "./lib",
+  "main": "./build/src",
   "scripts": {
     "build": "node scripts/build.js",
     "lint": "cross-env FORCE_COLOR=1 npm-run-all --parallel --aggregate-output lint:*",
```
XhmikosR, line 36: Unrelated but forgot to mention this the other day: why do we have `FORCE_COLOR=1` here? If it's for CI, this is now fixed there and locally it shouldn't be needed. It's one less devDependency :)

XhmikosR:
I'm not familiar with TypeScript either just a couple of quick thoughts:

1. make sure linting works properly for TS files and exclude the build folder
2. make sure tsconfig.json has the proper target for the supported Node.js versions; no point in targeting Node.js 6 or 8 for example
3. I'd switch to using xo for linting, or something else that does cover TS linting better

.github/workflows/test.yml
```diff
@@ -48,6 +48,9 @@ jobs:
       - name: Install npm dependencies
         run: npm ci
 
+      - name: Compile typescript
+        run: tsc
```
XhmikosR, line 52: Might make more sense to have this in an npm script and call it before publish or something, to make sure it's always the latest.

raineorshine:
> 1. make sure linting works properly for TS files and exclude the build folder

Yup, I still need to do that. I have a `@typescript-eslint` setup from another project I can use. I like the additional control of `eslint`.

> 2\. make sure tsconfig.json has the proper target for the supported Node.js versions; no point in targeting Node.js 6 or 8 for example

Good call. I just switched to "es2018" which targets node v10 (and soon can bump to "es2019" when we remove v10 support).

package.json
raineorshine: It's not just for CI. `npm-run-all` appears to have some [known limitations with coloring](https://github.com/mysticatea/npm-run-all/blob/bf91f94ce597aa61da37d2e4208ce8c48bc86673/docs/npm-run-all.md#known-limitations). If I remove `FORCE_COLOR=1` and run it locally it loses color output.

PR #889, opened 2021-05-03
doctor: Restore package.json to emulate --no-save in yarn

raineorshine:


blowery:
thanks! i noticed this earlier this morning and was writing up an issue for it when I noticed this PR. great timing! 😆

Issue #895, opened 2021-05-25
ncu --deep override and ignore parent .ncurc.yml settings (reject)


art-ws:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [x] I am using the latest version of `npm-check-updates`
- [x] I am using `node >= 10.17`

---------------------------
## Abstract

If we have a workspace (like [rushjs monorepo](https://rushjs.io/)) with hierarchy of nested projects sometimes we need to reject some specific packages for all projects in workspace. Currently it's possible to achieve via file system config with package-specific `.ncurc.yml` which works great after the fix #826 , but it require a lot of identical file updates in existing config files and explicit creating of additional configs. If we define a parent `.ncurc.yml` it works as expected only for packages without package-specific `.ncurc.yml` config.

## Steps to Reproduce

1. Define `.ncurc.yml` for some projects in root folder and reject some package
2. Define `.ncurc.yml`for some nested projects without rejected package
3. Not define `.ncurc.yml`for some nested package
3. Run `ncu -u --deep` 

## Current Behavior

`ncu` correctly recognize parent `.ncurc.yml` and apply it correctly only for nested packages, which have no own settings. If some project contains own local `.ncurc.yml`, then parent settings are discarded. 
As result different packages in monorepo even on same level of hierarchy may contains different versions and whole workspace state became inconsistent.

## Expected Behavior

Would be nice to inherit settings from parent configs (at least for 'reject' option), so tool behevior will be more consistent.

art-ws:
We tested locally a fix with literally few lines of code and it works as expected.
If you approve we can prepare the PR.

File `lib/index.js`

```javascript
...
async function getAnalysis() {
...
       const packages = await previousPromise
        // copy object to prevent share .ncurc options between different packageFile, to prevent unpredictable behavior
        const rcResult = getNcurc({ packageFile })

       // BEGIN: Possible #895 fix
        if (options.reject && rcResult && rcResult.config && rcResult.config.reject) {
          rcResult.config.reject = [...new Set([...options.reject, ...rcResult.config.reject])]
        }
       // END

        const pkgOptions = {
          ...options,
          ...rcResult && rcResult.config,
          packageFile,
        }
        const [pkgData, pkgFile] = await findPackage(pkgOptions)
...
```

Currently problem is in next lines which fully overrides the nested object property values, but, probably kind of more smart object merge implementation will be better here. WDT?

[...]

PR #897, opened 2021-05-26
feat: --mergeConfig option for --deep to inherit config settings

art-ws:
#895

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -508,9 +508,14 @@ async function run(options = {}) {
         const packages = await previousPromise
         // copy object to prevent share .ncurc options between different packageFile, to prevent unpredictable behavior
         const rcResult = getNcurc({ packageFile })
+        const rcConfig = rcResult && rcResult.config ? rcResult.config : {}
+        if (options.recursive && options.reject) {
+          // inherit parent config options
+          rcConfig.reject = [...new Set([...options.reject, ...rcConfig.reject || []])]
+        }
         const pkgOptions = {
           ...options,
-          ...rcResult && rcResult.config,
+          ...rcConfig,
           packageFile,
         }
         const [pkgData, pkgFile] = await findPackage(pkgOptions)
--- a/lib/index.js
+++ b/lib/index.js
@@ -46,6 +46,25 @@ function programError(options, message) {
   }
 }
 
+function mergeArrays(arr1, arr2) {
+  return [...new Set([...arr1 || [], ...arr2 || []])]
+}
+
+/**
+ * Merge options by a standard way, but if some properties both are arrays, then merge them also.
+ */
+function mergeOptions(options1, options2) {
+  options1 = options1 || {}
+  options2 = options2 || {}
+  const result = { ...options1, ...options2 }
+  Object.keys(result).forEach(key => {
+    if (Array.isArray(options1[key]) && Array.isArray(options2[key])) {
+      result[key] = mergeArrays(options1[key], options2[key])
+    }
+  })
+  return result
+}
+
 /**
  * Gets the name of the package file based on --packageFile or --packageManager.
  */
@@ -508,10 +527,10 @@ async function run(options = {}) {
         const packages = await previousPromise
         // copy object to prevent share .ncurc options between different packageFile, to prevent unpredictable behavior
         const rcResult = getNcurc({ packageFile })
-        const rcConfig = rcResult && rcResult.config ? rcResult.config : {}
-        if (options.recursive && options.reject) {
+        let rcConfig = rcResult && rcResult.config ? rcResult.config : {}
+        if (options.recursive) {
           // inherit parent config options
-          rcConfig.reject = [...new Set([...options.reject, ...rcConfig.reject || []])]
+          rcConfig = mergeOptions(options, rcConfig)
         }
         const pkgOptions = {
           ...options,
@@ -592,4 +611,4 @@ function getNcurc({ configFileName, configFilePath, packageFile } = {}) {
   return result ? { ...result, args } : null
 }
 
-module.exports = { run, getNcurc, getPeerDependencies, ...vm }
+module.exports = { run, getNcurc, getPeerDependencies, mergeOptions, ...vm }
--- a/lib/index.js
+++ b/lib/index.js
@@ -510,9 +510,9 @@ async function run(options = {}) {
         // copy object to prevent share .ncurc options between different packageFile, to prevent unpredictable behavior
         const rcResult = getNcurc({ packageFile })
         let rcConfig = rcResult && rcResult.config ? rcResult.config : {}
-        if (options.mergeConfig) {
-          // Merge config options. (Note: arguments order is matter)
-          rcConfig = mergeOptions(options, rcConfig, { keys: Object.keys(rcConfig) })
+        if (options.mergeConfig && Object.keys(rcConfig).length) {
+          // Merge config options.
+          rcConfig = mergeOptions(options, rcConfig)
         }
         const pkgOptions = {
           ...options,
--- a/lib/index.js
+++ b/lib/index.js
@@ -510,8 +510,8 @@ async function run(options = {}) {
         // copy object to prevent share .ncurc options between different packageFile, to prevent unpredictable behavior
         const rcResult = getNcurc({ packageFile })
         let rcConfig = rcResult && rcResult.config ? rcResult.config : {}
-        if (options.recursive) {
-          // Merge config options
+        if (options.mergeConfig) {
+          // Merge config options. (Note: arguments order is matter)
           rcConfig = mergeOptions(options, rcConfig, { keys: Object.keys(rcConfig) })
         }
         const pkgOptions = {
--- a/lib/index.js
+++ b/lib/index.js
@@ -22,6 +22,7 @@ const packageManagers = require('./package-managers')
 const { print, printJson, printUpgrades, printIgnoredUpdates } = require('./logging')
 const { deepPatternPrefix, doctorHelpText } = require('./constants')
 const cliOptions = require('./cli-options')
+const mergeOptions = require('./merge-options')
 
 // maps package managers to package file names
 const packageFileNames = {
@@ -44,25 +45,6 @@ function programError(options, message) {
   else {
     throw new Error(message)
   }
-}
-
-function mergeArrays(arr1, arr2) {
-  return [...new Set([...arr1 || [], ...arr2 || []])]
-}
-
-/**
- * Merge options by a standard way, but if some properties both are arrays, then merge them also.
- */
-function mergeOptions(options1, options2) {
-  options1 = options1 || {}
-  options2 = options2 || {}
-  const result = { ...options1, ...options2 }
-  Object.keys(result).forEach(key => {
-    if (Array.isArray(options1[key]) && Array.isArray(options2[key])) {
-      result[key] = mergeArrays(options1[key], options2[key])
-    }
-  })
-  return result
 }
 
 /**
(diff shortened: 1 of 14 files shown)


lib/index.js
```diff
@@ -508,9 +508,14 @@ async function run(options = {}) {
         const packages = await previousPromise
         // copy object to prevent share .ncurc options between different packageFile, to prevent unpredictable behavior
         const rcResult = getNcurc({ packageFile })
+        const rcConfig = rcResult && rcResult.config ? rcResult.config : {}
+        if (options.recursive && options.reject) {
+          // inherit parent config options
+          rcConfig.reject = [...new Set([...options.reject, ...rcConfig.reject || []])]
```
raineorshine, line 514: We should apply config inheritance to all options, not just reject.

raineorshine (changes requested):

lib/index.js
art-ws: Yep. Can you please advise where to find the complete list of nested structured properties?

raineorshine: All of the options are listed in [https://github.com/raineorshine/npm-check-updates/blob/main/lib/cli-options.js](cli-options.js). I hope you’re not planning on applying them individually?

art-ws: For merge we can apply such simple generic function with test coverage:
```javascript
function mergeArrays(arr1, arr2){
  return [...new Set([...(arr1 || []), ...(arr2 || [])])]
}

function mergeOptions(options1, options2){
  options1 = options1 || {}
  options2 = options2 || {}  
  const result = {...options1, ...options2 }
  for (const key in result) {
    if (result.hasOwnProperty(key)) {
      if (Array.isArray(options1[key]) && Array.isArray(options2[key])){
        result[key] = mergeArrays(options1[key], options2[key]) 
      }
    }
  }
  return result;
}
```

i.e. it is a normal merge, but if both properties are arrays, then we merge them also.
WDT?

raineorshine: Sure, I'm okay with that.

lib/index.js
```diff
@@ -587,4 +611,4 @@ function getNcurc({ configFileName, configFilePath, packageFile } = {}) {
   return result ? { ...result, args } : null
 }
 
-module.exports = { run, getNcurc, getPeerDependencies, ...vm }
+module.exports = { run, getNcurc, getPeerDependencies, mergeOptions, ...vm }
```
raineorshine, line 614: We don't want to add `mergeOptions` to the public api just for testing. Instead, define it in its own module in `/src/lib` and import it directly in the test.

lib/index.js
```diff
@@ -46,6 +46,25 @@ function programError(options, message) {
   }
 }
 
+function mergeArrays(arr1, arr2) {
+  return [...new Set([...arr1 || [], ...arr2 || []])]
+}
+
+/**
+ * Merge options by a standard way, but if some properties both are arrays, then merge them also.
```
raineorshine, line 54: You should specify in the JSDOC that is a shallow merge (as opposed to a recursive or deep merge).

test/deep.test.js
```diff
@@ -144,6 +144,67 @@ describe('--deep with nested ncurc files', function () {
...
+    deepJsonOut.should.have.property('pkg/sub3/package.json')
+    deepJsonOut['pkg/sub3/package.json'].should.not.have.property('cute-animals')
+    deepJsonOut['pkg/sub3/package.json'].should.have.property('fp-and-or')
+    deepJsonOut['pkg/sub3/package.json'].should.have.property('ncu-test-v2')
+  })
+
+  it('use ncurc of nested packages with --recursive option', async () => {
+
+    const configFilePath = path.join(__dirname, 'deep-ncurc', '.ncurc.js')
+    const deepJsonOut = await spawn('node', [bin, '--jsonUpgraded', '--deep', '--recursive', '--configFilePath', configFilePath], { cwd }).then(JSON.parse)
```
raineorshine, line 158: If you specify `--configFilePath` on the command line, it will use that config file on every nested package anyway (without `--recursive`), as command line options apply to every package. If you omit this option, it will use the config file in `cwd`.

lib/index.js
```diff
@@ -508,9 +527,14 @@ async function run(options = {}) {
         const packages = await previousPromise
         // copy object to prevent share .ncurc options between different packageFile, to prevent unpredictable behavior
         const rcResult = getNcurc({ packageFile })
+        let rcConfig = rcResult && rcResult.config ? rcResult.config : {}
+        if (options.recursive) {
+          // inherit parent config options
+          rcConfig = mergeOptions(options, rcConfig)
```
raineorshine, line 533: This inherits not just the parent's config options, but the parent's command-line options. We currently maintain this distinction:

- command line options apply to every package
- config options apply locally to the package they are in

Merging command-line options instead of just merging config options might have unintended consequences.

raineorshine (changes requested):

raineorshine:
I suspect that there may be another problem with this logic and that it is not exposed by the test case. I could be wrong but I wanted to bring it up in case you're not aware of it.

Subpackages are retrieved as a flat array of paths via `globby.sync`:

https://github.com/raineorshine/npm-check-updates/blob/486535fd605ceb8cf402029dde784bb404e36c0c/lib/index.js#L492-L497

This means that there is no notion of "parent" and "child" in the current `--deep` implementation. Practically this means that with a structure such as the following:

- a
  - b
    - c
  - d
    - e

There is no way in the current implementation to tell that "e" should inherit from "a" and "d" but not "b" or "c". I'm guessing that it will fail on such a test case.

You may have to replace globby with manual traversal in order to guarantee the order and know who the parent is.

lib/index.js
art-ws: OK

test/deep.test.js
art-ws: Without `--configFilePath` tests not works as implementation have a side-effects depended on `process.env.NCU_TESTS`
and internal state is diffrent due:
```javascript
// load .ncurc
// Do not load when global option is set
// Do not load when tests are running (an be overridden if configFilePath is set explicitly)
const rcResult = !program.global && (!process.env.NCU_TESTS || configFilePath)
  ? ncu.getNcurc({ configFileName, configFilePath, packageFile })
  : null
```
Behevior is very implicit and not obvious. Same code rinning from cli and from test has a different behevior.
As a lot of tests depends on `process.env.NCU_TESTS` it was a most simple way to arcjive the predictable behevior.

lib/index.js
art-ws: OK

lib/index.js
art-ws: Yes, you right. Will change implementation to merge onlu subset of keys from `rcConfig`

art-ws:
> I suspect that there may be another problem with this logic and that it is not exposed by the test case. I could be wrong but I wanted to bring it up in case you're not aware of it.

"Perfect is the enemy of the good" :) Actually, such enhancement  already successfully cover the practical cases and generally bring a new step in overall tool evolution. But, for sure, make sense to think how to cover more complex cases.

test/deep.test.js
raineorshine: > Behevior is very implicit and not obvious. Same code rinning from cli and from test has a different behevior.
> As a lot of tests depends on `process.env.NCU_TESTS` it was a most simple way to achieve the predictable behevior without core code modifications.

I agree, not the best solution. The tests should probably always run in a clean directory.

I like your solution of checking `recursive` to ensure the config file gets loaded.

lib/index.js
raineorshine: That's not quite what I had in mind. I will try to explain in a review comment!

lib/index.js
```diff
@@ -508,9 +509,14 @@ async function run(options = {}) {
         const packages = await previousPromise
         // copy object to prevent share .ncurc options between different packageFile, to prevent unpredictable behavior
         const rcResult = getNcurc({ packageFile })
+        let rcConfig = rcResult && rcResult.config ? rcResult.config : {}
+        if (options.recursive) {
+          // Merge config options
+          rcConfig = mergeOptions(options, rcConfig, { keys: Object.keys(rcConfig) })
```
raineorshine, line 515: `options` consist of command-line options and config options which are combined (prematurely imo) in `cli.js`. Thus we don't have direct access to the root package config at this point. My previous comment was saying that ideally we only merge config options (not command line options) but that doesn't seem possible without some additional refactoring. Luckily it doesn't seem to be a problem with the main use case we are solving here.

So, I may not have been following the `key` selection logic correctly, but I don't think we need it now and I don't think it offers much advantage if we are merging `options` anyway. Let me know if I'm missing something!

raineorshine (changes requested):

raineorshine:
Considering the functionality that we've settled on, I think a more appropriate name would be `--merge-config`. `--deep` is already recursive and this option specifically merges nested configs with the root config rather than overriding. What do you think?

Make sure to run `npm run build` to regenerate the README and type definitions.

lib/index.js
art-ws: Technically `{ keys: Object.keys(rcConf) }` guarantee(we have test coverage on it) that we merge only the properties from  `rcResult.config` and all other properties from `options` will not be changed. For our particular use case it may be even more explicit `{ keys: ['reject'] }`, but current implementation is more generic and looks pretty safe. Logic behind `mergeOptions` is simple - it's the same `Object.assign` but if some props are arrays, then we merge them also. Additional  `keys` parameter allow us to be sure that may be changed only properties explicitly specified in `rcConfig` to prevent  unintended consequences.

Any case all this feature isolated from all other code with `--deep` and `--recursive` options, so this enhancement fix the problem and not brings the breaking changes.

raineorshine: > Technically `{ keys: Object.keys(rcConf) }` guarantee(we have test coverage on it) that we merge only the properties from `rcResult.config` and all other properties from `options` will not be changed.

I'm not sure I understand. If an option does not exist in `rcResult`, then how could it overwrite `options`?

i.e. `_.pick(options2, Object.keys(options2))` is a noop

art-ws: Correct. If no any properties in `rcResult` then nothing to merge for this config, i.e. no need to override something in `options`. The idea is to override only properties, which explicitly defined in `rcResult`. If nothing defined, then nothing to do.

raineorshine: Thanks for your response. I was unclear in my last message, so let me try again. Please correct me if I am missing something.

What I'm seeing is that on L515 you call:


```js
mergeOptions(options, rcConfig, { keys: Object.keys(rcConfig) })
```

And in merge-options [L15](https://github.com/raineorshine/npm-check-updates/pull/897/files#diff-c37a68c010b04a2f15f188f945826efc4a558a607f7d1871fda1d557ce4d25ffR15) we have: 

```js
options2 = opts.keys ? _.pick(options2 || {}, opts.keys) : options2 || {} 
```

So when the arguments get substituted through, we get:

```js
options2 = opts.keys ? _.pick(rcConfig, Object.keys(rcConfig) : rcConfig
```

Which is always `options2 = options2`, regardless of the inputs. Thus the `keys` logic is entirely superfluous which I don't believe was your intention. If you could remove `opts.keys` I think everything else looks good to go.

raineorshine (changes requested):

art-ws:
Oh..yes:) Good catch. Such things happens when 2 diffrent ideas meets together during the series of refactoring. 
The purpose of `keys` argument was to have a generic approach to restrict keys to merge, but in case of actual usage of such function we always have a trivial case.

Let's simplify implementation.

raineorshine:
Thanks! I’ll take a look asap

raineorshine (approved):

raineorshine:
Published in `v11.6.0`

PR #917, opened 2021-07-07
Fix github dependency upgrades with partial semver

markablov:
I have added a test that shows a problem.

Error occurs because `revertMissingPatch` and `revertMissingMinorAndPatch` relies that `latest` is in format `major.minor.patch`, while github refs could be in other format.

Maybe it worth to make this functions more generic, but because of their specialisation i opted to add minimal changes rather than overengineer it.


raineorshine:
Good catch! Thank you.

> Maybe it worth to make this functions more generic, but because of their specialisation i opted to add minimal changes rather than overengineer it.

I like the pointed solution at the moment. I'll consider how to best generalize it later.

PR #919, opened 2021-07-12
replace mem with fast-memoize

43081j:
mem pulls in a few levels deep of dependencies for something pretty simple.

here i've replaced it with fast-memoize which has no dependencies and seems much faster/simpler.

let me know if you want any changes

FYI the default serializer is JSON.stringify internally so the cache will be keyed by the stringified args afaict (equivalent of before, almost)


raineorshine (approved):

raineorshine:
Looks good!

PR #979, opened 2021-11-04
Fix getPeerDependencies to use cwd

cfbender:
Hi there!

I noticed after upgrading to v12 that the `--peer` option seemed to be broken from https://github.com/raineorshine/npm-check-updates/commit/d2584170738690b99e1d32e27d98702fc7a3c906

This is a simple change that fixes that by not using the path above.

Example:
image


raineorshine (approved):

cfbender:
Wasn't sure if you wanted merge commits in to update or not so I went ahead and rebased it on main

raineorshine:
Good catch! Thanks!

PR #981, opened 2021-11-06
Update npm scripts to use npm-run-all

XhmikosR:
Notes:

* let me know if we can use `--parallel` anywhere here (or feel free to push)
* `chmod` is nix only and fails on Windows. I'm not sure if it's needed, just a FYI that it fails on Windows

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "main": "build/src/index.js",
   "scripts": {
-    "build": "npm-run-all build:*",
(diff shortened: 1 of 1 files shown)


raineorshine (approved):

raineorshine:
`npm-run-all` requires `cross-env FORCE_COLOR=1` to be set, as seen in the `lint` script.

raineorshine:
It looks like `--parallel` doesn't work here. I'm going to close this issue since it seems to be worse than the original.

PR #983, opened 2021-11-06
Update dependencies

XhmikosR:
Mostly patch and minor dependencies except for `@types/node`, `mocha` and `@types/mocha`.

Flattens the package-lock and should fix some of the vulnerabilities along the way.


XhmikosR:
I also updated pacote but being a major version bump it made me wonder... they might have made it a major version bump due to the Node.js version bump https://github.com/npm/pacote/compare/v11.3.5...v12.0.2, in which case we should be safe.

.ncurc.js
```diff
@@ -8,9 +8,9 @@ module.exports = {
 
     // "Must use import to load ES Module"
     // These can be removed once the tests are converted to Typescript
+    'find-up',
```
XhmikosR, line 11: Might be worth switching to [escalade](https://github.com/lukeed/escalade) sync or async (personally I still believe using async will benefit us but anyway). I tried to make the switch but I'm not familiar with TS.

raineorshine: I am fine with either find-up or escalade. It is unlikely that I will upgrade all the tests to typescript since there are so many, so if we stay with `find-up` it will be frozen at `5.0.0`.

raineorshine (approved):

.ncurc.js
XhmikosR: Unfortunately, I can't make it work, but I think it'll benefit us even if we stick with the `escalade/sync`. Just thinking out loud in case you or someone else has time to make the switch later.

PR #985, opened 2021-11-06
buildReadme.js: remove unused variables

XhmikosR:
Unsure why these weren't caught by ESLint? Maybe we should adapt ESLint?

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "lint": "cross-env FORCE_COLOR=1 npm-run-all --parallel --aggregate-output lint:*",
     "lint:lockfile": "lockfile-lint",
     "lint:markdown": "markdownlint \"**/*.md\" --ignore node_modules --ignore build --config .markdownlint.js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "lint": "cross-env FORCE_COLOR=1 npm-run-all --parallel --aggregate-output lint:*",
     "lint:lockfile": "lockfile-lint",
     "lint:markdown": "markdownlint \"**/*.md\" --ignore node_modules --ignore build --config .markdownlint.js",
-    "lint:src": "eslint --cache --cache-location node_modules/.cache/.eslintcache --report-unused-disable-directives src",
+    "lint:src": "eslint --cache --cache-location node_modules/.cache/.eslintcache --report-unused-disable-directives src scripts",
```
XhmikosR, line 35: Shouldn't we just lint everything? The build dir is ignored AFAICT so in theory we could just use `.`.

raineorshine: Yes, that would be better. I created #993.

PR #986, opened 2021-11-06
runLocal: add missing `await`

XhmikosR:
This is flagged by [lgtm.com](https://lgtm.com/projects/g/raineorshine/npm-check-updates?mode=list), unsure if it's really needed and if it is, maybe some tests need to be adapted so that this is caught?


raineorshine (approved):
I'm not sure why it works with or without await. Seems reasonable though.

The `--format ownerChanged` feature was added by a contributor and does not currently have test coverage. I'm not sure that that particular option has enough of a userbase to justify a test, although I would certainly be open if someone was motivated to create a PR.

PR #993, opened 2021-11-09
Lint all source files

raineorshine:
Lint the entire project. Ignore the `build` directory.

Fixes lint errors in previously unlinted tests.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
     'plugin:import/typescript',
     'raine'
   ],
+  ignorePatterns: ['build'],
   overrides: [
     {
       files: ['**/*.ts'],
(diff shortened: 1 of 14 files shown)


.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
     'plugin:import/typescript',
     'raine'
   ],
+  ignorePatterns: ['build'],
```
XhmikosR, line 13: I'd swear that I thought by default ESLint respects .gitignore but apparently it doesn't.

Personally I just use `--ignore-path .gitignore` so that I don't have 2 places to handle these, but either way your call :)

raineorshine: It uses the same format as .gitignore but it doesn't read from it.

Ah, I see. That's smart.

PR #994, opened 2021-11-09
Update devDependencies and remove the unused chokicar-cli dev dependency

XhmikosR:
@raineorshine isn't chokidar-cli unused?


raineorshine:
> @raineorshine isn't chokidar-cli unused?

You're right, it's not

XhmikosR:
With the recent PRs, we've got down to 609 packages from 783 :)

Only escalade is left for later.

XhmikosR:
@raineorshine is husky still used? Also I wonder about source-map-support: should it be in dev dependencies?

XhmikosR:
This is what I meant: https://github.com/raineorshine/npm-check-updates/commit/f1c2e91ae90fcba42907303e9527d285c352e9f1

raineorshine:
> is husky still used?

Yes:  https://github.com/raineorshine/npm-check-updates/blob/main/.husky/pre-push

> Also I wonder about source-map-support: should it be in dependencies?

I tried to find an answer about whether `source-map-support` belongs in `dependencies` or `devDependencies`, but it wasn't clear to me [in the docs](https://github.com/evanw/node-source-map-support) or [on StackOverflow](https://stackoverflow.com/questions/42088007/is-there-source-map-support-for-typescript-in-node-nodemon). Maybe you can infer what is correct, otherwise you'd have to test it out both ways and confirm manually that it still works.

PR #996, opened 2021-11-11
tests: fix the mocha config key for extensions

midgleyc:
Additionally, move one test to typescript as proof.

It might be better to target ".test.js" and ".test.ts" and enable "recursive", which should prevent you from needing to specify the package manager tests by themselves (when appropriately renamed).

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
   },
   "mocha": {
     "check-leaks": true,
-    "extensions": [
+    "extension": [
       "js",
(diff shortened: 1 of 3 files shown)


raineorshine (approved):
Looks good!

midgleyc:
No, that's all for now.

The other comments were about things that could potentially be done in the future, but I think it's nicer to do it iteratively.

PR #997, opened 2021-11-11
migrate tests to TypeScript

midgleyc:
Following on from #996, I just did all of the remaining tests.

Some `src` changes, using more precise types or accepting nulls.

Not using `recursive` due to the `timeout` test being in the same folder as the rest. It does look like the problem that required its separation may no longer apply, though.


src/types.ts
```diff
@@ -288,3 +293,5 @@ export type Options = RunOptions & {
   peerDependencies?: Index<any>,
   rcConfigPath?: string,
 }
+
+export type PreInitOptions = Omit<Options, 'packageData'> & { packageData?: string | PackageFile }
```
raineorshine, line 297: Can we make adjustments so that `Options` can be used everywhere? It's confusing to have multiple options types.

test/cli-options.test.ts
```diff
@@ -1,14 +1,12 @@
-'use strict'
-
-const chai = require('chai')
-const cliOptions = require('../src/cli-options').default
+import chai from 'chai'
+import cliOptionsSorted from '../src/cli-options'
```
raineorshine, line 2: `cliOptions` is a better name than `cliOptionSorted` here since the sorting is not relevant and we can hide that detail from the test component.

raineorshine (changes requested):
Awesome!!! Thanks so much!

src/types.ts
midgleyc: The issue here was that:
* before `initOptions` `packageData` can be string or PackageFile
* after `initOptions` it's definitely a string
and adding `| PackageFile` to the definition on `Options` lead to some (impossible) type errors later down the line.

I can see if there's a better way of doing it: I could hard cast everywhere to string, but I thought this was cleaner (though I agree it's not great).

I could try having a base class that doesn't have `packageData` at all -- that could be used almost everywhere, I expect.

test/cli-options.test.ts
midgleyc: I agree, but I can't find the right syntax to import a default import with a different name. If you can, I'll be happy to change this!

src/types.ts
midgleyc: The "hard cast" versions adds casts to the calls to `runLocal` in `src/index.ts`:

```
await runLocal(pkgOptions, pkgData as string, pkgFile)
```

Let me know if you'd prefer that? I tend to avoid the casts where I can in TypeScript because it avoids the type checking, which can be an indicator that something's not right.

test/cli-options.test.ts
raineorshine:
```ts
import cliOptions from '../src/cli-options'
```

raineorshine: Fixed in fcf2ab1

src/types.ts
raineorshine: Excellent points! I see there isn't an easy answer.

We can rule out the cast, since that defeats the purpose of type checking, as you said.

What do you think about c44df47? It seems that when `run` accepts `Options` then all the issues go away.

midgleyc: That works and simplifies the code, but no longer tests the line 
```
...options.packageData && typeof options.packageData !== 'string' ? { packageData: JSON.stringify(options.packageData, null, 2) } : null,
```
in `initOptions.ts` (that does the PackageFile -> string conversion).

midgleyc: You could keep the test as it was, cast to `any`, and keep the line tested by the unit tests, but not by the types? As a sort-of deprecated feature?

test/cli-options.test.ts
midgleyc: Thanks, my IDE kept showing errors when I tried that, but it looks like it works :)

src/types.ts
raineorshine: You're right, and it gives `run` the wrong argument type. I do want to support `packageData: string | PackageFile` for module users.

I made another attempt in 3fcecba. I shift some options-related logic to `initOptions` and do the `PackageFile` conversion there. Let me know what you think.

midgleyc: Looks good to me! You might want to go back to `test/index.test.ts` and pass `packageData` as a `PackageFile` instead of a string.

I merged in `main` to get it building.

PR #998, opened 2021-11-11
types: prefer DT types

midgleyc:
Where possible, use types from `DefinitelyTyped` instead of redefining.


raineorshine (approved):
Thanks!

PR #999, opened 2021-11-13
chore: update commander to latest major

midgleyc:
Now that version 12 requires node 12, can bump commander to 8, which also requires node 12.


raineorshine (approved):
Great, thanks!

PR #1001, opened 2021-11-18
fix getRepoUrl undefined crash

CreativeTechGuy:
`fromUrl` can return `undefined` and would crash the script if the git directory was malformed or unrecognized (eg: for an internal/private git repo). This change handles the undefined case.


raineorshine (approved):
Thanks!

PR #1015, opened 2021-12-08
fix transforms strictSSL to boolean when used as a env var

efbenson:
For some reason when using npm scripts to run ncu npm sets the env var npm_config_ssl_strict="",  to override that we are using cross-env which works, but the env var is a string and npm will not use that as a value, "false" is ignored but false the boolean works.  This PR will add the ability to cast properties to boolean as needed for npm.  If there is a better way to do it you can throw this out.


efbenson:
Sorry my typescript is terrible, Its not working, but it looks like the issue is the matching logic.  I added a console log to the stringToBoolean function in the code and its not getting called.
```  
const stringToBoolean = (s) => {console.log(`called ${s}`); return s && s !== 'false' && s !== '0'}; 
```

I added a console dir to see the output here:

``` 
const config = {};
    libnpmconfig_1.default.read().forEach((value, key) => {
        // replace env ${VARS} in strings with the process.env value
        const normalizedValue = typeof value !== 'string' ? value
            // parse stringified booleans
            : value in booleanKeys ? stringToBoolean(value)
                : value.replace(/\${([^}]+)}/, (_, envVar) => process.env[envVar]);
        const { [key]: pacoteKey } = npmConfigToPacoteMap;
        if (lodash_1.default.isString(pacoteKey)) {
            config[pacoteKey] = normalizedValue;
        }
        else if (lodash_1.default.isFunction(pacoteKey)) {
            lodash_1.default.assign(config, pacoteKey(normalizedValue));
        }
        else {
            config[key.match(/^[a-z]/i) ? lodash_1.default.camelCase(key) : key] = normalizedValue;
        }
    });
    console.dir({config})
    config.cache = false;
```

Here is some of the output
```
    authType: 'legacy',
    nodeVersion: '14.16.1',
    tag: 'latest',
    gitTagVersion: 'true',
    commitHooks: 'true',
    scriptShell: '',
    shrinkwrap: 'true',
    strictSSL: 'false',
    fetchRetryFactor: '10',
    saveExact: '',
    dev: ''
```

So the strictSSL is there just not getting detected for conversion

raineorshine:
Okay, it could be a case mismatch. I converted it to lowercase and published to `v12.0.5-0`. Will you try `npx npm-check-updates@next` again?

If that doesn't work, you could try `console.log`ing the values that are read from `libnpmconfig` and posting theme here, so I can try to reproduce.

efbenson:
Found the issue just needed to make one small change

```
    const normalizedValue = typeof value !== 'string' ? value
      // parse stringified booleans
      : value in booleanKeys ? stringToBoolean(value)
      : value.replace(/\${([^}]+)}/, (_, envVar) =>
        process.env[envVar] as string
      )
```
You are checking if the value is in booleanKeys, you want to check if the key is in booleanKeys.  When I changed that in my @next local copy it worked.

efbenson:
Works out of the box!

raineorshine:
Thanks for your help!

Published to `v12.0.5`.

PR #1021, opened 2021-12-29
Feature proposal: The runGlobal function now returns the Index<string> obj

Elschnagoo:
# The runGlobal function now returns the Index<string> obj of upgraded packages

While checking all global packages i recognize that the runGlobal function only return a void for global report.

> Feature: Module/Programmatic Usage now can use the the global scan report.

### Before
```typescript 
  const res = await run({
    global: true
  });
  console.log(res); // will alwas be undefined
```
### Now
```typescript 
  const res = await run({
    global: true
  });
  console.log(res); // res : Index<string>
```


raineorshine:
Looks good!

Issue #1034, opened 2022-01-21
[FR] Introduce `safe` option


Primajin:
I am currently using the `--filterVersion` option in order to run the checks only on non-pinned versions.
However I need to run it twice since
```
> npm-check-updates --filterVersion "/^[~^]/" -t minor

Checking C:\mypath\package.json
[====================] 130/130 100%

 @types/node                       ^17.0.9  →  ^17.0.10
 @types/react-router               ^5.1.17  →   ^5.1.18
 @typescript-eslint/eslint-plugin   ~5.9.0  →   ~5.10.0
 @typescript-eslint/parser          ~5.9.0  →   ~5.10.0
 eslint-plugin-jest                ~25.3.4  →   ~25.7.0
 npm-check-updates                 ^12.1.0  →   ^12.2.0
 typescript                         ^4.5.4  →    ^4.5.5
```
gives me not exactly want I want.
I actually don't want to bump a `minor` (`^`) on @typescript-eslint/parser but only a `patch` (`~`)

So instead I changed that to
```
> npm-check-updates --filterVersion "/^[\^]/" -t minor && npm-check-updates --filterVersion "/^[~]/" -t patch

Checking C:\mypath\package.json
[====================] 127/127 100%

 @types/node          ^17.0.9  →  ^17.0.10
 @types/react-router  ^5.1.17  →   ^5.1.18
 npm-check-updates    ^12.1.0  →   ^12.2.0
 typescript            ^4.5.4  →    ^4.5.5

Run ncu -u to upgrade package.json

Checking C:\mypath\package.json
[====================] 3/3 100%

 @typescript-eslint/eslint-plugin  ~5.9.0  →  ~5.9.1
 @typescript-eslint/parser         ~5.9.0  →  ~5.9.1

Run ncu -u to upgrade package.json
```

Which gives me the "correct" minor and patches for the ones I want and skips pinned versions.

Though I wonder if we could introduce a new target version like `safe` (naming is hard, maybe you have a better one) that does that automatically:

1. Skips all pinned versions
2. Checks for minor on caret #581 / #950
3. Checks for patch on tilde #581 / #950
4. **maybe even skips versions that start with 0** #958 

What do you think?

EDIT: Apparently I missed #581 / #950 - however if you like I can try to take a stab at it 👍🏻

_Originally posted by @Primajin in https://github.com/raineorshine/npm-check-updates/issues/958#issuecomment-1018305399_

raineorshine:
[...]

PR #1050, opened 2022-02-11
Feature proposal: provide function to filter

Primajin:
While thinking about how to solve https://github.com/raineorshine/npm-check-updates/issues/1034 I was playing around a bit and came up with this draft.

The idea is that one can provide a function - for example via the `.ncurc.js`

```js
target: (s) => { return s === matchesMyComplicatedChecksHere }
```

and then only get's back the matching packages. An alternative to a fancy regex where it can be solved more in detail and programmatically.

test provided:

```js
/*dependencies: {
  mocha: '1.2'
  lodash: '^3.9.3'
  moment: '^1.0.0'
  chalk: '^1.1.0'
  bluebird: '^1.0.0'
}*/
    
it('filter dependencies by function', () => {
  getCurrentDependencies(deps, { filter: (s:string) => s.startsWith('m') }).should.eql({
    mocha: '1.2',
    moment: '^1.0.0'
  })
})
```

What do you think?


raineorshine (changes requested):
Looks great!

Let's add tests for `reject`, `filterVersion`, and `rejectVersion` as well.

Primajin:
Hello there 👋🏻 😄 
I am currently adding the tests for `reject`, `filterVersion`, and `rejectVersion`:

---

I found that `test/getCurrentDependencies.test.ts` has tests for `filter` and `reject` - but the tests for `filterVersion` and `rejectVersion` look very differently and are inside `test/index.test.ts`.
Are you ok if I add a couple more tests to `test/getCurrentDependencies.test.ts` in the same style as `filter` and `reject` but for `filterVersion` and `rejectVersion`?

---

Then I found that filterVersion accepts
```ts
filterVersion?: string | string[] | RegExp | RegExp[],
```
while `rejectVersion` only accepts:
```ts
rejectVersion?: string | string[] | RegExp,
```
Is that deliberate? If not, are you OK if I type them all with `FilterRejectPattern` ?
Does the JSDoc need to be updated as well?

---

Speaking of JSDoc - I'm not a native speaker so up to you to decide, but my IDE was nagging me about a missing `the` - shall I add it while at it?
the-highest

raineorshine:
> I found that `test/getCurrentDependencies.test.ts` has tests for `filter` and `reject` - but the tests for `filterVersion` and `rejectVersion` look very differently and are inside `test/index.test.ts`.
> Are you ok if I add a couple more tests to `test/getCurrentDependencies.test.ts` in the same style as `filter` and `reject` but for `filterVersion` and `rejectVersion`?

Sure!

> Is that deliberate? If not, are you OK if I type them all with `FilterRejectPattern` ?
> Does the JSDoc need to be updated as well?

It is not deliberate. I think they should have the same type. Thanks!

> Speaking of JSDoc - I'm not a native speaker so up to you to decide, but my IDE was nagging me about a missing `the` - shall I add it while at it?

You are right! That code was written by a non-native speak also, but your IDE apparently is a native speaker :).

> I would make them all `string | string[] | RegExp | RegExp[] | Function` now - ok?

Yes, I agree.

Primajin:
> * [X]  Update [cli-options](https://github.com/raineorshine/npm-check-updates/blob/main/src/cli-options.ts).
> * [X]  Let's add tests for `reject`, `filterVersion`, and `rejectVersion` as well. I think they use the same predicate so they automatically get support for filter functions with this PR.

done 😄 

> * [ ]  Pass `name` and `versionRange` to the predicate function as described here: [#1034 (comment)](https://github.com/raineorshine/npm-check-updates/issues/1034#issuecomment-1019507970). This will make it easier for developers to write conditions using the dependency version. In effect, `filter` and `filterVersion` will be identical when a function is passed. I think it's simpler than having two separate predicate signatures.

Should this be part of this MR or a separate one - sunsetting `filterVersion`/`rejectVersion` feels like a breaking change that would require a major version bump - while the rest could be a new feature with a minor bump so far.
What do you think?

raineorshine:
> > * [ ]   Pass `name` and `versionRange` to the predicate function as described here: [#1034 (comment)](https://github.com/raineorshine/npm-check-updates/issues/1034#issuecomment-1019507970). This will make it easier for developers to write conditions using the dependency version. In effect, `filter` and `filterVersion` will be identical when a function is passed. I think it's simpler than having two separate predicate signatures.
> 
> Should this be part of this MR or a separate one - sunsetting `filterVersion`/`rejectVersion` feels like a breaking change that would require a major version bump - while the rest could be a new feature with a minor bump so far. What do you think?

It should be part of this PR, since I don't want to publish the feature with the wrong predicate signature.

My intention was to keep `filterVersion` and `rejectVersion`, so this won't be a breaking change. It's just the function signature of the predicate function that is in question. You just need to pass the result of `semverutils.parseRange(versionRange)` as the second argument [when the predicate is called](https://github.com/raineorshine/npm-check-updates/pull/1050/files#diff-036fd0e13c411beac42898792ba3dac30b7cb9fb0726cf3f87a683db966283e0R46), so that the developer can write conditions based on the version number.

raineorshine:
Yes, but `version` should be of type `semverutils.SemVer[]`. We parse the version range for the developer to make it even easier to write conditions.

You can see `semverutils.parseRange` used throughout the codebase.

See: https://git.coolaj86.com/coolaj86/semver-utils.js#semverutils-parserange-rangestring

raineorshine:
This is what I had in mind: 13cafc5

The `filterVersion` predicate function signature doesn't match `filter`, but I'm going to leave it for now since the `composeFilter` pipeline needs major refactoring to support this.

It looks ready to merge from my perspective. Let me know if I'm missing anything.

Primajin:
Ahh cool thanks! 👍🏻  Let me take a final glimpse at it on Monday - thanks for your support!

Primajin:
OK couldn't wait until Monday 😅 

Added a final cleanup commit with tiny changes:

- `src/lib/filterAndReject.ts:4` - we don't need to import all of semverutils here, we can just get the two things we need
- `src/types.ts:28` - `FilterFunction` isn't used outside any more, no need to export
- `test/getCurrentDependencies.test.ts` - both of us added dangling commas, while I personally like them, it looks like the rest of the codebase doesn't use them - so let's adhere to the common style 😉 

Very excited about this, it was a lot of fun - thank you!

raineorshine (approved):

raineorshine:
Thanks!

> * `test/getCurrentDependencies.test.ts` - both of us added dangling commas, while I personally like them, it looks like the rest of the codebase doesn't use them - so let's adhere to the common style 😉

I need to add Prettier...

raineorshine:
Published to `v12.4.0`

Note that I am going to patch the `filterVersion` predicate when I have time to do the necessary refactoring. It should have the same signature as the `filter` predicate. I'll update the README and publish an official release then.

Issue #958, opened 2021-10-21
[FR] --target minor or patch should exclude 0.x.x release


ashidaharo:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [x] I am using the latest version of `npm-check-updates`
- [x] I am using `node >= 10.17`

---------------------------

## Steps to Reproduce
```
ncu --target minor
```
or
```
ncu --target patch
```
## Current Behavior
Both results include the new release of 0.x.x.
like
```
ncu --target minor
 axios                              ^0.22.0  →   ^0.23.0
```
## Expected Behavior
0.x.x means that all release MAY contain BREAKING CHANGES.
So, release of 0.x.x should be  treated as a MAJOR release.
This should be a feasible feature, since it is correctly displayed as such when --color is set.

ashidaharo:
But... this fix seems to me to be a breaking change.
I can't decide whether to include it in the patch version as a bug fix or in the next major version update as a breaking change, so I guess I can't send you a PR for it...

ashidaharo:
> @ashidaharo I've just discovered it too) It should be something that was left intentionally or a new bug (but I think it was always like this). I should not think nobody hasn't noticed it during all this time, for it's a very important behavior. But what was the intension?

When I found the source of the problem, I confirmed that this was a behavior that had always existed, whether it was intentional or not. I think it's probably just that the person who originally wrote it didn't pay attention to exceptions like '0.x.x'.

However, the current behavior has been confirmed as a specification by the following tests.
https://github.com/raineorshine/npm-check-updates/blob/26ceb1e5aa81592adbd0d45a68ad4c7cbfbb7a62/test/version-util.test.js#L393-L396
For that reason, I considered the current behavior of this library to be as per the specification of this package, although it does not follow the SemVer specification. Therefore, I thought that fixing this problem should be treated as a destructive change, not a bug.

lunany188:
I've just noticed that the version has been upgraded, so I came to remind everyone.
The problem is still unresolved in any way, and there continues to be a difference in the actual behavior of coloring and `--target`.

raineorshine:
The coloring uses semver, while `target` refers simply to the digit (whatever that's worth). 

I imagine that `--target semver` is what a lot of people want (#581 for historical discussion).

kokushkin:
> target refers simply to the digit (whatever that's worth)

but this is how semver also treats it. Minor in semver is just the second digit!

[...]

Issue #950, opened 2021-09-21
feature request: target that respects range grammar in package.json (~1.0.0, ^1.0.0, etc)


ex-nerd:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [x] I am using the latest version of `npm-check-updates`
- [x] I am using `node >= 10.17`

## Steps to Reproduce

The current `--target` options lack a way to say "upgrade to the latest values allowed by my package.json" (following [semver range syntax](https://github.com/npm/node-semver#versions)). This would basically be the equivalent of `rm -rf node_modules && npm install` and then finding the versions that were installed to put them back in `package.json`.

This isn't simply a matter of using `--target patch` because I often have mixed requirements in package.json where I want to allow only `patch` updates to some packages, but want `minor` updates on others.

I thought this was an issue with how `--minimal` works (since I interpreted that as "the update the minimal amount" instead of "update a minimal number of packages"), but found https://github.com/raineorshine/npm-check-updates/issues/746 saying this is expected behavior. So instead of a bug, this is a feature request.

## Current Behavior

package1 `~1.0.0` → `~1.2.0`
package2 `^1.0.0` → `^2.0.0`
package3 `1.0.0` → `1.1.0`

## Expected Behavior

package1 `~1.0.0` → `~1.0.1`
package2 `^1.0.0` → `^1.2.0`
package3 `1.0.0` → `1.0.0`

raineorshine:
Thanks for the suggestion! This is a great idea. For most users I would suggest using [npm update](https://docs.npmjs.com/cli/v7/commands/npm-update) as designed to achieve this behavior with semver. However, you are not the first to want greater control over the version numbers in the package.json.

The proposed addition would be a welcome contribution, it's just in need of a contributor. For historical discussion, see #581.

raineorshine:
As of `v12.5.0`, you can specify a custom target function in your .ncurc.js file, or when importing npm-check-updates as a module:

```js
/** Custom target.
  @param dependencyName The name of the dependency.
  @param parsedVersion A parsed Semver object from semver-utils.
    (See https://git.coolaj86.com/coolaj86/semver-utils.js#semverutils-parse-semverstring)
  @returns One of the valid target values (specified in the table above).
*/
target: (dependencyName, [{ semver, version, operator, major, minor, patch, release, build }]) => {
  if (major === 0) return 'minor'
  return 'latest'
}
```

[...]

Issue #581, opened 2019-09-11
Option to bump existing ranges to reflect latest semver version


medikoo:
- [x] I have read the list of [known issues](https://github.com/tjunnone/npm-check-updates#known-issues) before filing this issue.
- [x] I have searched for [similiar issues](https://github.com/tjunnone/npm-check-updates/issues) before filing this issue.

* node version: v12.10.0
* npm version: 6.10.3
* npm-check-updates version: 3.1.22

---------------------------

#573 seems related, and #159 seems about same thing

I'm after an option that will bump existing ranges in `package.json`, so they point to the latest version (within a reflected range).

Currently `ncu` does that well, only if we use `^` and rely on latest major of given dependency.

However if there's a package where we rely on non-latest major, it automatically updates to latest major (which I find unwanted, as any major version upgrade should rather be done manually ensuring breaking changes are handled).

#### Explaining on example

There's a package where:
- Latest version is `3.2.5`
- Latest v2 version is `2.8.5`
- Latest v2.4 version is `2.4.8`

##### Case 1

Range in is package.json is `^2.4.5`.

I'd love ncu to upgrade the range to `^2.8.5` (and not `^3.2.5` as it does  now).

##### Case 2

Range in is package.json is `~2.4.5`.

I'd love ncu to upgrade the range to `~2.4.8` (and not `^3.2.5` as it does  now).

--- 

In  #159 it was suggested that `npm update` does that. That's not true

`npm update` updates package.json only if:
1. Given package __is installed in `node_modules`__
2.  Package __is installed at version lower than latest as supported by range__.

So in above scenario if  in `node_modules` package is already installed at  `2.8.5` (or is not installed at all), there's no way to force`npm update` to update the range in `package.json`.

raineorshine:
Thank you for the detailed explanation! And thank you for looking through existing issues.

I marked this as an enhancement and am open to PR's or +1's to signal interest if others are wanting this option.

As a side note, I'm somewhat unclear what benefit there is to having the number changed in the package.json when the version range gives you the control you are looking for. This likely does not change the status of this ticket, but I thought I would mention that this came up for me.

medikoo:
> what benefit there is to having the number changed in the package.json when the version range gives you the control you are looking for. 

Bumping the ranges ensures that non-latest (in some cases buggy) versions of package dependencies will not be installed.
[...]

PR #1052, opened 2022-02-23
Target as a function

### LINKED ISSUE #1034: [FR] Introduce `safe` option

Primajin:
Closes #1034
Related: #958, #950, #581

The idea is to pass a function to target so that one can for example create a version of `semver`-like target:
- return `'minor'` for `'^'`
- return `'patch'` for `'~'`

```js
target: (name, [{ operator }]) => operator === '^' ? 'minor' : operator === '~' ? 'patch' : 'latest'
```

(with the recent introduction in #1050 of filter as a function both can be combined to filter out pinned versions and versions that start with a `0` - re: #958) to fully achieve "safe" semver like versioning 🎉 

```js
target: (name, [{ operator }]) => operator === '^' ? 'minor' : operator === '~' ? 'patch' : 'latest',
filter: (_, [{ major, operator }]) => !(major === '0' || major === undefined || operator === undefined)
```

### DIFF
--- a/src/lib/queryVersions.ts
+++ b/src/lib/queryVersions.ts
@@ -3,7 +3,7 @@ import cint from 'cint'
 import chalk from 'chalk'
 import pMap from 'p-map'
 import ProgressBar from 'progress'
-import { supportedVersionTargets, targetFallback } from '../constants'
+import { supportedVersionTargets } from '../constants'
 import getPackageManager from './getPackageManager'
 import packageManagers from '../package-managers'
 import { createNpmAlias, isGithubUrl, isPre, parseNpmAlias } from '../version-util'
@@ -18,7 +18,7 @@ import { GetVersion, Index, Options, Version, VersionSpec } from '../types'
  */
 async function queryVersions(packageMap: Index<VersionSpec>, options: Options = {}) {
 
-  const target = options.target || targetFallback
+  const target = options.target || 'latest'
   const packageList = Object.keys(packageMap)
   const packageManager = getPackageManager(options.packageManager)
 
--- a/src/lib/queryVersions.ts
+++ b/src/lib/queryVersions.ts
@@ -28,14 +28,6 @@ async function queryVersions(packageMap: Index<VersionSpec>, options: Options = 
     bar.render()
   }
 
-  // set the getPackageVersion function from options.target
-  // TODO: Remove "as GetVersion" and fix types
-  const getPackageVersion = packageManager[target() as keyof typeof packageManager] as GetVersion
-  if (!getPackageVersion) {
-    const packageManagerSupportedVersionTargets = supportedVersionTargets.filter(t => t in packageManager)
-    return Promise.reject(new Error(`Unsupported target "${target}" for ${options.packageManager || 'npm'}. Supported version targets are: ${packageManagerSupportedVersionTargets.join(', ')}`))
-  }
-
   /**
    * Ignore 404 errors from getPackageVersion by having them return `null`
    * instead of rejecting.
@@ -47,6 +39,7 @@ async function queryVersions(packageMap: Index<VersionSpec>, options: Options = 
 
     const npmAlias = parseNpmAlias(packageMap[dep])
     const [name, version] = npmAlias || [dep, packageMap[dep]]
+    const targetResult = typeof target === 'string' ? target : target(name, version)
 
     let versionNew: Version | null = null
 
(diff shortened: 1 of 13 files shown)


test/index.test.ts
```diff
@@ -196,7 +196,7 @@ describe('run', function () {
           'ncu-mock-pre': '1.0.0'
         }
       }),
-      target: 'newest'
+      target: () => 'newest'
```
raineorshine, line 199: You'll want to leave these tests as-is. It should support static target or function target, just like `filter`.

raineorshine (changes requested):

src/lib/queryVersions.ts
```diff
@@ -30,7 +30,7 @@ async function queryVersions(packageMap: Index<VersionSpec>, options: Options =
 
   // set the getPackageVersion function from options.target
   // TODO: Remove "as GetVersion" and fix types
-  const getPackageVersion = packageManager[target as keyof typeof packageManager] as GetVersion
+  const getPackageVersion = packageManager[target() as keyof typeof packageManager] as GetVersion
```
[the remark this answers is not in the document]
Primajin: Alrighty - moving it in there did the trick!
So far I only pass `version` in as a string - could not yet wrap my head around SemVer utils though.

I started writing tests but there are some questions popping up:
1. The tests seem to run against live npm - how do we make sure that the expected latest version never changes other than choosing an old version and hope it doesn't get bumped? Can we not instead just mock the npm response?
2. I couldn't reliably run the tests locally in `test/index.test.ts` - it always complained that `pkgData` is just a plain object and has no `.should.have.stuff()` - how do I run them locally as single tests?

test/index.test.ts
Primajin: Since it should support a static target as well there is actually no big benefit to set it to a function in the initOptions then, or? 🤔

raineorshine:
> So far I only pass `version` in as a string - could not yet wrap my head around SemVer utils though.

Use the same approach [we used for `filter`](https://github.com/raineorshine/npm-check-updates/blob/d5f3778f7447e66f0f7048661d481d742e7802b2/src/lib/filterAndReject.ts#L47):

```ts
filterFunction(name, semverutils.parseRange(version))
```

> 1. The tests seem to run against live npm - how do we make sure that the expected latest version never changes other than choosing an old version and hope it doesn't get bumped? Can we not instead just mock the npm response?

I currently test against a package I control, https://github.com/raineorshine/ncu-test-v2, which will always have `v2` as the latest. There is a [suite of test packages](https://github.com/raineorshine/ncu-test-v2) used for testing other types of upgrades.

I think mocking the npm response would be a much better approach. I wish I had done that from the beginning. You're welcome to use either approach.

> 2\. I couldn't reliably run the tests locally in `test/index.test.ts` without running everything - it always complained that `pkgData` is just a plain object and has no `.should.have.stuff()` - how do I run them locally as single tests?

I did some digging and turns out that was a bug! I was missing `chai.should()` from the test file. I added it in . It should work with that.

raineorshine (approved):

Primajin:
Alrighty ready for review 🤞🏻 🎉

Primajin:
One last thing - should we somehow add the two filter + target as a new param/cli option?
As in `semver = true` sets both - so that people don't need to come up with the two functions I posted in the top?

```js
target: (name, [{ operator }]) => operator === '^' ? 'minor' : operator === '~' ? 'patch' : 'latest',
filter: (_, [{ major, operator }]) => !(major === '0' || major === undefined || operator === undefined)
```

raineorshine:
> One last thing - should we somehow add the two filter + target as a new param/cli option?
> As in `semver = true` sets both - so that people don't need to come up with the two functions I posted in the top?

Let's discuss in a separate issue. There's a bit more complexity to that.

raineorshine (approved):

raineorshine (approved):

raineorshine:
Published in `v12.5.0`

Primajin:
Oh by focusing only on the possible errors and tests etc. - I just realized we might want to tweak the success message 🤦🏻‍♂️ 

image

Maybe when it's a function we just call it "given" --> `All dependencies match the given package versions :)` ?

raineorshine:
Oops! Yes, I forgot it was used as a string. 

Let's say "target package versions".

PR #1053, opened 2022-02-28
Fix the bad code snippet

dcyuki:
node example code syntax error


raineorshine:
Thanks! I've taken the opportunity to update the code snippet for node => 14.

PR #1066, opened 2022-03-03
Fix success message when target is not a string

Primajin:
Fixes https://github.com/raineorshine/npm-check-updates/pull/1052#issuecomment-1055439710


Primajin:
🎉

raineorshine:
Published in `v12.5.1`

PR #1071, opened 2022-03-16
fix: allow updating to eslint 8

midgleyc:
Fix lints added in eslint 8.

`eslint-plugin-promise` has allowed eslint 8 for a while (~since end of December).

Correct comments in `.ncurc.js`: while node ESM support was initially targeted for TypeScript 4.6, it's been delayed, and is now targeted for TypeScript 4.7.


raineorshine (approved):
Looks good, thanks!

PR #1074, opened 2022-04-01
fix: updates hosted-git-info to version 5.x

bbeesley:
Version 4.x of hosted-git-info has a dependency on a broken version of the lru-cache module, which causes the error "ReferenceError: AbortController is not defined". The lru-cache module has been fixed, and hosted-g
it-info has been updated to use the fixed version, but that update is only on version 5.x of hosted-git-info. This change updates to version 5.x of hosted-git-info to pull in that fix, although 5.x is a breaking chan
ge, the only breaking aspect is that versions of node prior to 12 LTS are no longer supported. This should not be an issue as npm-check-updates doesn't support older versions of node.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "fp-and-or": "^0.1.3",
         "get-stdin": "^8.0.0",
         "globby": "^11.0.4",
-        "hosted-git-info": "^4.1.0",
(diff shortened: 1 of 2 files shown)


raineorshine:
Published in `v12.5.5`

Issue #626, opened 2020-02-08
Running ncu through scripts showing weather


simkessy:
I'm trying to add ncu to our build steps so that whenever we're building users see current package status so they might be more willing to update regularly. 

in my package scripts I added this: 

`    "deps": "echo \"\\x1b[30m\\x1b[103mOUTDATED DEPENDENCIES\\x1b[0m\" & ncu",`

Problem is, it's running a weather app instead of showing me outdated packages?

stoically:
Probably some local binary that has precedence - could try `npm-check-updates` instead of `ncu`.

raineorshine:
@simkessy Run `which ncu` or `type ncu` from the package directory to see what command is aliased.

abensur:
@simkessy You might have installed [ncu](https://www.npmjs.com/package/ncu).
Please note that this repository npm's entry is `npm-check-updates` and not `ncu`. 
You should try `npm -g uninstall ncu && npm -g install npm-check-updates`

Drecu:
I run into the same problem but figured it out with the weather app. 
I am also struggling integrating npm-check-updates into our CI build steps.
I simply can't invoke ncu. It can not find the command inside the gitlab CI stage. 
And when i use it as a dev dependency i cant invoke it like `npm run-script ncu` which would be the solution. Any ideas?
Sorry if this is the wrong place for my question but i didnt find anything related since the past hours of my research on get this going.

xoxefdp:
i have ran into the same issue, but in my case im not installing it, i just ran "npx ncu" when the weather thing was called, if there is a precedence of other package for "ncu" in npm repository please update docs so anybody can call the correct one.

"npx npm-check-updates" works fine though.

raineorshine:
"ncu" is a `bin` alias for `npm-check-updates`. `npx` uses the package name, which is `npm-check-updates`. I can see how that would be confusing.

Updated README in deb4356.

bivainis:
I ended here by running `npx ncu` before installing `npm-check-updates` globally. This installed `ncu` package, which even has this deprecation warning:
```
npm WARN <...> and this is not npm-check-updates
```

Ran `npx npm-check-updates`, which installed the package I want.  
Ran `ncu` directly instead of `npx ncu`, since it was in .bin after install.

PR #1088, opened 2022-04-05
Use npx in upgrade message if invoked via npx

dnicolson:
This pull request checks if npm-check-updates was invoked via `npx npm-check-updates` and adjusts the package.json upgrade message.

This replaces what appears to be the only instance of `ncu` which is part of a complete command in output, it doesn't seem to make sense to replace other instances of `ncu` in output and help messages.

Relates to https://github.com/raineorshine/npm-check-updates/issues/626.


raineorshine:
Looks good! Thanks!

PR #1095, opened 2022-04-07
Fix example

p-chan:
The run function appears to be exported by default.


raineorshine (changes requested):
Thanks for the suggestion. The modification you made causes an error: `TypeError: ncu is not a function`. This is probably due to the way that typescript is configured.

If you can find a way to export `run` as the default, and maintain compatibility with node >= 12, I'd be open to that.

raineorshine:
Closing due to inactivity

PR #1096, opened 2022-04-07
Set permissions for GitHub actions

naveensrinivasan:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,6 +13,9 @@ env:
   FORCE_COLOR: 2
   NODE: 16
 
+permissions:
+  contents: read
+
 jobs:
   lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,8 +13,14 @@ env:
   FORCE_COLOR: 2
   NODE_COV: 16
 
+permissions:
+  contents: read
+
 jobs:
   run:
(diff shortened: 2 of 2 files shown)


raineorshine:
Thanks for your contribution! Looks like one of the tests fails on node 12 and 14.

naveensrinivasan:
This is a YAML change shouldn't cause the tests fail. 

Could you please rerun those tests?

raineorshine:
I've already re-run them a couple times. If you'd like to make changes or trigger a re-run, you can do a force push.

raineorshine:
Definitely seems to be working now. I'll run it a couple more times to confirm.

raineorshine:
Three successes in a row—I'm convinced! Thanks :)

PR #1097, opened 2022-04-12
chore: type Target using string literals and export type RunOptions

jensbodal:
I'm using this tool programmatically to maintain package versions in a mono repo and had to provide my own types for RunOptions and additional typing for the explicit strings that are allowed for `Target`.  Would be great if I can just grab these from the library.

### DIFF
--- a/src/lib/initOptions.ts
+++ b/src/lib/initOptions.ts
@@ -6,7 +6,7 @@ import { deepPatternPrefix } from '../constants'
 import programError from './programError'
 import getPackageFileName from './getPackageFileName'
 import { print } from '../logging'
-import { Options, RunOptions } from '../types'
+import { Options, RunOptions, Target } from '../types'
 
 /** Initializes, validates, sets defaults, and consolidates program options. */
 function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } = {}): Options {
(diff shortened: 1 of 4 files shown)


src/lib/initOptions.ts
```diff
@@ -76,9 +76,9 @@ function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } = {}): O
     programError(options, chalk.red(`Cannot specify both --packageFile and --deep. --deep is an alias for --packageFile '${deepPatternPrefix}package.json'`))
   }
 
-  const target = options.newest ? 'newest'
+  const target: Target = options.newest ? 'newest'
     : options.greatest ? 'greatest'
-    : options.target || options.semverLevel || 'latest'
+    : options.target || (options as any).semverLevel || 'latest'
```
jensbodal, line 81: I could type this as `{ semverLevel: string }` instead if preferred, but I figured since it's deprecated anyways it will probably go away soon.

jensbodal: And actually if I type that properly then the check will fail because `string` is not assignable to type `Target | undefined`.  So leaving it as `any` to support the deprecated usage is easiest, but I can figure out another typing if needed.

jensbodal:
Didn't realize linting wasn't run as part of `build` script, so I just updated that

raineorshine:
Thanks!

raineorshine:
Published in `v12.5.9`

PR #1106, opened 2022-04-30
Breaking change! camelcase `logLevel` option

ath0mas:
CLI option `loglevel` is the only one not using camelcase, but lowercase.

This PR changes it to camelcase: `--logLevel`, `"logLevel": …,`.


raineorshine:
`--loglevel` is intentionally lowercase to match `npm`:
https://docs.npmjs.com/cli/v8/using-npm/logging#setting-log-levels

I recognize the awkwardness of the inconsistency though. I would be nice of Commander had support for case-insensitive options.

ath0mas:
Fine, I understand. And so I close this PR.

PR #1113, opened 2022-05-19
fix: allow eslint upgrade now node 14 is required

midgleyc:
#1071 jumped the gun slightly, but now that node 14 is required (and `eslint-config-standard` is released) that unblocks eslint-related upgrades properly.

Sorry for the earlier issues.


raineorshine (approved):
Great, thanks!

PR #1114, opened 2022-05-19
fix: remove semverLevel completely

midgleyc:
`semverLevel` was removed in 13.0.0, and cannot be passed in using the CLI or a typed `Options`. Remove it completely.


raineorshine (approved):
Good catch!

Issue #1108, opened 2022-05-10
Replace libnpmconfig dependency


o-mdr:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [x] I am using the latest version of `npm-check-updates`
- [x] I am using `node >= 12`

---------------------------

## Steps to Reproduce

Installing ncu via `npm install -g npm-check-updates` gives me this error
> npm WARN deprecated libnpmconfig@1.2.1: This module is not used anymore. npm config is parsed by npm itself and by @npmcli/config

## Current Behavior
An older library is used to pull config

## Expected Behavior
`@npmcli/config` is used rather than `libnpmconfig`

I am on:
```
Node: v18.1.0
npm: 8.8.0
ncu: 12.5.11
```

raineorshine:
Thanks for reporting. Would you like to open a PR to make this change?

Zamiell:
I took a quick look and it seems like upgrading is non-trivial.

First, there are no typescript definitions for @npmcli/config, so you would have to create something.

Second, the `Config` export is async, and the current code is sync, so it's not just a simple find and replace.

Third, there seems to be legacy code that is strongly coupled with libnpmconfig:

```ts
  // needed until pacote supports full npm config compatibility
  // See: https://github.com/zkat/pacote/issues/156
  const config: Index<string | boolean> = {}
  // libnpmconfig incorrectly (?) ignores NPM_CONFIG_USERCONFIG because it is always overridden by the default builtin.userconfig
  // set userconfig manually so that it is prioritized
  const builtinsWithUserConfig = {
    userconfig: process.env.npm_config_userconfig || process.env.NPM_CONFIG_USERCONFIG
  }
```

Raine, this is probably something that you should take a look at.

raineorshine:
Thanks for the synopsis, @Zamiell.

I have brought the `libnpmconfig` logic internal to `npm-check-updates` in `v13.0.2` so there is no longer a deprecation warning.

As for integrating `@npmcli/config`, it seems nice in theory, but is a lot of risk for little gain. If I get reports that the `libnpmconfig` logic is not working correctly, or if new versions of `npm` add new features that are only supported by `@npmcli/config`, I will consider replacing it.

Zamiell:
Fantastic work, thanks @raineorshine .

[...]

PR #1115, opened 2022-05-21
Use internal libnpmconfig since the npm package is deprecated

raineorshine:
#1108


raineorshine:
Published to `v13.0.2`

PR #1116, opened 2022-05-21
Use internal figgy-pudding since the npm package is soon to be deprecated

raineorshine:
#1108


raineorshine:
Published to `v13.0.3`

PR #1117, opened 2022-05-22
Print error if `ncu --global` is used with Volta installed

### LINKED ISSUE #1072: Wrong NPM package is checked (NPM is installed using Volta)
- [x] I am using `node >= 12`

---------------------------

> Not sure it's really a bug, can be also a feature request. Because Volta is using a kind of shims/executable proxy for managing installed tools.

In case Node.js/NPM are managing using [Volta](https://volta.sh/), `ncu -g` is checking the wrong NPM package.

## Steps to Reproduce
```
~$ volta install npm

~$ npm -v
8.5.5

~$ ncu -g
npm  8.3.1  →  8.5.5

~$ npm -g install npm@8.5.5

~$ npm -v
8.5.5

~$ ncu -g
npm  8.3.1  →  8.5.5
```

## Current Behavior

`ncu -g` checks an incorrect version of the NPM package. I guess it's system installed version of NPM.

## Expected Behavior

`ncu -g` checks the correct NPM package.

srmagura:
Closes #1072.

I discovered that Volta has its own fancy system for installing global packages, and that this results in `npm ls -g` returning inaccurate / incomplete results. In particular, `npm ls -g` will typically show a different version of npm than what has been installed by Volta. 

I don't think npm-check-updates should be expected to work in cases where even npm itself does not know what packages are installed. So I've added a check that aborts the program if Volta is installed. Volta has its own CLI for managing globally-installed tools, so Volta users can just use that rather than npm-check-updates.

npm-check-updates should still work fine even if Volta is used as long as the `--global` option is not passed.

Tested on Ubuntu and Windows.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,12 +15,36 @@ import initOptions from './lib/initOptions'
 import programError from './lib/programError'
 import runGlobal from './lib/runGlobal'
 import runLocal from './lib/runLocal'
-import { Index, PackageFile, RunOptions, VersionSpec } from './types'
+import { Index, Options, PackageFile, RunOptions, VersionSpec } from './types'
 
 // exit with non-zero error code when there is an unhandled promise rejection
 process.on('unhandledRejection', err => {
   throw err
 })
+
+/**
+ * Volta is a tool for managing JavaScript tooling like Node and npm. Volta has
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Thanks for investigating!

PR #1119, opened 2022-05-22
Set up Prettier

srmagura:
Motivations for this change:

- Prettier adds a lot of value to a project. It makes code formatting consistent and saves you time because you no longer have to manually format stuff.
- Many contributors to the project, myself included, will have the Prettier VS Code extension set to format all files on save. It's very annoying when you make one change to file, save, and end up with a ton of unrelated formatting changes from Prettier.

I looked at your existing code and saw that you use single quotes and do not use semicolons, so I set the corresponding options in `prettier.config.cjs`.

I've cleanly separated the commits for this PR: commit 1 and 3 are manual changes, while commit 2 is solely formatting changes from running Prettier on all files: `npm run prettier -- --write`.

BTW, if you like, I can also set up a `precommit` hook to run Prettier on all staged files. This is useful if you want to enforce that all files are prettified, even if some contributors are not using the Prettier extension in their editor.


raineorshine (changes requested):
> * Prettier adds a lot of value to a project. It makes code formatting consistent and saves you time because you no longer have to manually format stuff.

Yes, I like prettier! I hadn't gotten around to adding it to this project but it's definitely needed.

Can you use this prettier config? https://github.com/cybersemics/em/blob/dev/.prettierrc.json. It contains my preferred settings.

> BTW, if you like, I can also set up a `precommit` hook to run Prettier on all staged files. This is useful if you want to enforce that all files are prettified, even if some contributors are not using the Prettier extension in their editor.

Let's add it to [pre-push](https://github.com/raineorshine/npm-check-updates/blob/main/.husky/pre-push). I like instantaneous commits.

srmagura:
Should be good now!

I also made it so Husky automatically installs the hooks when you run `npm install` in the repository — unless there was a reason you had left this out?

raineorshine:
As long as it doesn't install husky for those who install `npm-check-updates` via npm. Are you sure `npm install` is not called as part of the user install process?

srmagura:
Yes, I am sure `prepare` only runs on local installs (and I understand it would be bad if we got this wrong).

Here is the official npm documentation on `prepare`:

image

And here is the link to where the Husky docs recommend running `husky install` in the `prepare` script: https://typicode.github.io/husky/#/?id=install

I published this branch as `@srmagura/npm-check-updates` and installed it with both npm and Yarn 3. The install was successful, which indicates `prepare` did not run (if it did run, it would error since `husky` is a devDependency and would therefore be unavailable).

raineorshine:
Great, let's do it. 

I had a vague memory of the [old autoinstall](https://blog.typicode.com/husky-git-hooks-autoinstall/) which caused problems. I'm glad it's no longer an issue.

raineorshine:
CodeQL seems to be failing. Can you investigate?

srmagura:
Yes, I looked into that earlier. This is the line causing the problem: https://github.com/srmagura/npm-check-updates/blob/a745705807dcca417648fc3e49693c90878d008c/src/version-util.ts#L24

It seems like CodeQL thinks `WILDCARDS_PURE` comes from user input, making this a "regex injection" vulenarability. But in reality, that variable is hardcoded, so there is no risk. I think this warning likely existed before, but CodeQL is treating it as a new warning now due to reformatting.

Is there an `/* eslint-disable */` equivalent for CodeQL? I am not at all familiar with it.

raineorshine:
I don't know of one. You could try hard coding `REGEX_PURE` into the Regexp if nothing else works.

srmagura:
Hey Raine, I did some research on this. CodeQL is flagging this now because Prettier reformatted the code to be on a single line. The warning may have been dismissed in the past but is now considered a new warning because the line numbers changed.

image

Users with write access to a repository can [dismiss CodeQL alerts](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/triaging-code-scanning-alerts-in-pull-requests#dismissing-an-alert-on-your-pull-request). So the easiest way to resolve this is to follow [this link](https://github.com/raineorshine/npm-check-updates/pull/1119/files#diff-7bc1dc9c536bf3ca81c34b40f4cab4bb53fbaad16c0fed4eeec4a7d4027dfaad) to view the alert in the PR diff, then select _Dismiss > False positive_.

BTW, I also updated `codeql.yml` to use the CodeQL v2 actions since v1 is being deprecated.

raineorshine:
Great! Thanks for investigating.

Issue #1125, opened 2022-05-31
Multiple --reject/-x only consider the last one


jwickers:
- [X ] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [X ] I am using the latest version of `npm-check-updates`
- [ X] I am using `node >= 14`

---

## Steps to Reproduce
Add multiple reject patterns like `-x foo -x bar`

## Current Behavior
Multiple --reject/-x only consider the last one, so 
`ncu  -x foo -x bar` only rejects `bar` and ignores `-x foo`

## Expected Behavior
All reject patterns should be considered so
`ncu  -x foo -x bar`
and
`ncu  -x foo,bar`
should be equivalent.

raineorshine:
Fixed and published in `v13.0.4`. Thanks!

chemicalkosek:
Not sure if I should open a new issue, but on latest version 16.7.5 I see the same behavior.

`ncu -x axios -x typeorm`

```bash
[====================] 39/39 100%

 axios                              ^0.27.2  →    ^1.3.3
 ```

raineorshine:
@chemicalkosek Thanks for reporting! That is definitely a regression. I've published a fix to `v16.7.6` and added tests to ensure this doesn't happen again.

PR #1126, opened 2022-05-31
Allow multiple -x/--reject options to all be considered

jwickers:
Signed-off-by: Jeremy Wickersheimer <jwickers@gmail.com>

For #1125

### DIFF
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -10,7 +10,7 @@ export interface CLIOption<T = any> {
   deprecated?: boolean
   description: string
   help?: string
-  parse?: (s: string) => T
+  parse?: (s: string, p?: any) => T
   long: string
   short?: string
   type?: string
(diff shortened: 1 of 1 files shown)


src/cli-options.ts
```diff
@@ -10,7 +10,7 @@ export interface CLIOption<T = any> {
   deprecated?: boolean
   description: string
   help?: string
-  parse?: (s: string) => T
+  parse?: (s: string, p?: any) => T
```
raineorshine, line 13: I think this should be `p?: string[]` since it is parsing strings from the command line.

raineorshine (changes requested):
Thanks for the improvement!

src/cli-options.ts
jwickers: Yes, I originally just took the type from commander.js `program.options` that expects a `fn: (value: string, previous: any) => any` but `any` should actually be `T` looking at the Typescript definition for it.
Made the change.

raineorshine (approved):

PR #1133, opened 2022-06-10
Handle strict-ssl empty string to boolean conversion

### LINKED ISSUE #1132: When running ncu inside a package from npm-scripts .yarnrc / .npmrc, false boolean values are treated as ""
- [x] I am using `node >= 14`

---

## Steps to Reproduce
1. `yarn install npm-check-updates -D` to install package locally
2. Create .yarnrc with the following options:
2.1. `"strict-ssl" false`
2.2. registry "path-to-registry-with-self-signed-cert" (typical scenario for companies behind corporate proxy)
3. Add npm-script `"updates": "ncu"` to package.json
4. Call `yarn updates` 

## Current Behavior
Unable to verify first certificate in chain

## Expected Behavior
Working properly

Instead of steps 2.2, it is enough to just read npmConfig contents within src/package-managers/npm.ts.
It says "strict-ssl" ""

Wrapping false in quotes inside .yarnrc:
`"strict-ssl" "false"`
Fixes the bug, but that is incorrect syntax and every other tool stops working.

When calling globally - everything works fine. Version matches, tried rolling back to ^12.x.x - the result is the same.

Could try putting out a PR for this

IMalyugin:
Fixes #1132

### DIFF
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -43,7 +43,7 @@ const readNpmConfig = () => {
   const booleanKeys = { strictssl: true }
 
   /** Parses a string to a boolean. */
-  const stringToBoolean = (s: string) => s && s !== 'false' && s !== '0'
+  const stringToBoolean = (s: string) => !!s && s !== 'false' && s !== '0'
 
   // needed until pacote supports full npm config compatibility
   // See: https://github.com/zkat/pacote/issues/156


IMalyugin:
Got an extra commit hanging in PR, will try to get it out

raineorshine (approved):

raineorshine:
Code looks good, but you have an invalid lock file.

IMalyugin:
Oh, sorry I didn't notice it was npm, I built it with yarn and wondered why yarn.lock is ignored.

PR #1134, opened 2022-06-10
Add target=@[dist-tag] option support

### LINKED ISSUE #381: ncu with --dist-tag parameter
Imagine that you have installed a scoped dependency via dev tag as follows;

npm install @fdp_local_safak_test/component1@dev

and component1 has following dist-tags:
dev:1.1.0
latest:2.0.0 in the artifactory.

when you run, ncu, it will return to you 2.0.0. However it would be super helpful if it would get --tag parameter to specify which dist-tag to check. By this way, it would return the 1.1.0.

Any idea?
---------------------------

IMalyugin:
A working draft.

Small TODO:
- [x] Manual case testing
- [x] Search for possible edge cases
- [x] Tests: Update existing
- [x] Tests: Cover dist-tag

Resolves #381


src/lib/queryVersions.ts
```diff
@@ -37,79 +37,68 @@ async function queryVersions(packageMap: Index<VersionSpec>, options: Options =
...
+      return Promise.reject(
+        new Error(
+          `Unsupported target "${targetResult}" for ${packageManagerName}. Supported version targets are: ` +
+            packageManagerSupportedVersionTargets.join(', ') +
+            (!isGithubDependency ? ' and custom distribution tags, following "@" (example: @next)' : ''),
+        ),
+      )
+    }
+
+    try {
```
IMalyugin, line 68: So much copy-paste :)

Couldn't take it, had to "insert" some features into the 'git url' condition branch. They were exclusive to npm before unification.

Amongst them:
- retry
- try/catch

src/lib/queryVersions.ts
```diff
@@ -37,79 +37,68 @@ async function queryVersions(packageMap: Index<VersionSpec>, options: Options =
...
-                chalk.red(
-                  'FetchError: Request Timeout. npm-registry-fetch defaults to 30000 (30 seconds). Try setting the --timeout option (in milliseconds) to override this.',
-                ) +
-                '\n',
-            )
-          }
-
-          throw err
+      versionNew = !isGithubDependency && npmAlias && versionNew ? createNpmAlias(name, versionNew) : versionNew
+    } catch (err: any) {
```
IMalyugin, line 78: That whole part now applies to github too. Didn't change texts, but they are more-or-less general

src/package-managers/yarn.ts
```diff
@@ -220,10 +220,29 @@ export const list = async (options: Options = {}, spawnOptions?: SpawnOptions) =
  * @param options
  * @returns
  */
-export const latest: GetVersion = async (packageName: string, currentVersion: Version, options: Options = {}) => {
-  const latest = (await viewOne(
+export const greatest: GetVersion = async (packageName, currentVersion, options = {}) => {
```
IMalyugin, line 223: Just moved greatest up, to comply with eslint use-before-define

README.md
```diff
@@ -121,6 +121,9 @@ ncu "/^(?!react-).*$/" # windows
   - `0.1.0` → `0.2.1`
 - With `--target patch`, only update patch:
   - `0.1.0` → `0.1.2`
+- With `--target @next`, latest version with dist-tag=next is used
+  _NOTE: pre: true is required if you want to upgrade from a stable version to a dist-tag_
```
raineorshine, line 125: Is this desired? It seems like the default behavior for non-latest dist-tags should assume `pre: true`, like the other non-latest targets. Tags are often used for prelease versions or alternative branches.

IMalyugin: Didn't give that too much thought. But you are right that "default" behavior for non-latest assumes pre: true. If there is some case where that is not true, one can always go pre: false.

Good find!

src/types.ts
```diff
@@ -37,9 +37,11 @@ export interface VersionResult {
 export type FilterFunction = (packageName: string, versionRange: SemVer[]) => boolean
 export type FilterRejectPattern = string | string[] | RegExp | RegExp[] | FilterFunction
 
+// Can't really type a string that starts with @
```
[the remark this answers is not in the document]
IMalyugin: Ouch, another day - another thing learnt. I even surfed the web before typing that. The only thing I could find - was how to make a regex`ish guard. Guess that was added in latter versions of typescript. Nice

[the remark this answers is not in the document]
raineorshine: It's a bit irrelevant here since `` `@${string}` | string`` always equals `string`, but it's a cool trick anyway.

IMalyugin:
@raineorshine how do we handle the mock data needed for testing? The currently used mocks do not have any dist-tags. There is ncu-mock-pre: "2.0.0-alpha.0", but it is published under dist-tag="latest".

Probably need at least 3 versions for proper testing, such as:
"ncu-mock-pre@beta": "2.0.0-beta.0"
"ncu-mock-pre@next": "2.0.0-next.0"
"ncu-mock-pre@next": "2.0.0-next.1"

Doesn't seem right to create and publish a new mock-package under my namespace for this.

As an alternative, we could use some community owned packages:
https://www.npmjs.com/package/ryan-npm-dist-tag-test (by ryan)
https://www.npmjs.com/package/glasser-test-dist-tag-a (by apollo)
https://www.npmjs.com/package/glasser-test-dist-tag-b (by apollo)

But that doesn't seem like the brightest idea :)

src/lib/queryVersions.ts
```diff
@@ -70,7 +70,7 @@ async function queryVersions(packageMap: Index<VersionSpec>, options: Options =
         ...options,
         // upgrade prereleases to newer prereleases by default
         distTag,
-        pre: options.pre != null ? options.pre : isPre(version),
+        pre: options.pre != null ? options.pre : distTag !== 'latest' || isPre(version),
```
IMalyugin, line 73: Considered adding the check into isPre, as the second argument, but seeing how isPre is used that was a bad idea.
In all the other usage locations, the checks are made against version to determine if it is has a prefix. And there is no distTag in that context

IMalyugin:
Got another issue, if current version is newer than the target version with given dist-tag, the update doesn't happen.
Probably going to do the easy-way-resolve - dist-tag is always in priority.

raineorshine:
> @raineorshine how do we handle the mock data needed for testing?

I have created and published https://github.com/raineorshine/ncu-test-dist-tag/tree/main. Will that work for you?

You mentioned needing three different packages for testing, but I didn't quite understand the reason. Could you explain?

raineorshine:
> Got another issue, if current version is newer than the target version with given dist-tag, the update doesn't happen.

Yes, this is expected. ncu should never suggest upgrades that are lower than the specified version, only higher.

> Probably going to do the easy-way-resolve - dist-tag is always in priority.

I'm not sure this is a good idea, as it differs from an important invariant in ncu's behavior. 

If a dependency is on a specific tag, it is unlikely to exceed that tag. If a dependency is on `latest` and the user decides to point it to a tag with a lower version than `latest`, I think it is on them to fix the version before running `npm-check-updates`.

> That could become an issue in cases such as this:
> 7 packages were in 1.0.0-next.0@next
> 7 packages were upgraded to 1.0.0@latest
> 3 packages were upgraded to 1.0.1-next.0@next

To clarify, do you mean "upgraded" by the package authors, by the package user, by `npm update/install`, or by `ncu`?

IMalyugin:
> > @raineorshine how do we handle the mock data needed for testing?
> 
> I have created and published https://github.com/raineorshine/ncu-test-dist-tag/tree/main. Will that work for you?
> 
> You mentioned needing three different packages for testing, but I didn't quite understand the reason. Could you explain?

Not packages, needed 3 different versions, as an alternative, i suggested using open source ones. But it's much better having the packages native to ncu. Thanks!

> > Got another issue, if current version is newer than the target version with given dist-tag, the update doesn't happen.
> 
> Yes, this is expected. ncu should never suggest upgrades that are lower than the specified version, only higher.

Going into different branches could be a show stopper, but current algorithm allows to go from latest to dist with newer version and vice versa.
> 
> > Probably going to do the easy-way-resolve - dist-tag is always in priority.
> 
> I'm not sure this is a good idea, as it differs from an important invariant in ncu's behavior.
> 
> If a dependency is on a specific tag, it is unlikely to exceed that tag. If a dependency is on `latest` and the user decides to point it to a tag with a lower version than `latest`, I think it is on them to fix the version before running `npm-check-updates`.

That is more or less how it works in pull request. I bumped into the [readme](https://github.com/npm/node-semver).
image

Different prerelease tags with same major/minor/patch can NOT be compared by semver. But when there are different prerelease tags such as 1.0.0-task101.0 vs 1.0.0-next.5, semver default comparison ends up putting a > sign before them, because "task" can be string-compared to "next" with ">" sign. That behavior should actually be banned as well. So what last commit in pull request does, is allow switching from given versions.

> > That could become an issue in cases such as this:
> > 7 packages were in 1.0.0-next.0@next
> > 7 packages were upgraded to 1.0.0@latest
> > 3 packages were upgraded to 1.0.1-next.0@next
> 
> To clarify, do you mean "upgraded" by the package authors, by the package user, by `npm update/install`, or by `ncu`?

I meant as a user, if i have 7 remote packages used in my project, ncu would work like that. In comparison, If i ran npm update/install given distTag, it would install versions with this distTag without looking at versions. So that's how PR works right now.

IMalyugin:
To save some of your time, heres the set of publish commands:
```
---0.2---
0.1.0 -> npm version preminor && npm publish --tag=next -> 0.2.0-0
0.1.0 -> npm version preminor --pre=task-42 && npm publish --tag=task-42-> 0.2.0-0-task-42.0

---1.0.0---
0.1.0 -> npm version premajor --pre=task-42 && npm publish --tag=task-42 -> 1.0.0-task-42.0
0.1.0 -> npm version premajor --pre=beta && npm publish --tag=beta -> 1.0.0-beta.0
0.1.0 -> npm version major && npm publish -> 1.0.0

---1.1.0---
1.0.0 -> npm version prepatch --pre=beta && npm publish --tag=beta -> 1.1.0.beta.0
1.0.0 -> npm version minor && npm publish -> 1.1.0

**Comment by IMalyugin** [comment]
Also need 1.1.1-dev.0 to cover downgrading from dev to latest 1.1.0. By current algorithm it is going to fail. Guess that is right thing to do, unless forced by another flag

**Comment by raineorshine** [comment]
> To save some of your time, heres the set of publish commands

Thanks! I created a new repo and did the versions and publishing as described. Let me know if anything is off: https://github.com/raineorshine/ncu-test-tag

**Comment by raineorshine** [comment]
> Going into different branches could be a show stopper, but current algorithm allows to go from latest to dist with newer version and vice versa.

I'm not sure what you mean by going into different branches. `npm-check-updates` is not branch aware. On any given branch, it will use the versions defined in the `package.json` in that branch. I'm guessing you have a specific workflow in mind. Feel free to enumerate the steps for me if it's important.

> Different prerelease tags with mismatching major/minor/patch can NOT be compared by semver. But when there are different prerelease tags on the same major/minor/patch such as 1.0.0-task101.0 vs 1.0.0-next.5, semver default comparison ends up putting a > sign before them, because "task" can be string-compared to "next" with ">" sign. That behavior should actually be banned as well. String comparing only works for alpha, beta, next, because n>b>a. So what last commit in pull request does, is allow switching from given versions.

Yes, agreed. ncu should suggest `1.0.0-next.5` as an upgrade for `1.0.0-next.4`, but not `1.0.0-task101.0`.

> To sum up, given already published:
> 
> ```
> latest: 0.1.0
> next: 1.0.0-0, 1.0.0-1
> ```

The commands you gave earlier had `0.2.0-0` published to `next`, but here you are showing `next: 1.0.0-0, 1.0.0-1`. Does this indicate something else?

> For proper testing we need:
> 
> ```
> latest: 1.0.0, 1.1.0 (one latest between low/hig dists, one to go over dist-tags)
> ...
> ```

When you say "one latest between low/hig dists, one to go over dist-tags", you just mean chronologically published, right? Because there is no history of versions in the npm registry that were published to a specific tag. There are just the currently published tags and a list of all versions with their publish times.

> This way we can settle on the "which version should update to which version" within tests. For example:

The table you provides is great, though I am still figuring out how to parse it. Maybe once the above questions are clarified it will become clearer. What I am looking for in each test case (or pseudo code test case for our discussion here) is the following:

1. current dependencies
2. published dependencies
3. ncu options
4. expected upgrades

**Comment by IMalyugin** [comment]
> > Going into different branches could be a show stopper, but current algorithm allows to go from latest to dist with newer version and vice versa.
> 
> I'm not sure what you mean by going into different branches. `npm-check-updates` is not branch aware. On any given branch, it will use the versions defined in the `package.json` in that branch. I'm guessing you have a specific workflow in mind. Feel free to enumerate the steps for me if it's important.

I should be more precise with words. By "branches" I actually meant different dist-tags, like going from @next to @dev . Nobody can tell wether next is bigger than dev or the opposite.

> > Different prerelease tags with mismatching major/minor/patch can NOT be compared by semver. But when there are different prerelease tags on the same major/minor/patch such as 1.0.0-task101.0 vs 1.0.0-next.5, semver default comparison ends up putting a > sign before them, because "task" can be string-compared to "next" with ">" sign. That behavior should actually be banned as well. String comparing only works for alpha, beta, next, because n>b>a. So what last commit in pull request does, is allow switching from given versions.
> 
> Yes, agreed. ncu should suggest `1.0.0-next.5` as an upgrade for `1.0.0-next.4`, but not `1.0.0-task101.0`.
That depends, it should suggest upgrading:
from 1.0.0-next.5 to 1.0.0-task101.0  on target=@task101
it should also suggest upgrading:
from 1.0.0-task101.0 to from 1.0.0-next.5  on target=@next

Because we can't compare two prerelease branches just by their prerelease increment. Comparing them as strings is also wrong.

> > To sum up, given already published:
> > ```
> > latest: 0.1.0
> > next: 1.0.0-0, 1.0.0-1
> > ```
> 
> The commands you gave earlier had `0.2.0-0` published to `next`, but here you are showing `next: 1.0.0-0, 1.0.0-1`. Does this indicate something else?

I think I missed a few commands, it was 1 am when i wrote those, specifically these:
```
0.1.0 -> npm version premajor && npm publish --tag=next -> 1.0.0-0
1.0.0-0 -> npm version prerelease && npm publish --tag=next -> 1.0.0-1
```

Also in the following comment i said one more dist-tag needed for a special case:
`1.1.0 -> npm version preminor --preid=dev && npm publish --tag=dev -> 1.1.1-dev.0`

> > For proper testing we need:
> > ```
> > latest: 1.0.0, 1.1.0 (one latest between low/hig dists, one to go over dist-tags)
> > ...
> > ```
> 
> When you say "one latest between low/hig dists, one to go over dist-tags", you just mean chronologically published, right? Because there is no history of versions in the npm registry for a specific tag. There are just the currently published tags and a list of all versions with their publish times.
More like semver sorting order, out of versionordering there is one certainthing - latest versions, two latest versions with different major/minor/patch -  will always be sorted in the same order. As for the prerelease versions (with dist-tags - it's not as simple)

> > This way we can settle on the "which version should update to which version" within tests. For example:
> 
> The table you provides is great, though I am still figuring out how to parse it. Maybe once the above questions are clarified it will become clearer. What I am looking for in each test case (or pseudo code test case for our discussion here) is the following:
> 
> 1. current dependencies
> 2. published dependencies
> 3. ncu options
> 4. expected upgrades

Well that's what the table is for
1. that's the source column
2. that's all the versions listed
3. second column is --target=[target]
4. last column, result - is an expected upgrade 

Going to write the tests as one-liners with the same parameters

**Comment by IMalyugin** [comment]
Okay got in a bunch of tests, still need the following, though:
```
0.1.0 -> npm version premajor && npm publish --tag=next -> 1.0.0-0
1.0.0-0 -> npm version prerelease && npm publish --tag=next -> 1.0.0-1
1.1.0 -> npm version preminor --preid=dev && npm publish --tag=dev -> 1.1.1-dev.0
```

File: test/index.test.ts [hunk 7]
```diff
@@ -416,6 +417,62 @@ describe('run', function () {
...
+      await testNcuTargetUpgrade('@beta', '1.0.0-task-42.0', '1.0.1-beta.0')
+    })
+
+    it('upgrade to same semver between dist-tags', async () => {
+      // can't detect which prerelease is higher, so just allow switching
+      await testNcuTargetUpgrade('@task-42', '1.0.0-beta.0', '1.0.0-task-42.0')
+
+      // same here, major/minor/patch is the same, can't determine which one is bigger task-42 or next
+      // await testNcuTargetUpgrade('@next', '1.0.0-task-42.0', '1.0.0-1')
+      await testNcuTargetUpgrade('@next', '1.0.0-task-42.0', '0.2.0-0')
```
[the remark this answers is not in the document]
IMalyugin: Easiest way to explain why is this:
1. Latest is at 0.1.0
2. First developer does npm version premajor --preid=task-42 && npm publish --tag=task-42
3. Second developer goes npm version preminor && npm publish --tag=next
4. Since both branches are made out of 0.1.0, can we really say that 1.0.0-task-42 is greater than 0.2.0-0 ? They are both prereleases, so there was no release
There is no way to "fix" this for second-dev Because merging task-42 into next is something that should not be done as task-42 could be in a separate testing stage. So I suggest we use the intention, if somebody wants to go from one prerelease branch to another - let them.
Normally all the upgrading should be done from latest -> dist-tag -> latest, or within same dist-tag.

test/index.test.ts
```diff
@@ -416,6 +417,62 @@ describe('run', function () {
     })
   }) // end 'target'
 
+  describe('distTag as target', () => {
+    /** Compact function to test ncu */
+    const testNcuTargetUpgrade = async (target: Target, currentVersion: string, upgradeVersion: string) => {
```
raineorshine, line 422: My preference for unit tests is to use low-level, literal statements when possible and avoid creating abstractions. This makes each test self-contained and saves developers from having to learn a special API to read the tests.  This contrasts with my priorities in production code, where I tend to value abstraction and the elimination of redundancies. I only use abstractions in test code when the low-level code is so convoluted as to make it hard to read. 

Please inline the test code and remove the `testNcuTargetUpgrade` function.

raineorshine:
> Okay got in a bunch of tests: [6a9d571](https://github.com/raineorshine/npm-check-updates/commit/6a9d571066f7838b218f6f585e6aae1f44a15311), still need the following, though:
> 
> ```
> 0.1.0 -> npm version premajor && npm publish --tag=next -> 1.0.0-0
> 1.0.0-0 -> npm version prerelease && npm publish --tag=next -> 1.0.0-1
> 1.1.0 -> npm version preminor --preid=dev && npm publish --tag=dev -> 1.1.1-dev.0
> ```

Done

```sh
$ npm view ncu-test-tag dist-tags
{
  latest: '1.1.0',
  next: '1.0.0-1',
  'task-42': '1.0.0-task-42.0',
  beta: '1.0.1-beta.0',
  dev: '1.2.0-dev.0'
}
```

test/index.test.ts
```diff
@@ -416,6 +417,62 @@ describe('run', function () {
...
+      // same here, major/minor/patch is the same, can't determine which one is bigger task-42 or next
+      // await testNcuTargetUpgrade('@next', '1.0.0-task-42.0', '1.0.0-1')
+      await testNcuTargetUpgrade('@next', '1.0.0-task-42.0', '0.2.0-0')
+    })
+
+    it('upgrade to lower semver between dist-tags', async () => {
+      // comparing semver between different dist-tags is incorrect, both versions could be released from the same latest
+      // so instead of looking at numbers, we should focus on intention of the user upgrading to specific dist-tag
+      await testNcuTargetUpgrade('@task-42', '1.0.1-beta.0', '1.0.0-task-42.0')
+    })
```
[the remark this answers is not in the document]
raineorshine: Cool, got it. Makes sense.

raineorshine (changes requested):

raineorshine:
The new unit tests cleared up a lot of my previous questions. Sometimes tests are the best documentation :).

IMalyugin:
Some test descriptions are now off, not matching the contents.

Got it about abstractions in tests, though one might disagree that copying 10 lines function provided in the same scope 20 times makes it clearer to read. It surely creates more noise, that takes focus away.

it’s night time for me, so I’ll clean it up tomorrow.

As for the downgrading aspect between prerelease versions, there is a more complicated, yet better fitting solution. New option —downgrade-tolerance could solve all the disagreement of different parties. With values such as: minor, patch, major, build, none. That option would also bring benefit to ncu core mechanics, allowing to go: —target=latest —downgrade-tolerance=minor (downgrading prerelease versions to latest stable, if it is only a minor away). But I’d say not in this PR.

raineorshine:
> Some test descriptions are now off, not matching the contents.

Please point out which ones are off and why, and I can revise.

> Got it about abstractions in tests, though one might disagree that copying 10 lines function provided in the same scope 20 times makes it clearer to read. It surely creates more noise, that takes focus away, but you’re the boss here :)

Fair enough. The abstracted version is certainly more terse. For this project though, I prefer inline.

> As for the downgrading aspect between prerelease versions, there is a more complicated, yet better fitting solution. New option —downgrade-tolerance could solve all the disagreement of different parties. With values such as: minor, patch, major, build, none. That option would also bring benefit to ncu core mechanics, allowing to go: —target=latest —downgrade-tolerance=minor (downgrading prerelease versions to latest stable, if it is only a minor away). But I’d say not in this PR.

That's a great idea. Yes, I think a separate PR is appropriate.

test/index.test.ts
IMalyugin: done

raineorshine (approved):
Looks great! Ready to merge?

IMalyugin:
> Looks great! Ready to merge?

Ready when you are 👍

PR #1143, opened 2022-06-22
Static registry

agrouse:
For some reason, running the lint script errored out originally, telling me to add 'eslint-plugin-n.' Let me know however, if this dependency needs to be removed.


src/cli-options.ts
```diff
@@ -301,7 +301,7 @@ As a comparison: without using the --peer option, ncu will suggest the latest ve
   {
     long: 'registry',
     short: 'r',
-    arg: 'url',
+    arg: 'url' || 'path',
```
raineorshine, line 304: `arg` must be a single argument name, to be used by commander. (As written this will always resolve to `url`, since the string literal `'url'` is always truthy.)

I would suggest we change this to `uri`. Even though we don't use an explicit [file URI scheme](https://en.wikipedia.org/wiki/File_URI_scheme), both urls and files fit conceptually into a "URI".

src/package-managers/staticRegistry.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @param packageName
+ * @param currentVersion
+ * @param options
+ * @returns A promise that fulfills to string value or null
+ */
+export const latest: GetVersion = (packageName: string, currentVersion: Version, options: Options = {}) => {
+  if (Object.keys(registry).length === 0) {
+    initializeRegistry(options.registry)
+  }
```
raineorshine, line 33: Caching the registry would break the `--deep` option for monorepos that have different static registries per repo. Perhaps unlikely, but caching bugs are a nightmare to troubleshoot and this is easy to avoid.

I would suggest one of two approaches:

1. Create a separate cache for each value of `options.registry`. i.e. change `registry` to type `{ [key: string]: StaticRegistry }`. This will effectively memoize the `registry` by location.

2. Use `fast-memoize` which is already used by the `npm` adapter to optimize repeat registry access. This effectively hides the implementation details of the cache for the same effect:

    https://github.com/raineorshine/npm-check-updates/blob/d7df1515391ce2084fd73bd4d8b55b09fa57ae1c/src/package-managers/npm.ts#L215-L216

test/package-managers/staticRegistry/index.test.ts
```diff
@@ -0,0 +1,14 @@
...
+import * as staticRegistry from '../../../src/package-managers/staticRegistry'
+
+chai.should()
+chai.use(chaiAsPromised)
+
+describe('staticRegistry', function () {
+  it('latest', async () => {
+    const registry = './test/package-managers/staticRegistry/staticRegistry.json'
+    const version = await staticRegistry.latest('express', '', { cwd: __dirname, registry })
+    parseInt(version!, 10).should.be.above(1)
```
raineorshine, line 12: Let's do an exact version comparison rather than `should.be.above`. That we we guarantee that it is using the static registry and not the `npm` registry by accident.

src/package-managers/staticRegistry.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Assigns static registry dependencies to variable given a valid path
+ *
+ * @param path
+ * @returns
+ */
+const initializeRegistry = (path: string | undefined): void => {
+  if (path === undefined || !fs.existsSync(path)) {
+    throw new Error('No or invalid path to static registry was provided. Please try again')
+  }
```
raineorshine, line 18: I'd like to separate these two checks in order to provide more useful feedback to the user.

Please use [programError](https://github.com/raineorshine/npm-check-updates/blob/d7df1515391ce2084fd73bd4d8b55b09fa57ae1c/src/lib/programError.ts) for errors with known causes rather rather than throwing an uncaught exception. This will make it clear that it is a user error, not a bug in npm-check-updates.

1. The missing `options.registry` check should go in [initOptions](https://github.com/raineorshine/npm-check-updates/blob/main/src/lib/initOptions.ts) where other options validation lives. Rather than `Please try again`, say `When --package-manager staticRegistry is specified, you must provide the path for the registry file with --registry. Run "ncu --help --packageManager" for details.`

2. Change "invalid path" to say "The specified static registry file does not exist: [FILE]".

raineorshine (changes requested):
Thanks! Overall looks great.

src/package-managers/staticRegistry.ts
agrouse: Thanks for the tip--I think I'll go with the second option!

raineorshine (approved):
Thanks so much!

Issue #1145, opened 2022-06-23
Update `update-notifier` to v6


MrArnoldPalmer:
- [X] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [X] I am using the latest version of `npm-check-updates`
- [X] I am using `node >= 14`

`update-notifier` recently released v6 which upgrades it's dependency on `got` (deep down the tree) which currently has a [CVE related](https://github.com/advisories/GHSA-pfrx-2q88-qq97).

v6 of `update-notifier` also migrates to pure ESM which would require consumers to also migrate. So upgrading requires NCU to use ESM and drop support for node 12 (which was already done), but also programmatic NCU consumers would have to migrate to ESM.

Not sure how interested you are in migrating NCU to pure ESM and/or if this CVE is a forcing function for that (NCU may not even be vulnerable I haven't found details to support one way or another) but I started making a PR to do the version upgrade but with the ESM changes it started growing quite large so I figured I should check first before dumping it here 😆 

https://github.com/raineorshine/npm-check-updates/compare/main...MrArnoldPalmer:chore/upgrade-notifier current state though `build:options` currently failing and I haven't dug in.

raineorshine:
Thanks for the information and initial investigation.

I've followed the dependencies down and npm-check-updates does not appear to be vulnerable. 

```sh
$ npm ls got
npm-check-updates@14.1.0 /Users/raine/projects/npm-check-updates
└─┬ update-notifier@5.1.0
  └─┬ latest-version@5.1.0
    └─┬ package-json@6.5.0
      └── got@9.6.0
```

- `update-notifier` only uses `got` via [latest-version](https://github.com/sindresorhus/latest-version#readme) which uses `package-json`. 
- [package-json](https://github.com/sindresorhus/package-json/blob/fbbee76b615efdb489e72439130b4075b2ec793a/index.js#L43-L76) defaults to https://registry.npmjs.org/ via [registry-url](https://registry.npmjs.org/). 
- `update-notifier` [does not provide](https://github.com/yeoman/update-notifier/blob/3fdb21876aa391f9bc7dc35b7d81151677fb533d/index.js#L106) a custom `registry` option to `latest-version`: 
- Thus, the registry will always be https://registry.npmjs.org/ and no UNIX redirect is possible.

> programmatic NCU consumers would have to migrate to ESM.

[...]

PR #1147, opened 2022-06-25
Update `update-notifier` from ^5.1.0 to ^6.0.2

louislam:
One of my GitHub repo got some kind of warning because of `update-notifier`, updating to 6.0.2 should solve the issue.


raineorshine:
This breaks all the tests because `update-notifier` is a pure ESM module. 

The previous discussion is here: #1145.

Issue #1120, opened 2022-05-23
npm erroneously used in Yarn Workspaces monorepo


srmagura:
## Steps to Reproduce

Create a repository using Yarn 3 and workspaces:

```text
.yarnrc.yml
package.json    # Root package.json which defines workspaces
yarn.lock
package-a/
  package.json   
```

Now `cd` into `package-a` and run `ncu -u`. It will say "Run npm install to install new versions" at the end, indicating that npm-check-updates is using `src/package-managers/npm.ts`.

This does not cause any problems unless you have configuration like `npmScopes` in `.yarnrc.yml` which npm-check-updates needs to know about. 

Will PR a fix for this on the weekend.

PR #1148, opened 2022-06-25
Add `determinePackageManager`, fix Yarn detection when using workspaces

srmagura:
## Issues Addressed

- Resolves https://github.com/raineorshine/npm-check-updates/issues/1120.
- Resolves an issue I noticed during implementation: if the `packageFile` option was passed, the code still looked in `.` (current directory) when checking for a `yarn.lock`. Now it starts looking in the directory of the package file. I believe this is the correct behavior.

## Notes

- I put the new test file in the same directory as the file being tested. This is my preferred pattern ([rationale here](https://kentcdodds.com/blog/colocation)), but I know this differs from the organization used for the other test files in the repo. I'm happy to move the test file if you tell me where it should go.
- I added a `mocha` script to enable running a single test file, e.g. `npm run mocha -- src/lib/determinePackageManager.test.ts`
- If you'd like to test this change against a "real" repository that uses Yarn workspaces, you can clone https://github.com/srmagura/npm-check-updates-1120-test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "prepare": "husky install",
     "prepublishOnly": "npm run build",
     "prettier": "prettier .",
+    "mocha": "mocha",
     "test": "npm run test:src && npm run test:timeout",
     "test:src": "mocha test src/test test/package-managers/npm test/package-managers/yarn test/package-managers/staticRegistry",
     "test:timeout": "mocha --exit test/timeout",
--- a/src/lib/initOptions.ts
+++ b/src/lib/initOptions.ts
@@ -8,6 +8,7 @@ import { print } from '../logging'
 import { Options } from '../types/Options'
 import { RunOptions } from '../types/RunOptions'
 import { Target } from '../types/Target'
+import { determinePackageManager } from './determinePackageManager'
 
 /** Initializes, validates, sets defaults, and consolidates program options. */
 function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } = {}): Options {
@@ -115,11 +116,9 @@ function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } = {}): O
 
   const format = options.format || []
 
-  // autodetect yarn
-  const files = fs.readdirSync(options.cwd || '.')
-  const autoYarn =
-    !options.packageManager && !options.global && files.includes('yarn.lock') && !files.includes('package-lock.json')
-  if (autoYarn) {
+  const packageManager = determinePackageManager(options)
+
+  if (!options.packageManager && packageManager === 'yarn') {
     print(options, 'Using yarn')
   }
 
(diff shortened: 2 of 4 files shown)


srmagura:
Converting to draft because, I realized the logic for finding the `.npmrc`/`.yarnrc.yml` is probably also broken...

package.json
```diff
@@ -37,6 +37,7 @@
     "prepare": "husky install",
     "prepublishOnly": "npm run build",
     "prettier": "prettier .",
+    "mocha": "mocha",
```
raineorshine, line 40: You can remove this. Users can add `./node_modules/.bin` to their PATH, or they can run `npx mocha`, both of which are easier than `npm run mocha`. It would be helpful if there were cli arguments, but for the default execution we don't need an npm script.

src/lib/findLockAndConfigFiles.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const defaultPackageManager = 'npm'
+
+/**
+ * If the packageManager option was not provided, look at the lockfiles to
+ * determine which package manager is being used.
+ *
+ * @param readdirSync This is only a parameter so that it can be used in tests.
+ */
+export function determinePackageManager(
```
raineorshine, line 60: Could you split this file up into multiple files? That is a convention we follow.

The file name should be the same as the function name, and there should be a single default export.

src/lib/initOptions.ts
```diff
@@ -115,11 +116,9 @@ function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } = {}): O
...
   const format = options.format || []
 
-  // autodetect yarn
-  const files = fs.readdirSync(options.cwd || '.')
-  const autoYarn =
-    !options.packageManager && !options.global && files.includes('yarn.lock') && !files.includes('package-lock.json')
-  if (autoYarn) {
+  const packageManager = determinePackageManager(options)
+
+  if (!options.packageManager && packageManager === 'yarn') {
```
raineorshine, line 121: I think this can be just `if (packageManager === 'yarn')`?

raineorshine (changes requested):
Looks great!

package.json
srmagura: Nice, forgot you could just use `npx`.

srmagura:
Thanks for the feedback, should be good now :)

src/lib/initOptions.ts
raineorshine: Ah, I remember the original intention now. If the user specifies `options.packageManager` explicitly, there is no need to output the message. ncu aims for minimal design when possible.

srmagura:
Thanks for merging. I checked out your changes, it all looked good except, I find nested ternaries to be hard to read. I guess not everyone feels that way though. No need to change 👍

raineorshine:
Thanks for your contribution! I'll release it shortly.

> Thanks for merging. I checked out your changes, it all looked good except, I find nested ternaries to be hard to read. I guess not everyone feels that way though. No need to change 👍

👏👏👏 *preferences* 🙃🙃🙃

PR #1150, opened 2022-06-30
Bump version of update-notifier

### LINKED ISSUE #1145: Update `update-notifier` to v6

JustusFluegel:
Bump version of update-notifier. This comes with a few challenges as update-notifier past v5 is only available as an es module, therefore the following things are necessary:

- Use dynamic imports as node only supports dynamic imports for es modules inside cjs
- Bump engine version to v14.14 as only thereafter esm modules were introduced in node.
- Use nodenext module resulution which prevents polyfilling dynamic imports with require by typescript.

Closes #1145

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,7 @@
         "semver-utils": "^1.1.4",
         "source-map-support": "^0.5.21",
         "spawn-please": "^1.0.0",
-        "update-notifier": "^5.1.0",
+        "update-notifier": "^6.0.2",
         "yaml": "^2.1.1"
       },
       "bin": {
@@ -92,7 +92,7 @@
         "yarn": "^1.22.19"
       },
       "engines": {
-        "node": ">=14"
+        "node": ">=14.14"
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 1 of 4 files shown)


JustusFluegel:
well yea I am an idiot haven't updated package-lock.json sec

JustusFluegel:
should finally work now

raineorshine:
Thank g0d for ci 😁

raineorshine (approved):

PR #1154, opened 2022-07-02
Fixed custom target function helper text

TimESQuit:
Helper text incorrectly shows `major` as a number, when it is a string.

Previously custom target helper text:
`If (major === 0) return 'minor'`

Change:
`if (major === '0') return 'minor'`

Note: this behavior is already correctly tested for [here](https://github.com/raineorshine/npm-check-updates/blob/0b9dedccada7aff7d5fdce47894c9aeebab7392e/test/index.test.ts#L407).


raineorshine (approved):
Thanks!

PR #1156, opened 2022-07-05
Add group customization option

### LINKED ISSUE #1152: Feature Request: Configure list of non-semver packages
As you probably know, there's some major packages which don't follow semver. (TypeScript being the most notable.) With the new `--format group` option, it seems like it'd be very useful to be able to customize via the `.ncurc.js` file a list of packages which don't follow semver.

Proposals:

* Easy: A list of packages and any time any version change is done to those packages, it is considered breaking by default. Probably overly aggressive but simplest to configure and implement.
* Advanced: A map of package names to semver patterns. So for TypeScript it could be `"typescript": "X.X.Z"` Unsure of the format pattern, but some way to indicate the pattern that the particular library uses for Major (X), Minor (Y), Patch (Z) which then ncu can use to categorize it correctly. This seems like a much bigger task so I'm not advocating for it, but just throwing it out for completeness.

A side benefit to this is that this feature can be used to highlight important packages. I know some libraries I really want to know whenever they update regardless of semver. If we want this feature to be usable for this purpose, we can rebrand it to "flagPackages" (name pending) so it can be used for any reason that someone might want to highlight when a package has been updated (either because it doesn't use semver, or because they just want to not miss an update). This would require no extra work, just renaming.

If accepted, I'm happy to contribute this feature.

CreativeTechGuy:
Closes #1152

### DIFF
--- a/src/version-util.ts
+++ b/src/version-util.ts
@@ -7,6 +7,7 @@ import util from 'util'
 import { keyValueBy } from './lib/keyValueBy'
 import { Maybe } from './types/Maybe'
 import { VersionLevel } from './types/VersionLevel'
+import { UpgradeGroup } from './types/UpgradeGroup'
 
 const VERSION_BASE_PARTS = ['major', 'minor', 'patch'] as VersionPart[]
 const VERSION_ADDED_PARTS = ['release', 'build'] as VersionPart[]
@@ -32,8 +33,6 @@ interface UpgradeOptions {
   wildcard?: string
   removeRange?: boolean
 }
-
-export type UpgradeGroup = 'major' | 'minor' | 'patch' | 'pre-v1' | 'none'
 
 /**
  * @param version
--- a/src/version-util.ts
+++ b/src/version-util.ts
@@ -6,8 +6,8 @@ import semverutils, { SemVer } from 'semver-utils'
 import util from 'util'
 import { keyValueBy } from './lib/keyValueBy'
 import { Maybe } from './types/Maybe'
-import { VersionLevel } from './types/VersionLevel'
 import { UpgradeGroup } from './types/UpgradeGroup'
+import { VersionLevel } from './types/VersionLevel'
 
 const VERSION_BASE_PARTS = ['major', 'minor', 'patch'] as VersionPart[]
 const VERSION_ADDED_PARTS = ['release', 'build'] as VersionPart[]
--- a/src/version-util.ts
+++ b/src/version-util.ts
@@ -33,6 +33,8 @@ interface UpgradeOptions {
   removeRange?: boolean
 }
 
+export type UpgradeGroup = 'major' | 'minor' | 'patch' | 'pre-v1' | 'none'
+
 /**
  * @param version
  * @returns The number of parts in the version
@@ -158,7 +160,7 @@ export function isWildPart(versionPartValue: Maybe<string>) {
  * @param from
  * @param to
  */
-export function partChanged(from: string, to: string): 'major' | 'minor' | 'patch' | 'majorVersionZero' | 'none' {
+export function partChanged(from: string, to: string): UpgradeGroup {
   if (from === to) return 'none'
 
   // separate out leading ^ or ~
--- a/src/version-util.ts
+++ b/src/version-util.ts
@@ -197,7 +197,8 @@ export function getDependencyGroups(
     const from = oldDependencies[dep]
     const defaultGroup = partChanged(from, to)
     const userDefinedUpgradeGroup =
-      options.group?.(dep, defaultGroup, parseRange(from), parseRange(to), parse(newDependencies[dep])) ?? defaultGroup
+      options.groupFunction?.(dep, defaultGroup, parseRange(from), parseRange(to), parse(newDependencies[dep])) ??
+      defaultGroup
     if (userDefinedUpgradeGroup === 'none') {
       return accum
     }
--- a/src/version-util.ts
+++ b/src/version-util.ts
@@ -2,10 +2,13 @@ import chalk from 'chalk'
 import _ from 'lodash'
 import parseGithubUrl from 'parse-github-url'
 import semver from 'semver'
-import semverutils, { SemVer } from 'semver-utils'
+import semverutils, { SemVer, parse, parseRange } from 'semver-utils'
 import util from 'util'
 import { keyValueBy } from './lib/keyValueBy'
+import { getGroupHeadings } from './logging'
+import { Index } from './types/IndexType'
 import { Maybe } from './types/Maybe'
+import { Options } from './types/Options'
 import { UpgradeGroup } from './types/UpgradeGroup'
 import { VersionLevel } from './types/VersionLevel'
 
@@ -179,6 +182,48 @@ export function partChanged(from: string, to: string): UpgradeGroup {
   // minor = cyan
   // patch = green
   return partsTo[0] === '0' ? 'majorVersionZero' : i === 0 ? 'major' : i === 1 ? 'minor' : 'patch'
+}
+
+/**
+ * Returns a list of group heading and a map of package names and versions.
+ * Used with --format group and takes into account the custom --group function.
+ */
+export function getDependencyGroups(
+  newDependencies: Index<string>,
+  oldDependencies: Index<string>,
+  options: Options,
+): { heading: string; groupName: string; packages: Index<string> }[] {
+  const groups = keyValueBy<string, Index<string>>(newDependencies, (dep, to, accum) => {
+    const from = oldDependencies[dep]
+    const defaultGroup = partChanged(from, to)
+    const userDefinedUpgradeGroup =
+      options.group?.(dep, parseRange(from), parseRange(to), parse(newDependencies[dep])) ?? defaultGroup
+    if (userDefinedUpgradeGroup === 'none') {
+      return accum
+    }
+    return {
+      ...accum,
+      [userDefinedUpgradeGroup]: {
+        ...accum[userDefinedUpgradeGroup],
+        [dep]: to,
+      },
+    }
+  })
+
+  const headings = getGroupHeadings(options)
+  const groupOrder = _.uniq(['patch', 'minor', 'major', 'majorVersionZero', ..._.sortBy(Object.keys(groups))])
+
+  return groupOrder
+    .filter(groupName => {
+      return groupName in groups
+    })
+    .map(groupName => {
+      return {
+        groupName,
+        heading: groupName in headings ? headings[groupName as keyof typeof headings] : groupName,
+        packages: groups[groupName],
+      }
+    })
 }
 
 /**
--- a/src/version-util.ts
+++ b/src/version-util.ts
@@ -1,11 +1,10 @@
-import chalk from 'chalk'
+import Chalk from 'chalk'
 import _ from 'lodash'
 import parseGithubUrl from 'parse-github-url'
 import semver from 'semver'
 import semverutils, { SemVer, parse, parseRange } from 'semver-utils'
 import util from 'util'
 import { keyValueBy } from './lib/keyValueBy'
-import { getGroupHeadings } from './logging'
 import { Index } from './types/IndexType'
 import { Maybe } from './types/Maybe'
 import { Options } from './types/Options'
@@ -193,6 +192,7 @@ export function getDependencyGroups(
   oldDependencies: Index<string>,
   options: Options,
 ): { heading: string; groupName: string; packages: Index<string> }[] {
+  const chalk = options.color ? new Chalk.Instance({ level: 1 }) : Chalk
   const groups = keyValueBy<string, Index<string>>(newDependencies, (dep, to, accum) => {
     const from = oldDependencies[dep]
     const defaultGroup = partChanged(from, to)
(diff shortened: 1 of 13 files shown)


raineorshine:
Looks good!

> I really wanted to send the function SemVer objects instead of strings but I couldn't find a reliable way to parse version ranges into SemVer without either losing some data (like prerelease tags) or failing entirely and returning null. I think if this is treated as a power-user feature, making them parse it themselves isn't too bad.

You can use [parseRange](https://git.coolaj86.com/coolaj86/semver-utils.js#semverutils-parserange-rangestring) from `semver-utils`. That's how we pass the versions to the `--target` function:

https://github.com/raineorshine/npm-check-updates/blob/a63904b08a1444bd92bbb61112127215a21599ec/src/lib/queryVersions.ts#L4

https://github.com/raineorshine/npm-check-updates/blob/a63904b08a1444bd92bbb61112127215a21599ec/src/lib/queryVersions.ts#L46

You can add this before I merge if you'd like.

Another thing that would be good would be a unit test if you're up for it. We currently don't have test coverage for the `--format group` option, so we need a test there anyway.

CreativeTechGuy:
I saw `parseRange` but I didn't really like it. From looking at [the docs](https://www.npmjs.com/package/semver-utils#semverutilsparserangerangestring) of the type of response it can return, that'd be gnarly to try to use as a user and wouldn't give you the information you probably want. Really all a user cares about (I'd think) is the specific version which is currently installed and the specific version which will be installed. So if my `package.json` said  `>=3.x` and `3.4.1` was currently installed and `4.6.3` was about to be installed, ideally I'd want this `customizeGroup` function to receive `3.4.1` and `4.6.3` as the arguments instead of whatever the semver range happens to be. Maybe that's a bad example, but hopefully you get the idea. Someone would just care about the two specific versions that they are comparing when upgrading, not the ranges which might be applied as a result. I didn't see a good way to get that information without wiring up a lot of new data. Any suggestions or thoughts?

And I can try to write some tests once we get this nailed down. 😃

raineorshine:
> Really all a user cares about (I'd think) is the specific version which is currently installed and the specific version which will be installed. So if my `package.json` said `>=3.x` and `3.4.1` was currently installed and `4.6.3` was about to be installed, ideally I'd want this `customizeGroup` function to receive `3.4.1` and `4.6.3`

npm-check-updates actually ignores installed dependencies. This is by design. It only considers what is in the `package.json`. This avoids complications such as whether `node_modules` is up-to-date, the lock file is in sync, or the user has a package manager that stores dependencies somewhere else. In earlier versions of npm-check-updates, the installed package version was checked to avoid unnecessary upgrades, but this proved confusing so it was removed.

The implication is that `currentVersion` will always be a range, while `upgradedVersion` will always be a specific version.

> I saw `parseRange` but I didn't really like it. From looking at [the docs](https://www.npmjs.com/package/semver-utils#semverutilsparserangerangestring) of the type of response it can return, that'd be gnarly to try to use as a user and wouldn't give you the information you probably want.

In practice, the `parseRange` result is pretty easy to work with since it almost always has a length of 1. It only contains multiple entries for multi ranges like `>= 1.1.7 < 2.0.0 || 1.1.3`, which are typically only used in `peerDependencies` (not upgraded by default).

CreativeTechGuy:
Okay that all makes sense!

> The implication is that currentVersion will always be a range, while upgradedVersion will always be a specific version.

So are you thinking that we should use `parseRange` for to get `currentVersion` but something else for `upgradedVersion`? Or the same for both? How can/should we get the value for `upgradedVersion`?

So the function signature would be?

```ts
(packageName: string, currentVersion: SemVer[], upgradedVersion: SemVer, defaultGroup: UpgradeGroup) => UpgradeGroup
```

raineorshine:
> So are you thinking that we should use `parseRange` for to get `currentVersion` but something else for `upgradedVersion`? Or the same for both?

I would suggest using [parseRange](https://www.npmjs.com/package/semver-utils#semverutilsparserangerangestring) for `currentVersion` and [parse](https://www.npmjs.com/package/semver-utils#semverutilsparsesemverstring) for `upgradedVersion`.

> So the function signature would be?
> 
> ```ts
> (packageName: string, currentVersion: SemVer[], upgradedVersion: SemVer, defaultGroup: Upgr
> ```

To be complete, I would pass current version spec (`SemVer[]`),  upgraded version spec (`SemVer[]`), and upgraded version (`SemVer`). 

These correspond to `oldDependencies`, `newDependencies`, and `newVersions` respectively.

```ts
(packageName: string, 
 currentSpec: SemVer[], 
 upgradedVersionSpec: SemVer[], 
 upgradedVersion: SemVer, 
 defaultGroup: UpgradeGroup) => UpgradeGroup
```

https://github.com/raineorshine/npm-check-updates/blob/19c0503d195cd0617d721630072a4b07b217f835/src/lib/upgradePackageData.ts#L30-L36

> How can/should we get the value for `upgradedVersion`?

`semverUtils.parse(newVersions[dep])`

CreativeTechGuy:
There's some code paths which I couldn't figure out how to get the latest version map (sometimes called `newVersions` usually called `latest`). Those would be doctor and global. It seems like those should get this feature too, but I'm not sure the correct place to get the data from. I'm sure you know off the top of your head though 😄. I pushed what I have and fixed the conflicts.

And I poked around the tests but I couldn't see a good way to test the console output. Ideally I'd mock the `print` function or `console.log` and assert on the values that it was called with. This would probably even be a good case for a snapshot test of everything that was sent to stdout. I couldn't see a way to accomplish any of those things though with the current testing tools. Do let me know if I missed something or how to approach this.

raineorshine:
> There's some code paths which I couldn't figure out how to get the latest version map (sometimes called `newVersions` usually called `latest`). Those would be doctor and global. It seems like those should get this feature too, but I'm not sure the correct place to get the data from. I'm sure you know off the top of your head though 😄. I pushed what I have and fixed the conflicts.

I think you got them all. Global mode uses the same output function as Local mode, `printUpgrades`. Doctor mode is really just a task runner, so there is no special handling needed there.

> And I poked around the tests but I couldn't see a good way to test the console output. Ideally I'd mock the `print` function or `console.log` and assert on the values that it was called with. This would probably even be a good case for a snapshot test of everything that was sent to stdout. I couldn't see a way to accomplish any of those things though with the current testing tools. Do let me know if I missed something or how to approach this.

Those would be in https://github.com/raineorshine/npm-check-updates/blob/main/test/bin.test.ts. They spawn `npm-check-updates` and then you can assert stdout.

The one case that is harder to test is `--interactive`. We don't currently have any test coverage for interactive mode. If you want you can set up a test using [inject](https://github.com/terkelg/prompts#injectvalues), but you don't have to.

src/version-util.ts
```diff
@@ -34,6 +34,8 @@ interface UpgradeOptions {
   removeRange?: boolean
 }
 
+export type UpgradeGroup = 'major' | 'minor' | 'patch' | 'pre-v1' | 'none'
```
raineorshine, line 37: To follow the project's conventions for type definitions, please move `UpgradeGroup` to its own file at `/src/types/UpgradeGroup.ts`.

raineorshine (commented):
Sorry for the merge conflicts. I was finalizing #1158. Should be all clear now.

src/version-util.ts
CreativeTechGuy: Done!

CreativeTechGuy:
Sorry for the noise. I screwed up the merge resolution. It should be good now. Thanks for your patience!

raineorshine (changes requested):
Looks good! From what I can tell, there are just two tasks left:

- [ ] Custom groups
- [ ] Unit test

Currently, the group function can only return major, minor, patch, majorVersionZero, or none. In order to allow the user to define their own group, I suggest changing the function to return the custom header string, or null for the default group. For example, to group dependencies from a specific organization:

```
module.exports = {
  group: (name, current, upgraded) =>
    name.startsWith('@myorg') ? 'My Organization' : null
}
```

**Note**: I renamed `customizeGroups` to `group`.

CreativeTechGuy:
I never thought about the idea of custom groups. Is that something that we should support or is that scope creep? So you are suggesting that there's 3 different types of things this function can return:

* A ncu defined group
* null (equivalent of returning the ncu defined group)
* A custom group (which is implied if it's not any of the above)

If this is the direction you want to go, I'd suggest either changing the function signature to not have the 4th argument (the default group) if we are going to use `null` to represent the default group, or keep the 4th argument and not support `null`. It seems confusing to have both.

And do you have a list of tests you think should be written? It's hard to determine where the line is since there are no tests for any of the features this builds upon.

raineorshine:
> I never thought about the idea of custom groups. Is that something that we should support or is that scope creep?

Interesting, I thought that was the plan the whole time!

I would say it's not feature creep, as it's just adding more expressiveness to the same feature. There is no need to constrain the groups when the user can return any string.

> So you are suggesting that there's 3 different types of things this function can return:

* A ncu defined group
* null (equivalent of returning the ncu defined group)
* A custom group (which is implied if it's not any of the above)

Yes, exactly.

> If this is the direction you want to go, I'd suggest either changing the function signature to not have the 4th argument (the default group) if we are going to use `null` to represent the default group, or keep the 4th argument and not support `null`. It seems confusing to have both.

Yes, I agree. The function signature is already quite long, so getting rid of the 4th argument would be nice.

> And do you have a list of tests you think should be written? It's hard to determine where the line is since there are no tests for any of the features this builds upon.

I just added a test for `--format group`, so you have a good starting point:

https://github.com/raineorshine/npm-check-updates/blob/main/test/group.test.ts

I would suggest one test for each of the types of things the function can return (major/minor/patch/majorVersionMinor, custom, and null).

CreativeTechGuy:
> Interesting, I thought that was the plan the whole time!

Sorry for the confusion! I think I had brought it up as an idea but wasn't sure if it should be included or not.

> I just added a test for --format group, so you have a good starting point:

Thanks! Somehow I broke that test? I'll need to trace it to figure out where and why this is happening and then I'll write the remaining tests. (It is returning the `"Default stdin handling is deprecated."` error message for some reason.)

raineorshine:
> Thanks! Somehow I broke that test? I'll need to trace it to figure out where and why this is happening and then I'll write the remaining tests. (It is returning the `"Default stdin handling is deprecated."` error message for some reason.)

Maybe a bad merge? 

Note that the `Default stdin handling is deprecated` warning should not be a problem in itself. As long as you test the output with `stripAnsi(stdout).should.include(...)` it will just ignore it.

CreativeTechGuy:
> Note that the Default stdin handling is deprecated warning should not be a problem in itself. As long as you test the output with stripAnsi(stdout).should.include(...) it will just ignore it.

This was the key for me to figure it out. Thanks! The new code outputs the categories in a different order which is deterministic based on the order of the packages that need updating in the `package.json` list, but that's really weird so I'll make that more consistent.

We are getting there! Thanks for your help.

CreativeTechGuy:
Okay I think we are all good! It should be ready to go if you think so.

raineorshine (changes requested):
Looks great! 

Could you add a test for `--interactive` mode? That is the one other place that `--format group` is used. You can find a test to base it off here:

https://github.com/raineorshine/npm-check-updates/blob/main/test/interactive.test.ts#L81-L113

CreativeTechGuy:
> Could you add a test for --interactive mode?

I can do that, but I'm not sure what it will accomplish aside from ensuring it doesn't crash. The tests for interactive mode say they cannot assert on the output which is the only difference which a custom grouping function will have. I'll just copy the existing test and add the config option as a sanity test.

raineorshine:
Ah, good point. I forgot about that. Thanks anyway!

raineorshine:
Ah, good point. I guess we'll have to test manually. 

You can see that currently the custom group functionality does not work in interactive mode, as the logic lives in a different place: https://github.com/raineorshine/npm-check-updates/blob/main/src/logging.ts.

Unfortunately there is a lot of redundancy between `printUpgradesTable` and `runLocal`. I will leave it up to you whether you want to simply reproduce the logic from `runLocal`, or if you would like to try to abstract it.

Thanks!

CreativeTechGuy:
I think you are going to really like what I came up with. A huge chunk of logic is now shared and the code for generating the interactive choices is way simplified as a result!

raineorshine (changes requested):
> I think you are going to really like what I came up with. A huge chunk of logic is now shared and the code for generating the interactive choices is way simplified as a result!

That's much better!

I found one regression, and added a test in 77cb75a.

It looks like `printUpgradesTable` is being passed `latest` in place of `upgraded`. `latest` are the new exact versions, while `upgraded` is the result of passing the new versions to `upgradePackageData` which preserves ranges.

CreativeTechGuy:
I'm glad you haven't merged it yet! I've been thinking and I think we've gotten away from the original problem we are trying to solve. The most likely case for someone to use this is with a package like TypeScript which they would just be remapping any `'minor'` updates to `'major'`. This would be very easy to write if we still had the `defaultGroup` argument.

```js
(packageName, defaultGroup) => {
    if (packageName === "typescript" && defaultGroup === "minor") {
        return "major";
    }
}
```

But now that the default group has been removed, someone will have to compare the old and new version numbers to figure out what type of update it is. That is much less straight forward. So I'm arguing for adding back `defaultGroup` but as the second argument since it'll likely be far more commonly used than the SemVer options which are for more advanced use-cases. I think without this change, this feature will be too difficult to use and need too much explaining to figure out how to do basic things.

----

On an unrelated note, I've made the mistake a few times now of using `--group` when I should have used `--format group`. The fact these are named the same makes it really confusing since `--group` makes it seem like it'd enable grouping when in reality it is a grouping function. Would you be open to renaming `--group` to something more explicit so they don't share the same name?

raineorshine:
> So I'm arguing for adding back defaultGroup but as the second argument since it'll likely be far more commonly used than the SemVer options which are for more advanced use-cases. I think without this change, this feature will be too difficult to use and need too much explaining to figure out how to do basic things.

Yes, I forgot we needed to make the default group available to the function. I'm fine with removing `null` return in that case. 

> On an unrelated note, I've made the mistake a few times now of using --group when I should have used --format group. The fact these are named the same makes it really confusing since --group makes it seem like it'd enable grouping when in reality it is a grouping function. Would you be open to renaming --group to something more explicit so they don't share the same name?

I was trying to think of a way to combine them, but I want users to be able to define a custom group function in their .ncurc and then invoke `--format group` as needed on the command line. 

I would be okay with changing `group` to `groupFunction` if you prefer.

CreativeTechGuy:
I think we are in a really good place. Let me know what you think! :)

PR #1157, opened 2022-07-05
Update guide message with current options on done

lh0x00:
This commit will make the instruction command after running it more clear.

For example, with the user case using:
```bash
$ ncu -t minor -u 
```

The best suggestions are:
```bash
Run `ncu -t minor -u` to upgrade package.json
```
instead of (as it is):
```bash
Run `ncu -u` to upgrade package.json
```

After the PR is merged, you get the following output.
image


lamhieu-vk:
thanks for review @raineorshine 🍀

raineorshine (approved):
Thanks!

PR #1168, opened 2022-07-22
feat: remove scoped package / when pattern without /

magicdawn:
image

the reason for the screenshot is that `minimatch` treats `/` as a directory.
and if we want to update a scoped package, we must use a explicit slash like `*keyword*/*subkeyword*`

a solution may be:
if provided pattern does not include a slash, and the dep is a scoped package with a slash.
then replace `/` with `_`, so `@scope/name` turns to `@scope_name`, so that `*` can finally match the "slash"

an other solution may be, split the `@scope/name` and let split partial to match pattern,
like
```
(!pattern.includes('/') &&
              dependencyName.includes('/') &&
              dependencyName.split('/').some(partial => minimatch(partial, pattern)))
```

### DIFF
--- a/src/lib/filterAndReject.ts
+++ b/src/lib/filterAndReject.ts
@@ -1,4 +1,4 @@
-import { and } from 'fp-and-or'
+import { and, or } from 'fp-and-or'
 import identity from 'lodash/identity'
 import negate from 'lodash/negate'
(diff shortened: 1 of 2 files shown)


raineorshine (approved):
Looks great!

PR #1185, opened 2022-08-31
feature: cache

ly3xqhl8g9:
The `--cache` feature decreases significantly the `ncu` update time (at least 50%) when running `ncu` over multiple projects using similar packages. The feature is especially useful when running `ncu` programatically over many packages in a short amount of time. The `--cacheExpiration` and `--cacheFile` args can be used to control the caching behavior.

Example running `ncu --cache` on a new `create-react-app`:

ncu-cache

### DIFF
--- a/src/lib/getCacher.ts
+++ b/src/lib/getCacher.ts
@@ -49,7 +49,7 @@ export default function getCacher(runOptions: RunOptions): Cacher | undefined {
     const expired = checkCacheExpiration(cacheData, runOptions.cacheExpiration)
     if (expired) {
       // reset cache
-      fs.rmSync(cacheFile)
+      fs.promises.rm(cacheFile)
       cacheData = {}
     }
   } catch (error) {
@@ -70,8 +70,8 @@ export default function getCacher(runOptions: RunOptions): Cacher | undefined {
       return cacheData.packages ? cacheData.packages[key] : undefined
     },
     set: (key, value) => {
-      if (!key) return
-      cacheData.packages![key] = value
+      if (!key || !cacheData.packages) return
+      cacheData.packages[key] = value
     },
     save: async () => {
       await fs.promises.writeFile(cacheFile, JSON.stringify(cacheData))
--- a/src/lib/getCacher.ts
+++ b/src/lib/getCacher.ts
@@ -2,7 +2,8 @@ import fs from 'fs'
 import os from 'os'
 import path from 'path'
 import { CacheData, Cacher } from '../types/Cacher'
-import { RunOptions } from '../types/RunOptions'
+import { Options } from '../types/Options'
+import { print } from './logging'
 
 /**
  * Check if cache is expired if timestamp is set
@@ -31,22 +32,23 @@ export const resolvedDefaultCacheFile = path.join(os.homedir(), defaultCacheFile
  *
  * @returns
  */
-export default function getCacher(runOptions: RunOptions): Cacher | undefined {
-  if (!runOptions.cache) {
+export default function getCacher(options: Omit<Options, 'cacher'>): Cacher | undefined {
+  if (!options.cache) {
     return
   }
 
-  const file = runOptions.cacheFile
+  const file = options.cacheFile
   if (!file) {
     return
   }
   const cacheFile = file === defaultCacheFile ? resolvedDefaultCacheFile : file
   let cacheData: CacheData = {}
+  let cacheUsed: string[] = []
 
   try {
     cacheData = JSON.parse(fs.readFileSync(cacheFile, 'utf-8'))
 
-    const expired = checkCacheExpiration(cacheData, runOptions.cacheExpiration)
+    const expired = checkCacheExpiration(cacheData, options.cacheExpiration)
     if (expired) {
       // reset cache
       fs.promises.rm(cacheFile)
@@ -64,10 +66,14 @@ export default function getCacher(runOptions: RunOptions): Cacher | undefined {
   }
 
   return {
-    key: (name, version) => name + version,
+    key: (name, version) => name + '@' + version,
     get: key => {
-      if (!key) return
-      return cacheData.packages ? cacheData.packages[key] : undefined
+      if (!key || !cacheData.packages) return
+      const cached = cacheData.packages[key]
+      if (cached) {
+        cacheUsed = [...cacheUsed, key + `: ${cached}`]
+      }
+      return cached
     },
     set: (key, value) => {
       if (!key || !cacheData.packages) return
@@ -75,6 +81,18 @@ export default function getCacher(runOptions: RunOptions): Cacher | undefined {
     },
     save: async () => {
       await fs.promises.writeFile(cacheFile, JSON.stringify(cacheData))
+    },
+    log: () => {
+      if (cacheUsed.length === 0) return
+
+      const message = '\nUsing cached package versions'
+
+      if (options.loglevel === 'verbose') {
+        print(options, message + ':', 'verbose')
+        cacheUsed.forEach(cache => print(options, `  ${cache}`, 'verbose'))
+      } else {
+        print(options, message, 'warn')
+      }
     },
   } as Cacher
 }
--- a/src/lib/getCacher.ts
+++ b/src/lib/getCacher.ts
@@ -1,0 +1,80 @@
+import fs from 'fs'
+import os from 'os'
+import path from 'path'
+import { CacheData, Cacher } from '../types/Cacher'
+import { RunOptions } from '../types/RunOptions'
+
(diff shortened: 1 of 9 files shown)


ly3xqhl8g9:
The tests were failing at the fork. The 'cache' suite's only test 'generates a cache file' passes.

raineorshine:
Thanks for your contribution! I am traveling this week but I'll take a closer look when I return. 

This will be great for running ncu multiple times in a short period. I'm not sure how it is improving the performance of a single run, as the version fetch is already memoized, but maybe I am missing something.

ly3xqhl8g9:
Running `ncu --cache` for the first time in a package `one` merely populates the cache and has completely the same performance, say 10 seconds on a slow connection. Running right after another `ncu --cache`, another shell, another call, in a package `two` which has similar dependencies as package `one` will circumvent the entire logic from `src/lib/queryVersions.ts@getPackageVersionProtected` to resolve the new version and can bring the update time down to even 1-2 seconds if many of the dependencies overlap, as is often the case with development dependencies even in unrelated projects, and is even more the case in a mono-repo. The fetch memoization is not used across multiple `ncu` calls in different shells, different processes.

raineorshine:
Ah, of course! That's what I get for doing a code review on my phone ;)

src/lib/queryVersions.ts
```diff
@@ -44,6 +53,17 @@ async function queryVersions(packageMap: Index<VersionSpec>, options: Options =
   async function getPackageVersionProtected(dep: VersionSpec): Promise<VersionResult> {
     const npmAlias = parseNpmAlias(packageMap[dep])
     const [name, version] = npmAlias || [dep, packageMap[dep]]
+
+    const cacheKey = options.cacher?.key(name, version)
+    const cached = options.cacher?.get(cacheKey)
+    if (cached) {
+      barTick()
```
raineorshine, line 60: I would suggest `bar?.tick?.()` rather than creating a new function.

raineorshine (changes requested):
Looks good! Just a few minor requests.

raineorshine (approved):

raineorshine:
Thanks!

One thing I would like to add is a brief log when there are cache hits: `Using cached package versions`. Or in verbose mode:

```
Using cached package versions:
 
  chalk@0x.y.z
  react@0x.y.z
  typescript@0x.y.z
```

This is motivated by the fact that silent cache hits are particularly nefarious when the user expects something different. For example, they may add `cache: true` to their ncurc and forget it is there until a moment when they can't understand why they are not getting the latest version. The message will provide some feedback that will help avoid potential confusion. 

If you would like to add this I would be most grateful. If not, I will try to add it as soon as I have some free time. Thanks!

ly3xqhl8g9:
Added the cache log, looks like this:

log

src/lib/getCacher.ts
```diff
@@ -0,0 +1,98 @@
...
+  const file = options.cacheFile
+  if (!file) {
+    return
+  }
+  const cacheFile = file === defaultCacheFile ? resolvedDefaultCacheFile : file
+  let cacheData: CacheData = {}
+  let cacheUsed: string[] = []
+
+  try {
+    cacheData = JSON.parse(fs.readFileSync(cacheFile, 'utf-8'))
```
raineorshine, line 49: Use async

src/lib/initOptions.ts
```diff
@@ -140,6 +141,9 @@ async function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } =
     ...(options.interactive && options.upgrade === undefined ? { upgrade: !json } : null),
     packageManager,
   }
+  resolvedOptions.cacher = getCacher(resolvedOptions)
```
raineorshine, line 144: Please place this within the object literal initializer rather than as an assignment on a separate line.

src/lib/getCacher.ts
```diff
@@ -0,0 +1,98 @@
...
+ * for quick updates across `ncu` calls.
+ *
+ * @returns
+ */
+export default function getCacher(options: Omit<Options, 'cacher'>): Cacher | undefined {
+  if (!options.cache) {
+    return
+  }
+
+  const file = options.cacheFile
```
raineorshine, line 40: Let's use `options.cacheFile` inline since it is so short rather than introducing another variable.

src/lib/getCacher.ts
```diff
@@ -0,0 +1,98 @@
...
+      await fs.promises.writeFile(cacheFile, JSON.stringify(cacheData))
+    },
+    log: () => {
+      if (cacheUsed.length === 0) return
+
+      const message = '\nUsing cached package versions'
+
+      if (options.loglevel === 'verbose') {
+        print(options, message + ':', 'verbose')
+        cacheUsed.forEach(cache => print(options, `  ${cache}`, 'verbose'))
```
raineorshine, line 92: I have a small change I'd like to make to the verbose log. Let's output an object of cached versions. It would look like this:

```
Using 2 cached package versions:
{
  "react": "x.y.z"
  "typescript": "x.y.z"
}
```

Note that it should only show the cache hits, not the entire cache.

src/lib/getCacher.ts
```diff
@@ -0,0 +1,98 @@
...
+      if (!key || !cacheData.packages) return
+      cacheData.packages[key] = value
+    },
+    save: async () => {
+      await fs.promises.writeFile(cacheFile, JSON.stringify(cacheData))
+    },
+    log: () => {
+      if (cacheUsed.length === 0) return
+
+      const message = '\nUsing cached package versions'
```
raineorshine, line 88: Let's add the number of cache hits since we have it: `Using 6 cached package versions`

raineorshine (changes requested):

src/lib/initOptions.ts
ly3xqhl8g9: The conundrum is that I wanted to keep the pattern of passing the entire `Options` object to `getCacher`, so that it can be passed to `print` and other functions, hence why the signature for `getCatcher` is  `options: Omit<Options, 'cacher'>`. Since `getCatcher` returns an `Options` field but `getCatcher` requires the `Options` as parameter, it's sort of a chicken or the egg causality dilemma. How would like to solve this?

src/lib/getCacher.ts
ly3xqhl8g9: By cache hit you mean a difference in package versions? The cacher caches everything, so a package which does not change `one@1.0.0` is also cached with the version `1.0.0` and future queries for the package `one` will also be a cache hit, even if the same version. This is what decreases the update time since it goes over redundant checks, and since the default cache expiration is only 10 minutes, there is a very small chance that the package was updated precisely in those 10 minutes.

src/lib/initOptions.ts
raineorshine: Ah, I see. This is fine then.

`Options` is kind of intended to be serializable, but storing the cacher object in options is fine.

src/lib/getCacher.ts
raineorshine: What I mean is, it should only show cached packages that have upgrades. If the user really needs to look at the entire cache (which I imagine would be rare), they can look at the cache file itself. Within the context of running ncu in verbose mode, we want to tell them exactly which upgrades came from the cache.

ly3xqhl8g9: Right. In the example from above in a React package we have:

Screenshot 2022-09-04 at 10 10 56

`react@^18.2.0` is also a cache hit, even if the 'new' version is the same version, `18.2.0`. You would like to display only `@types/react-dom` since it's a cache hit with a package upgrade, correct?

raineorshine: Yes, exactly.

ly3xqhl8g9:
The logs look like this now:

log

Thank you for taking your time with this PR. Truth be told, having this deployed would be really useful to me first of all. I regularly run `ncu` over 100+ packages.

ly3xqhl8g9:
The logs look like this now:

log

Thank you for taking your time with this PR. Truth be told, having this deployed would be really useful to me first of all. I regularly run `ncu` over 100+ packages.

raineorshine (approved):

raineorshine:
Looks great, thanks!

raineorshine:
@ly3xqhl8g9 Is there a reason we are including the old version in the cache key?

ly3xqhl8g9:
Initially, I thought it would be needed for manual inspection. In practice these past months I found it is simply easier to `rm ~/.ncu-cache.json`. Wanted to submit another PR for a `ncu cache-clean` command which would erase all known caches, if any, but didn't get to it. I will try tomorrow.

PR #1189, opened 2022-09-14
Support npm config cafile which includes a tilde

darkv:
**Problem**

npm-check-updates fails when reading in the npm configuration if cafile setting points to a file relative to the user‘s home directory, i.e. starts with a tilde.

**Why**

In corporate environment you often have no direct access to npmjs.org and have to use a proxy instead. As such internal proxies often use self signed certificate, you need to configure npm to use your own ca for verification. It is not unusual to distribute a general .npmrc config file that contains something like

> registry=https://my.private.npm.proxy
> cafile=~/path/to/ca.pem

Where npm is working fine, npm-check-updates fails to locate the ca file as it cannot interpret the tilde for the current user‘s home and exits with an error:

> Error: ENOENT: no such file or directory, open '~/path/to/ca.pem'

This pull request adds the npm module [untildify](https://github.com/sindresorhus/untildify) to convert such kind of path to an absolute one so that the referenced file can be read successfully by ncu.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,6 +34,7 @@
         "semver-utils": "^1.1.4",
         "source-map-support": "^0.5.21",
         "spawn-please": "^1.0.0",
+        "untildify": "^4.0.0",
         "update-notifier": "^6.0.2",
         "yaml": "^2.1.1"
       },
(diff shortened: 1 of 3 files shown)


raineorshine (approved):
Looks good! Thanks!

PR #1195, opened 2022-09-26
Fix module usage docs

qmhc:
When using `ncu` as module in TypeScript, follow the docs usage will get a type error:

image

Because the type exports of default is only the `run` method.

I think it may better to change the usage to `import * as ncu from 'npm-check-updates'`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,9 +322,9 @@ For example, `.ncurc.json`:
 npm-check-updates can be imported as a module:
 
 ```js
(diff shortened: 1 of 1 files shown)


raineorshine (changes requested):
Thanks for the fix. Let's actually do it the other way around since there is only one import we care about: change `ncu.run(...)` to `ncu(..)`.

raineorshine (approved):

PR #1197, opened 2022-09-29
fix: convert Windows path to posix path for globby

### LINKED ISSUE #1196: `ncu -ws --root -i` is not working in pnpm workspaces.
- [x] I am using `node >= 14`

---

## Steps to Reproduce

See this repo: https://github.com/imguolao/ncu-pnpm-workspaces

## Current Behavior

Only working in the root.

image

No anything hanppen.

image

os: windows 10
node: v18.2.0
ncu@version: 16.3.2

imguolao:
fix #1196

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -226,7 +226,7 @@ export async function run(
       }
       const workspacePackageGlob = ([] as string[])
         .concat(workspaces || [])
-        .map(workspace => path.join(workspace, defaultPackageFilename))
+        .map(workspace => path.join(workspace, defaultPackageFilename).replace(/\\/g, '/'))
       pkgs = [
         // include root project package file when --root is used
         ...(options.root ? pkgs : []),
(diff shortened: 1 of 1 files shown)


rchl:
There is a `path.posix` export in Node that sounds more appropriate to me that manual replace.
https://nodejs.org/api/path.html#pathposix

imguolao:
> There is a `path.posix` export in Node that sounds more appropriate to me that manual replace. https://nodejs.org/api/path.html#pathposix
> 
> (Disclaimer: I know knowing about the issue at hand, just advocating for `path.posix` over manual replace)

I think we should let @raineorshine decide.

raineorshine:
> There is a `path.posix` export in Node that sounds more appropriate to me that manual replace.
> [nodejs.org/api/path.html#pathposix](https://nodejs.org/api/path.html#pathposix)

Those are posix-specific versions of node's built-in `path` module, but in this case it is the external library `globby` that is the problem. There are no direct methods in `node/posix` for converting from Win32 to POSIX that I am aware of.

PR #1210, opened 2022-10-28
build: add missing dependency ini

meteorlxy:
We are importing `ini` directly, but missing in dependencies. This will cause an error when installing ncu with pnpm.

https://github.com/raineorshine/npm-check-updates/blob/fd2f566baf7e74a668e94f8e1e14a43966341c4e/src/package-managers/npm.ts#L3

image


raineorshine:
Good catch, thanks!

PR #1219, opened 2022-11-11
Cache: Clear cache option

ly3xqhl8g9:
Added the `cacheClear` option.

### DIFF
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -277,13 +277,11 @@ const cliOptions: CLIOption[] = [
   {
     long: 'cache',
     description: 'Cache versions to the cache file',
-    default: false,
     type: 'boolean',
   },
   {
     long: 'cacheClear',
     description: 'Remove the default cache file',
-    default: false,
     type: 'boolean',
   },
   {
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -281,6 +281,12 @@ const cliOptions: CLIOption[] = [
     type: 'boolean',
   },
   {
+    long: 'cacheClear',
+    description: 'Remove the default cache file',
+    default: false,
+    type: 'boolean',
+  },
+  {
     long: 'cacheExpiration',
     arg: 'min',
     description: 'Cache expiration in minutes',
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -281,7 +281,7 @@ const cliOptions: CLIOption[] = [
   },
   {
     long: 'cacheClear',
-    description: 'Remove the default cache file',
+    description: 'Clear the default cache, or the cache file specified by --cacheFile',
     type: 'boolean',
   },
   {
--- a/src/lib/cache.ts
+++ b/src/lib/cache.ts
@@ -28,8 +28,17 @@ export const defaultCacheFilename = '.ncu-cache.json'
 export const defaultCacheFile = `~/${defaultCacheFilename}`
 export const resolvedDefaultCacheFile = path.join(os.homedir(), defaultCacheFilename)
 
+/** Remove the default cache file. */
+export async function cacheClear() {
+  try {
+    fs.promises.rm(resolvedDefaultCacheFile)
+  } catch (error) {
+    // ignore file read/parse/remove errors
+  }
+}
+
 /**
- * The cacher stores key (name + version) - value (new version) pairs
+ * The cacher stores key (name + target) - value (new version) pairs
  * for quick updates across `ncu` calls.
  *
  * @returns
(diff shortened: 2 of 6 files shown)


src/cli-options.ts
```diff
@@ -280,6 +280,12 @@ const cliOptions: CLIOption[] = [
     default: false,
     type: 'boolean',
   },
+  {
+    long: 'cacheClear',
+    description: 'Remove the default cache file',
+    default: false,
```
raineorshine, line 286: Let's omit the `default` property since it is assumed to be false when unspecified.

src/lib/cache.ts
```diff
@@ -28,8 +28,17 @@ export const defaultCacheFilename = '.ncu-cache.json'
 export const defaultCacheFile = `~/${defaultCacheFilename}`
 export const resolvedDefaultCacheFile = path.join(os.homedir(), defaultCacheFilename)
 
+/** Remove the default cache file. */
+export async function cacheClear() {
+  try {
+    fs.promises.rm(resolvedDefaultCacheFile)
```
raineorshine, line 34: This should handle a custom `--cacheFile` also.

src/lib/cache.ts
```diff
@@ -28,8 +28,17 @@ export const defaultCacheFilename = '.ncu-cache.json'
...
+  } catch (error) {
+    // ignore file read/parse/remove errors
+  }
+}
+
 /**
- * The cacher stores key (name + version) - value (new version) pairs
+ * The cacher stores key (name + target) - value (new version) pairs
```
raineorshine, line 41: Thanks

raineorshine (changes requested):
Thanks! Can you add a unit test?

src/lib/cache.ts
ly3xqhl8g9: It replicates `rm` behaviour in a cumbersome way. Perhaps a better design would be to store all the custom cache filepaths in the default cache and simply remove them altogether at `cacheClear`.

raineorshine: Currently, the caching mechanism uses a single cache file, either the default or a specified file. It would make sense to me that `--cacheClear` would also operate on a single cache file, either the default or a specified file. So I would prefer that `--cacheFile` determine the cache file in both cases.

raineorshine: Note that if you add `{ force: true }` you don't need to wrap it in a try-catch.

src/cli-options.ts
```diff
@@ -277,7 +277,11 @@ const cliOptions: CLIOption[] = [
   {
     long: 'cache',
     description: 'Cache versions to the cache file',
-    default: false,
+    type: 'boolean',
+  },
+  {
+    long: 'cacheClear',
+    description: 'Remove the default cache file',
```
raineorshine, line 284: Let's set the description to `Clears the default cache, or the cache file specified by --cacheFile`.

src/lib/cache.ts
```diff
@@ -28,8 +28,22 @@ export const defaultCacheFilename = '.ncu-cache.json'
...
 
+/** Resolve the cache file path based on os/homedir */
+export function resolveCacheFile(optionsCacheFile: string) {
+  return optionsCacheFile === defaultCacheFile ? resolvedDefaultCacheFile : optionsCacheFile
+}
+
+/** Remove the default cache file. */
+export async function cacheClear(options: Options) {
+  if (!options.cacheFile) {
+    return
```
raineorshine, line 39: If `--cacheFile` is not specified, it should clear the default cache.

test/cache.test.ts
```diff
@@ -89,4 +89,32 @@ describe('cache', () => {
...
+        'ncu-test-tag': '1.0.0',
+        // latest: no upgrade
+        // greatest: prerelease → 2.0.0-alpha.2
+        'ncu-test-alpha': '1.0.0',
+      },
+    }
+
+    await ncu.run({ packageData, cache: true })
+    const cacheData: CacheData = await fs.readFile(resolvedDefaultCacheFile, 'utf-8').then(JSON.parse)
+    expect(cacheData.timestamp).lessThanOrEqual(Date.now())
```
raineorshine, line 109: You can remove the timestamp assertion from this test since it is not relevant here.

raineorshine (changes requested):

src/lib/cache.ts
```diff
@@ -28,8 +28,22 @@ export const defaultCacheFilename = '.ncu-cache.json'
 export const defaultCacheFile = `~/${defaultCacheFilename}`
 export const resolvedDefaultCacheFile = path.join(os.homedir(), defaultCacheFilename)
 
+/** Resolve the cache file path based on os/homedir */
+export function resolveCacheFile(optionsCacheFile: string) {
+  return optionsCacheFile === defaultCacheFile ? resolvedDefaultCacheFile : optionsCacheFile
+}
+
+/** Remove the default cache file. */
```
raineorshine, line 36: This comment needs to be updated to match the new functionality.

raineorshine (approved):

PR #1223, opened 2022-11-17
Removed some off eslint rules

piwysocki:
Some rulest are off and can be removed.

Fixed code for rule: no-use-before-define in two files.


raineorshine (approved):
Thanks!

Issue #1215, opened 2022-11-08
Filter/reject using the registry version, not only the package version.


wwwouter:
I like to filter or reject using the registry version.
Right now the `filter` and `reject` functions only have the current package version as parameter.
I would like an extra parameter, with the version ncu plans on updating the package to.
Or if this is not possible, maybe a new function.

I currently have two use cases.

The first is "say no, to dot oh". So if a new version of `x.0.0` is proposed, I want to skip that.

The second is that sometimes the current version of a package still has some problems and I want to wait for the next release. In my case it's `rc-slider`. I want to upgrade to the version that will be released after `10.0.1`. I don't care if it's `10.0.2`, `10.1.0` or `11.0.1`.

So my `reject` function would look like something like this:
```
  reject: (packageName, currentSemVer, proposedSemVer) => {
        if (proposedSemVer[0].minor === '0' && proposedSemVer[0].patch === '0') {
            return true;
        }
        if (packageName === 'rc-slider' && proposedSemVer[0].semver === '10.0.1') {
            return true;
        }

        return false;
    },
```

Is this something that you would like to support? 

If so, I can try to create a PR.

BTW, I tried `rc-slider  8.x || >=10.0.2` in my package.json.
 
But with `upgrade:false`, the result was

```
rc-slider  8.x || >=10.0.2  →  10.x
```

And with `upgrade:true`

```
rc-slider  8.x || >=10.0.2  →  10.0.1
```

- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [x] I am using the latest version of `npm-check-updates`
- [x] I am using `node >= 14.14`

---

## Steps to Reproduce

.ncurc:

```js

```

Dependencies:

```json

```

Steps:

wwwouter:
Thanks, I'll try to have a look this week.

Nevermind the "BTW", I misunderstood how npm-check-updates  works.

raineorshine:
Implemented in #1290 as `filterResult`.

PR #1226, opened 2022-11-23
filterMeta

wwwouter:
Based on #1215 

@raineorshine I want to add some tests, but wasn't sure how/where. Is there already test code that deals with the fetched data? Where would you like me to add the tests for `filterFetched`?

### DIFF
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -393,6 +393,13 @@ const cliOptions: CLIOption[] = [
     type: 'string | string[] | RegExp | RegExp[] | FilterFunction',
   },
   {
+    long: 'filterMeta',
+    arg: 'fn',
+    description: `Filter on Packument using predicate function.`,
+    type: 'FilterMetaFunction',
+    help: extendedHelpGroup,
+  },
+  {
     long: 'filterVersion',
     arg: 'p',
     description: 'Filter on package version using comma-or-space-delimited list, /regex/, or predicate function.',
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -111,6 +111,23 @@ npm run test
 Saving partially upgraded package.json
 `
 
+/** Extended help for the filterMeta option. */
+const extendedHelpFilterMeta = (): string => {
+  return `Filter packages after the metadata is fetched. Only available in .ncurc.js or when importing npm-check-updates as a module:
+
+  ${chalk.gray(`/**
+    @param packument        The object returned from NPM.
+    @returns                Boolean value, true to include the dependency in the output, false to exclude it.
+  */`)}
+  ${chalk.cyan('filterMeta')}: (packument} {
+    if (packument.name === 'typescript') {
+      return true
+    }
+    return return false
+  }
+`
+}
+
 /** Extended help for the --format option. */
 const extendedHelpFormat = (): string => {
   const header =
@@ -397,7 +414,7 @@ const cliOptions: CLIOption[] = [
     arg: 'fn',
     description: `Filter on Packument using predicate function.`,
     type: 'FilterMetaFunction',
-    help: extendedHelpGroup,
+    help: extendedHelpFilterMeta,
   },
   {
     long: 'filterVersion',
--- a/test/filter.test.ts
+++ b/test/filter.test.ts
@@ -41,6 +41,21 @@ describe('filter', () => {
     upgraded.should.have.property('lodash.filter')
   })
 
+  it('filter metadata', async () => {
+    const upgraded = (await ncu({
+      packageData: {
+        dependencies: {
+          lodash: '2.0.0',
+          'lodash.map': '2.0.0',
+          'lodash.filter': '2.0.0',
+        },
+      },
+      filterMeta: packument => packument.name === 'lodash.map',
+    })) as Index<string>
+    upgraded.should.have.property('lodash.map')
+    upgraded.should.not.have.property('lodash.filter')
+  })
+
   it('filter with wildcard for scoped package', async () => {
     const pkg = {
       dependencies: {
--- a/README.md
+++ b/README.md
@@ -167,6 +167,8 @@ ncu "/^(?!react-).*$/" # windows
 -f, --filter <p>           Include only package names matching the given
                            string, wildcard, glob, comma-or-space-delimited
                            list, /regex/, or predicate function.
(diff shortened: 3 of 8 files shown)


raineorshine:
Let's add tests to filter.test.ts. You can use this test as an example:

https://github.com/raineorshine/npm-check-updates/blob/main/test/filter.test.ts#L29-L42

Also, please rename to `filterMeta`. Sorry if that wasn't clear from my previous comments. I want the name to reflect the fact that it can filter on any package metadata.

Thanks!

wwwouter:
Ok, I renamed the function and added a simple test.
Are there any more test you would like me to add?

src/cli-options.ts
```diff
@@ -392,6 +392,13 @@ const cliOptions: CLIOption[] = [
       'Include only package names matching the given string, wildcard, glob, comma-or-space-delimited list, /regex/, or predicate function.',
     type: 'string | string[] | RegExp | RegExp[] | FilterFunction',
   },
+  {
+    long: 'filterMeta',
+    arg: 'fn',
+    description: `Filter on Packument using predicate function.`,
+    type: 'FilterMetaFunction',
+    help: extendedHelpGroup,
```
raineorshine, line 400: This needs to have its own help documentation, not `extendedHelpGroup`.

test/filter.test.ts
```diff
@@ -41,6 +41,21 @@ describe('filter', () => {
     upgraded.should.have.property('lodash.filter')
   })
 
+  it('filter metadata', async () => {
```
raineorshine, line 44: Looks good.

raineorshine (changes requested):
Looks good! Just a few small requests. 

One thing we need to do is fill out the `Packument` type. It currently only defines fields that are used internally, but if we're going to export it we ought to have a more complete set of properties. It doesn't have to be perfect, but let's cover the most common fields. 

https://github.com/raineorshine/npm-check-updates/blob/main/src/types/Packument.ts

raineorshine:
Let's use `pacote.Packument` directly! I don't know why I didn't use that internally to begin with.

src/cli-options.ts
```diff
@@ -111,6 +111,23 @@ npm run test
...
+  ${chalk.gray(`/**
+    @param packument        The object returned from NPM.
+    @returns                Boolean value, true to include the dependency in the output, false to exclude it.
+  */`)}
+  ${chalk.cyan('filterMeta')}: (packument} {
+    if (packument.name === 'typescript') {
+      return true
+    }
+    return return false
+  }
```
raineorshine, line 127: Simpler: `packument => packument.name === 'typescript'`.

However, I don't think `packument.name` is the best example to use here since it is available pre-fetch. There would be no reason to filter on the name in this way. Instead, let's give an example of filtering on version like your use case.

src/cli-options.ts
```diff
@@ -111,6 +111,23 @@ npm run test
 Saving partially upgraded package.json
 `
 
+/** Extended help for the filterMeta option. */
+const extendedHelpFilterMeta = (): string => {
+  return `Filter packages after the metadata is fetched. Only available in .ncurc.js or when importing npm-check-updates as a module:
+
+  ${chalk.gray(`/**
+    @param packument        The object returned from NPM.
```
raineorshine, line 119: Let's say: `The metadata object returned from the npm registry.`

raineorshine (changes requested):

raineorshine:
@wwwouter Let me know if I can provide further guidance!

wwwouter:
@raineorshine sorry, it's quite busy.

I took a quick look at using `pacote.Packument`, but a lot of errors appear, so I don't think that's a good idea for this PR.

raineorshine:
> I took a quick look at using `pacote.Packument`, but a lot of errors appear, so I don't think that's a good idea for this PR.

Sure, no problem

src/cli-options.ts
```diff
@@ -392,6 +409,13 @@ const cliOptions: CLIOption[] = [
       'Include only package names matching the given string, wildcard, glob, comma-or-space-delimited list, /regex/, or predicate function.',
     type: 'string | string[] | RegExp | RegExp[] | FilterFunction',
   },
+  {
+    long: 'filterMeta',
+    arg: 'fn',
+    description: `Filter on Packument using predicate function.`,
```
CreativeTechGuy, line 415: As an average user, this description creates more questions than it answers. It'd be helpful if the description was more like the extended help docs.

Suggestion: `Filter packages given the metadata about the new version.`

CreativeTechGuy:
I love this feature, but it seems like it's might be missing a critical part. (Or at least a test to ensure this behavior is working.)

If I'm on `1.0.0` and the version that NCU wants to install is `2.0.0` and I write a filter to say no packages with `x.0.0`, then the update will be rejected entirely. Let's say there's a `1.5.2` that exists as the last version before the major version bump. Logically if you wrote that `x.0.0` filter, you'd want `1.5.2` to be installed instead as it's the latest version which isn't filtered out.

This might be what is already happening given the code, but it seems important enough to ensure there's a test for this.

wwwouter:
@CreativeTechGuy that sounds like a good test to add.

I'm busy with other stuff this week, but this is still on my todo list 😄

raineorshine:
> If I'm on `1.0.0` and the version that NCU wants to install is `2.0.0` and I write a filter to say no packages with `x.0.0`, then the update will be rejected entirely. Let's say there's a `1.5.2` that exists as the last version before the major version bump. Logically if you wrote that `x.0.0` filter, you'd want `1.5.2` to be installed instead as it's the latest version which isn't filtered out.

This would make the most sense intuitively. However, there is a problem in that the registry does not contain a history of versions published to `latest`. What that means is that we may be able to see that `1.5.2` was been published, but we don't know that it was published to `latest`. Assuming such would be inappropriate for the default `--target latest` which strictly upgrades the version published to the `latest` tag. 

The behavior you describe could be achieved by giving up the `latest` requirement, and instead using `--target newest`, or better, `--target newest --pre 0`. We should at least document this in the extended help.

raineorshine:
This has been implemented in #1290 which added `--filterResult`.

One thing that this PR contains that #1290 does not is the package's Packument. If you would like to pass the Packument to the predicate function, please open a new PR.

PR #1231, opened 2022-12-05
Add code coverage test-run

doublethefish:
chore(coverage): adds code coverage to tests

Rather than replace the existing test, which would require merging the
coverage output of the two test-run phases (`test:src` and `test:timeout`),
we add a new 'test:src:cov' script.

There's a couple of things we notice doing this; the first is that some
of the tests, especially doctor tests, do not report coverage because
they use sub-procs via spawn. This leads us to realise that the
test-types sometimes veer into 'integration' type tests.

We configure the coverage to:
 * cover test-code in addition to run-time code.
 * use high- and low-tide marks that seem reasonable for the library.
 * set a global pass threshold that currently passes but should catch
   most regressions.
 * print coverage results to the console.
 * create a nice html coverage report in './coverage/index.html'.
 * only show the non-fully covered files in text output

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 build/
+coverage/
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "prettier": "prettier .",
     "test": "npm run test:src && npm run test:timeout",
     "test:src": "mocha test test/package-managers/npm test/package-managers/yarn test/package-managers/staticRegistry",
+    "test:src:cov": "npm run c8 npm run test:src",
     "test:timeout": "mocha --exit test/timeout",
     "ncu": "node build/src/bin/cli.js"
   },
--- a/src/index.ts
+++ b/src/index.ts
@@ -72,7 +72,7 @@ const getPackageManagerForInstall = async (options: Options, pkgFile: string) =>
   return pnpm ? 'pnpm' : 'npm'
 }
 
-/** Either suggest an install command based on the package manager, or in interactive mode, prompt to autoinstall. */
+/** Either suggest an install command based on the package manager, or in interactive mode, prompt to auto-install. */
 const npmInstall = async (
   pkgs: string[],
   analysis: Index<PackageFile> | PackageFile,
@@ -85,8 +85,8 @@ const npmInstall = async (
   const someUpgraded = Object.values(analysisNormalized).some(upgrades => Object.keys(upgrades).length > 0)
   if (!someUpgraded) return
 
-  // for the purpose of the install hint, just use the package manager used in the first subproject
-  // if autoinstalling, the actual package manager in each subproject will be used
+  // for the purpose of the install hint, just use the package manager used in the first sub-project
+  // if auto-installing, the actual package manager in each sub-project will be used
   const packageManager = await getPackageManagerForInstall(options, pkgs[0])
 
   // by default, show an install hint after upgrading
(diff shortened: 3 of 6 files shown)


raineorshine:
Hi, thanks for the PR.

You may already be aware, but we currently have test coverage in the github action:

https://github.com/raineorshine/npm-check-updates/blob/2be014c9943cdfdadb8548fd6329594e185174bc/.github/workflows/test.yml#L54-L62

Can you explain what your PR offers beyond the current coverage? If there are additional benefits, I am open to discussing an integrated solution.

doublethefish:
> You may already be aware, but we currently have test coverage in the github action:

Ah, no I had missed that CiCd step entirely. Kicking myself as I am sure I'd seen the badge when I first looked at the project.

> Can you explain what your PR offers beyond the current coverage? If there are additional benefits, I am open to discussing an integrated solution.

Absolutely, 

The benefits of the ability to get coverage data in the local checkout are:

-  you can check iteratively if you have added un-covered code (and see your progress when improving coverage - as I am)
- it's possible to add husky pre-push checks to ensure the coverage is above the minimum (before wasting CiCd resources).
- it's more obvious it exists in the package.

Personally, I have found that having coverage front-and-centre (obvious, easy, quick) mostly helps code quality, with caveats.

raineorshine:
> > You may already be aware, but we currently have test coverage in the github action:
> 
> Ah, no I had missed that CiCd step entirely. Kicking myself as I am sure I'd seen the badge when I first looked at the project.

No worries, it's a little out of the way.

> The benefits of the ability to get coverage data in the local checkout are:
> 
> * you can check iteratively if you have added un-covered code (and see your progress when improving coverage - as I am)
> * it's possible to add husky pre-push checks to ensure the coverage is above the minimum (before wasting CiCd resources).
> * it's more obvious it exists in the package.
> 
> Personally, I have found that having coverage front-and-centre (obvious, easy, quick) mostly helps code quality, with caveats.

That sounds reasonable to me. Now, can we do both local coverage and coveralls with just `c8`?I don't think we need both `c8` and `nyc`. However if there are features that `nyc` has that `c8` doesn't have I'm happy to switch over. I'm not super familiar with the differences myself.

doublethefish:
>  can we do both local coverage and coveralls with just c8? I don't think we need both c8 and NYC.

great question, I would hope/assume so, because you're right, we don't need/want both. I've not yet come across c8, but I do find NYC problematic, especially since they restructured everything from istanbul->nyc.

I'll have a look at c8.

doublethefish:
That was easier than I expected.

I don't know how you feel about covering test code, as well as production code, but I've found that test-code coverage occasionally highlights regressions in tests. I've not reintroduced that configuration change for c8, but it looks straight forward enough to do at some point in the future.

I hope you don't mind, I've thrown some typo fixes in (found via [cspell](https://github.com/streetsidesoftware/cspell), which might be worth integrating).

raineorshine (approved):

raineorshine:
> That was easier than I expected.

Great!

> I don't know how you feel about covering test code, as well as production code, but I've found that test-code coverage occasionally highlights regressions in tests. I've not reintroduced that configuration change for c8, but it looks straight forward enough to do at some point in the future.

I think with a larger codebase and lots of contributors this could be useful. npm-check-updates is pretty small and I personally review all the code, so it may not be as relevant here. Not that I trust my judgment over automated tools, just that changes to tests are pretty visible with the current PR review process.

> I hope you don't mind, I've thrown some typo fixes in (found via [cspell](https://github.com/streetsidesoftware/cspell), which might be worth integrating).

Thank you!

PR #1232, opened 2022-12-05
chore(test-data): moves test-data files into own dir

doublethefish:
The intent here is to make code-discovery easier as it separates code
from data more clearly.

One of the outcomes might be to add tests for yarn-workspaces
specifically in doctor.js, which seems to have some bugs determining
clean version info for package upgrade candidates.


raineorshine (approved):
Looks good!

doublethefish:
Excellent, I'm chuffed this went in.

Out of interest, are those windows failures common?

raineorshine:
> Out of interest, are those windows failures common?

The Windows tests have been failing since September. I think it has to do with the use of `fs.mkdtemp()` `os.tmpdir()` in the [workspaces tests](https://github.com/raineorshine/npm-check-updates/blob/acec811d4b3188491df9804853df5215137762fe/test/workspaces.test.ts#L27-L28), but I can't for the life of me figure out how to get it to work on Windows. They work fine on OSX, which is what I am on.

I was hoping that a generous Windows user might notice the failing tests and poke around, but nothing so far.

PR #1238, opened 2022-12-07
chore(tests): tests the code that gathers package files and workspace names

doublethefish:
There is no intentional functional change in this PR, we just capture existing behaviour.

Ahead of fixing `doctor` for workspaces we need to make changes to the code that gathers the package files in workspaces, ideally without adding regressions.

This PR adds unit-tests for a chunk of complex code. The only change to that chunk of code is to move it into a testable function. The code is 100% covered on all lines and branches.

Note that these tests set the cwd inside a test, for now, because the existing code doesn't properly handle the `cwd` option.


raineorshine:
Looks like a nice improvement.

Following project conventions, please use `-` instead of `_` in filenames.

raineorshine (approved):

raineorshine:
Closed by mistake?

doublethefish:
yep.

raineorshine:
It won't let me reopen for some reason. Can you open a new PR?

doublethefish:
Oh, I thought it had reopened, hmm. Too much to do and too little time. I'll reopen it now.

PR #1241, opened 2022-12-09
chore(tests): tests the code that gathers package files and workspace names

doublethefish:
Reopen of #1238 that was accidentally closed.

There is no intentional functional change in this PR, we just capture existing behaviour.

Ahead of fixing doctor for workspaces we need to make changes to the code that gathers the package files in workspaces, ideally without adding regressions.

This PR adds unit-tests for a chunk of complex code. The only change to that chunk of code is to move it into a testable function. The code is 100% covered on all lines and branches.

Note that these tests set the cwd inside a test, for now, because the existing code doesn't properly handle the cwd option.


doublethefish:
There's currently a merge conflict that I don't have time to look at just now.

raineorshine:
Sorry, tests failing is my fault. The package-lock got out of sync. Re-installing now.

Issue #1240, opened 2022-12-09
--workspace option under windows seems to be not working


grbd:
- [ X ] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [ X ] I am using the latest version of `npm-check-updates`
- [ X ] I am using `node >= 14.14`

---

Hi,
As a result of https://github.com/raineorshine/npm-check-updates/issues/1239
I think I've discovered that the --workspace option doesn't work under windows.

`--workspaces` (selecting all of them from the root) seems to work fine.
But `--workspace packagename` seems to just exit without doing anything / without displaying any kind of error so I suspect it's not finding the workspace.

I tried running some tests against version v16.5.1 and the latest v16.5.3 to see if the recent changes had made a difference, but it seems to be the same regardless. I think the github workflows are only setup to test against ubuntu.

Running `npm install` `npm build` `npm test`

The tail end of the output is:
```
  staticRegistry
    ✔ latest


  341 passing (7m)
  3 pending
  7 failing

  1) --deep
       ignore stdin if --packageFile glob is specified:

      AssertionError: expected '1' to not equal '1'
      + expected - actual


      at Context.<anonymous> (test\deep.test.ts:84:51)

  2) --deep
       update multiple packages:

      AssertionError: expected '1' to not equal '1'
      + expected - actual


      at Context.<anonymous> (test\deep.test.ts:103:52)

  3) --workspace
       update single workspace with --workspace:
     AssertionError: expected {} to have property 'packages/a/package.json'
      at Context.<anonymous> (test\workspaces.test.ts:239:26)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  4) --workspace
       update single workspace with -w:
     AssertionError: expected {} to have property 'packages/a/package.json'
      at Context.<anonymous> (test\workspaces.test.ts:252:26)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  5) --workspace
       update more than one workspace:
     AssertionError: expected {} to have property 'packages/a/package.json'
      at Context.<anonymous> (test\workspaces.test.ts:267:26)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  6) --workspace and --root
       update root project and single workspace:
     AssertionError: expected { 'package.json': { …(2) } } to have property 'packages/a/package.json'
      at Context.<anonymous> (test\workspaces.test.ts:337:26)
[...]

PR #1245, opened 2022-12-09
Fix windows paths when specifying workspace

grbd:
Hi,
this is a fix for windows when specifying the --workspace option

related to https://github.com/raineorshine/npm-check-updates/issues/1240

### DIFF
--- a/src/lib/getAllPackages.ts
+++ b/src/lib/getAllPackages.ts
@@ -92,7 +92,9 @@ async function getAllPackages(options: Options): Promise<[string[], string[]]> {
               /* c8 ignore next */
               workspaces?.some(
                 workspacePattern =>
-                  pkgFile === path.join(cwd, path.dirname(workspacePattern), workspace, 'package.json'),
+                  // convert Windows path to *nix path for filtering
+                  pkgFile ===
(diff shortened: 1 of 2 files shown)


raineorshine:
Yay!!! Thanks so much! 🙏

I see that there is just one test failing now. Could you possibly look into that one as well?

```
  1) --deep
       update multiple packages:
     AssertionError: expected { …(3) } to have property 'C:\Users\RUNNER~1\AppData\Local\Temp\…'
      at Context.<anonymous> (test\deep.test.ts:111:24)
```

grbd:
Having a look at the moment, I think the --deep option is working ok
so it might be another path issue in the expected result

grbd:
I think the last commit should allow the tests to now pass

raineorshine (approved):

grbd:
Looks like a timeout issue on windows node16
not sure whats causing that unless maybe its related to the workflow

raineorshine:
Just a slow connection exceeding the 30s mocha timeout. I reran the tests and they all pass!

Thanks again, I'm happy to see being all green 🟢🟢🟢🟢🟢🟢🟢🟢🟢🟢

PR #1247, opened 2022-12-09
feat(use-pkg-metadata) use metadata structure for package info, describing work to be done

doublethefish:
This work makes it easier to reason about what work will be done by `runUpgrade` and `doctor`.

This is a step towards solid handling of package.json files.

This PR also:

* clarifies that all workspace names are returned in workspace mode, irrespective of which workspace is analysed.
* moves the runUpgrades() closure to a function, making its return types slightly easier to reason about.
* splits out the workspace gathering code into its own function, also making it easier to reason about.
* extends `PackageInfo` to include `name` and `filepath` properties which are used by `runUpgrade` and `doctor`.
* uses the `PackageInfo` type in `getPackages`, `run` and `runUpgrade`, sharing information and normalising how data about `package.json` files is accessed (reducing confusion, hopefully).


raineorshine (approved):
Looks good! Much better handling of the workspace complexity.

PR #1250, opened 2022-12-10
chore(fix post commit): fixes post commit hooks borking the developer machine

doublethefish:


.husky/post-commit
```diff
@@ -12,5 +12,11 @@ notify() {
 # ensure failed lint exit code passes through sed
 set -o pipefail
 
+branch=$(git branch --show-current)
+# Do NOT run this when rebasing or we can't get the branch
+if [ -z "$branch" ]; then
+  exit 0
+fi
```
raineorshine, line 19: Great, I was running into exactly this issue and neither `--no-verify` nor `HUSKY_SKIP_HOOKS` worked.

.husky/post-commit
```diff
@@ -12,5 +12,11 @@ notify() {
...
+
 # omit first 4 lines printed by npm run
-npm run lint:src 2>&1 | sed 1,4d || notify "Lint Error" &
+npm run lint:src 2>&1 | sed 1,4d || notify "Lint Error"
```
raineorshine, line 22: The `&` is actually intentional. I often perform several commits very quickly (such as several small amends), and I prefer not to be slowed down by running the linter synchronously. 

When the linter completes in the background, I get a notification and can then easily fix the linter error and amend the last commit.

raineorshine (changes requested):

.husky/post-commit
doublethefish: hmmm, ok I've pushed an alternative solution where the lint output skips the terminal entirely and as many errors as possible get shown in the notification.

Issue #1259, opened 2023-01-09
Doesn't allow multiple commands for doctorTest


Maxim-Mazurok:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [x] I am using the latest version of `npm-check-updates`
- [x] I am using `node >= 14.14`

---

## Steps to Reproduce

I'm trying to run this:
```
npm-check-updates -u --doctor --doctorTest "npx -p typescript -c 'npm run test --ignore-scripts && tsc --noEmit'"
```

As you can see, I'm using npx with single quotes in order to run multiple commands in one go.

This is a simplified version that also doesn't work: `--doctorTest "npm test && tsc`

I understand that `npm test && tsc` doesn't work because here:
https://github.com/raineorshine/npm-check-updates/blob/HEAD/src/lib/doctor.ts#L117
We're splitting by `" "` and using first array element as a program to spawn.

And `--doctorTest "npx -p typescript -c 'npm run test --ignore-scripts && tsc --noEmit'"` doesn't seem to work because when we're splitting - we're not treating stuff in single quotes as a single argument.

A workaround, of course, is to just have an npm script that runs two commands, but in multi-root monorepo I'm trying to avoid duplication of that sort.

## Current Behavior

Doesn't spawn as expected

## Expected Behavior

It would be great if there was an option to specify two or more commands.

Maxim-Mazurok:
Potential solution to extract arguments: https://regex101.com/r/2rzibO/1

Maxim-Mazurok:
I've fixed the issue by applying the following patch with `npx patch-package`:

```diff
diff --git a/node_modules/npm-check-updates/build/src/lib/doctor.js b/node_modules/npm-check-updates/build/src/lib/doctor.js
index 15dfe45..debb95a 100644
--- a/node_modules/npm-check-updates/build/src/lib/doctor.js
+++ b/node_modules/npm-check-updates/build/src/lib/doctor.js
@@ -116,7 +116,15 @@ const doctor = async (run, options) => {
             },
         };
         if (options.doctorTest) {
-            const [testCommand, ...testArgs] = options.doctorTest.split(' ');
+            const regexp = /(?:'([^']+)'|([^ ]+))/g;
+            const matches = options.doctorTest.matchAll(regexp);
+            const groups = [];
+                
+            for (const match of matches) {
+              groups.push(match[1] === undefined ? match[2] : match[1]); // either first or second group
+            }
[...]

PR #1260, opened 2023-01-12
Support doctorTest command that includes quotes

Maxim-Mazurok:
Related to #1259 but probably doesn't really close it, as it only acts as a workaround for running multiple commands

"global should run" test failed for me both with and without my changes (I didn't forget to rebuild)

### DIFF
--- a/src/lib/doctor.ts
+++ b/src/lib/doctor.ts
@@ -114,7 +114,14 @@ const doctor = async (run: Run, options: Options): Promise<void> => {
     }
 
     if (options.doctorTest) {
-      const [testCommand, ...testArgs] = options.doctorTest.split(' ')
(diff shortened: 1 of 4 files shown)


Maxim-Mazurok:
I'll check test failures on the CI and address the issues later

raineorshine:
Thanks!

I think it would be better to use only simple, built-in shell functions rather than invoking npx or npm in the test.

Maxim-Mazurok:
Agree, just couldn't think of a better way to test it on both Windows and Linux.
Now will be running a simple node script, which should work much faster and better.

Maxim-Mazurok:
Hopefully fixed failing tests

raineorshine (approved):

PR #1270, opened 2023-02-15
feat: allow switching from pre-releases back to latest

MKruschke:
Dear maintainers, 
we discovered a small limitation while we used ncu. We notice it while we tested an actual pre-release on our branch and later after we tested it we liked to switch back to `latest`. Unfortunately that wasn't working out of the box so we hope to enable this functionality by providing this PR.

### DIFF
--- a/src/lib/isUpgradeable.ts
+++ b/src/lib/isUpgradeable.ts
@@ -9,9 +9,10 @@ import { fixPseudoVersion, isComparable, isWildCard, stringify } from './version
  *
  * @param current
  * @param latest
+ * @param downgrade  Allow downgrading
  * @returns
  */
-function isUpgradeable(current: VersionSpec, latest: Version) {
+function isUpgradeable(current: VersionSpec, latest: Version, { downgrade }: { downgrade?: boolean } = {}): boolean {
   // do not upgrade non-npm version declarations (such as git tags)
   // do not upgrade wildcards
   if (!semver.validRange(current) || isWildCard(current)) {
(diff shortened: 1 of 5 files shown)


raineorshine:
Hi, thanks for proposing your contribution.

A design invariant that perhaps should be advertised better is that ncu will *never downgrade your dependencies*. This provides a guarantee to avoid version regressions, and is consistent with ncu's conservative approach by default. So unfortunately I won't be able to accept this change, although I appreciate your use case.

If it is just one or two dependencies that you are experimenting with, it may be preferable to downgrade manually, i.e. `npm install PACKAGE@latest`.

The only exception to the above rule is when switching tags. If you have a package on `v2.0.0` and there is a `v1.0.0-alpha.1` published to the `alpha` tag, then running `ncu --target @alpha` will recommend downgrading to `v1.0.0-alpha.1`. In this case, it is assumed that the explicit tag means the user understands what they are doing, and the ability to upgrade or downgrade as needed makes it easier to work with projects that use different tags for different releases. For a full discussion, see #1134. Unfortunately, `--target @latest` is ignored (since `latest`) is the default), so we do not get the downgrade behavior for latest. However, I would be open to allowing downgrades when `--target @latest` is explicit.

MKruschke:
Hi Raine,
I understand the reasoning and adapted to the proposed behaviour of yours (`--target @latest`). Please let me know if the change is fine or need some more love.

raineorshine:
Looks good. I just need to check how non-latest tags handle this and see if there is an opportunity to unify the logic.

Issue #1282, opened 2023-03-20
Feature request: --format for a machine-readable, line-base output


vanodevium:
It would be nice to support next formats:
- lines
- shell
- comma

--- 
It will be useful for CLI cases like a
```
sudo npm i -g $(ncu -g --format shell)
```

raineorshine:
Published in `v16.8.0`.

PR #1283, opened 2023-03-21
options.format: lines (output packages line by line)

vanodevium:
#1282 

@raineorshine I have some problems with TS :(

Anyway, I'm trying. 

Is this correct implementation? Can I continue work with tests and readme?

### DIFF
--- a/src/lib/initOptions.ts
+++ b/src/lib/initOptions.ts
@@ -64,11 +64,15 @@ async function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } = 
   }
 
   // consolidate loglevel
-  const loglevel = options.silent ? 'silent' : options.verbose ? 'verbose' : options.loglevel
+  let loglevel = options.silent ? 'silent' : options.verbose ? 'verbose' : options.loglevel
 
   const json = Object.keys(options)
     .filter(option => option.startsWith('json'))
     .some(propertyOf(options))
+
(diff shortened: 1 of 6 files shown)


src/lib/initOptions.ts
```diff
@@ -64,12 +64,16 @@ async function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } =
...
-  const loglevel = options.silent ? 'silent' : options.verbose ? 'verbose' : options.loglevel
+  let loglevel = options.silent ? 'silent' : options.verbose ? 'verbose' : options.loglevel
 
   const json = Object.keys(options)
     .filter(option => option.startsWith('json'))
     .some(propertyOf(options))
 
+  if (options.format?.includes('shell') || options.format?.includes('lines') || options.format?.includes('comma')) {
+    loglevel = 'silent';
+  }
```
raineorshine, line 75: You can do this logic in the ternary and avoid changing `const` to `let`.

raineorshine (changes requested):

raineorshine:
`--format unix` is ambiguous in its own way, and I don't want to imply that this feature is platform specific and can't be used by Windows users. `--format lines` describes the output: one update per line, without anything else. The help system can clarify further.

vanodevium:
OK. I have done my best.

vanodevium:
Thanx!

Issue #1286, opened 2023-03-27
Update only major versions


mslowiak:
Hi!
I was playing around with npm-check-updates and did not found an option to update only major versions (excluding potential updates for minor/patch libraries).

**Why I need this?**
Updating deps update is tedious task for each I have the Jenkins job.
For the minor/patch updates, I would like to deploy update dependencies automatically (automated work)
For the major updates, I want to review the changes before I deploy new application version (manual work) 

**Example:**
application package.json contains following libraries:
- a - 1.2.0
- b - 13.0.0
- c - 5.0.1

There newest versions in artifactory:
- a - 1.3.0
- b - 14.1.1
- c - 5.0.3

**Expectation:**
Only version of artifact `b` should be updated from 13.0.0 to 14.1.1 because that is the major change in the example.
Currently there is a target option, but my use case is not supported with any of options there. 

**Workaround:**
1. Execute `npx npm-check-updates --format group`
2. Parse the output from point 1 and get arfifacts names from `Major` section
3. Run `npm-check-updates` with  regex based on artifacts found in point 2

Maybe there can be added new target option called `majorOnly` that will work like I described?

mslowiak:
> --target minor and --target patch, which appear to be closest functionally, actually match against all published versions. I assume you want to stick to the version published to the latest tag.

I am not sure if I understand correctly but for my use case I am interested only in major versions.
The `--target latest` returns majors, minors, and patches.

> FYI If you use ^ ranges, then you could use --minimal to achieve your desired behavior.
That would be an option, but taking into account multiple repository changes it's not feasible.

You mentioned the `filterResults`. Is the work regarding that thing already started?
If there is any option to help develop something in this direction, let me know 👍

mslowiak:
That would be great if you could put some light on the code, maybe that will help to get it done faster 👍

raineorshine:
The new `filterResults` option can be added to https://github.com/raineorshine/npm-check-updates/blob/main/src/cli-options.ts. This makes `options.filterResults` available throughout the codebase.

I would recommend the following signature for the use-provided predicate function:

```
export type FilterResultsFunction = (packageName: string, { currentVersion: VersionSpec, currentVersionSemver: SemVer[], upgradedVersion: Version }) => boolean
```

(I chose this signature to be somewhat future-proof; removing/changing arguments would require a major version change, but adding keys to the argument object would not.)

You will need to pass the appropriate arguments to the user-provided `filterResults` function:

[...]

PR #1290, opened 2023-04-01
feat(filterResults): add new option to filter results of update

mslowiak:
Closes: #1286


mslowiak:
Hi @raineorshine 👋 

This PR should cover the filterResults option.
I added `upgradedVersionSemver` to the `FilterResultsFunction` type as I feel it is easier for user to extract sem ver parts from version easier with that.

src/cli-options.ts
```diff
@@ -146,6 +146,31 @@ Example:
...
+const extendedHelpFilterResults = (): string => {
+  return `Filters the results of update based on user provided function. Only available in .ncurc.js or when importing npm-check-updates as a module:
+
+    ${chalk.gray(`/**
+      @param packageName                   The name of the dependency.
+      @param currentVersion                Current version declaration (may be range)
+      @param currentVersionSemver          Current version declaration in semantic versioning format (may be range)
+      @param upgradedVersion               Upgraded version declaration (may be range)
+      @param upgradedVersionSemver         Upgraded version declaration in semantic versioning format (may be range)
+      @returns                             A result in boolean format - true or false.
```
raineorshine, line 159: "Return true if the upgrade should be kept, otherwise it will be ignored."

src/cli-options.ts
```diff
@@ -146,6 +146,31 @@ Example:
...
+    */`)}
+    ${chalk.cyan(
+      'filterResults',
+    )}: (packageName, {currentVersion, currentVersionSemver, upgradedVersion, upgradedVersionSemver}) {
+      const currentMajorVersion = currentVersionSemver && currentVersionSemver[0] && currentVersionSemver[0].major
+      const upgradedMajorVersion = upgradedVersionSemver && upgradedVersionSemver[0] && upgradedVersionSemver[0].major
```
raineorshine, line 165: Please use the [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator to make this more concise.

src/cli-options.ts
```diff
@@ -146,6 +146,31 @@ Example:
...
+      if (currentMajorVersion && upgradedMajorVersion) {
+        return currentMajorVersion < upgradedMajorVersion
+      }
+      return true
```
raineorshine, line 169: Let's provide a comment that explains what this does when used as the filterResults function.

src/lib/upgradePackageDefinitions.ts
```diff
@@ -2,15 +2,37 @@ import isEmpty from 'lodash/isEmpty'
...
+import { Version } from '../types/Version'
 import { VersionResult } from '../types/VersionResult'
 import { VersionSpec } from '../types/VersionSpec'
 import getPeerDependenciesFromRegistry from './getPeerDependenciesFromRegistry'
 import keyValueBy from './keyValueBy'
 import queryVersions from './queryVersions'
 import upgradeDependencies from './upgradeDependencies'
 
+/**
+ *
```
raineorshine, line 17: Please add a descriptive JSDOCS comment.

src/lib/upgradePackageDefinitions.ts
```diff
@@ -2,15 +2,37 @@ import isEmpty from 'lodash/isEmpty'
...
+  version: Version,
+  options: Options,
+) {
+  return (
+    !options.filterResults ||
+    options.filterResults(dependencyName, {
+      currentVersion: currentDependencies[dependencyName],
+      currentVersionSemver: parseRange(currentDependencies[dependencyName]),
+      upgradedVersion: version,
+      upgradedVersionSemver: parseRange(version),
```
raineorshine, line 31: You can should use `parse` here since the upgraded version is not a range.

src/types/FilterResultsFunction.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Version } from './Version'
+import { VersionSpec } from './VersionSpec'
+
+export type FilterResultsFunction = (
+  packageName: string,
+  versioningMetadata: {
+    currentVersion: VersionSpec
+    currentVersionSemver: SemVer[]
+    upgradedVersion: Version
+    upgradedVersionSemver: SemVer[]
```
raineorshine, line 11: `SemVer`

test/filterResults.test.ts
```diff
@@ -0,0 +1,79 @@
...
+chai.should()
+chai.use(chaiAsPromised)
+chai.use(chaiString)
+
+const bin = path.join(__dirname, '../build/src/bin/cli.js')
+
+/**
+ * Sets up and tears down the temporary directories required to run each test
+ */
+async function filterResultsTestScaffold(
```
raineorshine, line 20: I prefer redundancy over abstraction in tests, so let's avoid a separate test scaffold function. You can do everything inline in the test case.

test/filterResults.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    JSON.stringify({
+      dependencies,
+    }),
+    'utf-8',
+  )
+  const configFile = path.join(tempDir, '.ncurc.js')
+  await fs.writeFile(configFile, `module.exports = { filterResults: ${filterResultsFn.toString()} }`, 'utf-8')
+
+  try {
+    const stdout = await spawn('node', [bin, '--configFilePath', tempDir], {
```
raineorshine, line 50: This is fine to run npm-check-updates in a child process, but note that this is a bit more verbose and less efficient than importing npm-check-updates as a module.

The only tests that run npm-check-updates as a child process are those that need to test functionality in the cli that can't be tested when importing it as a module.

test/filterResults.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    {
+      'ncu-test-v2': '3.0.0',
+      'ncu-test-tag': '2.1.0',
+      'ncu-test-return-version': '1.2.0',
+    },
+    { spawn: true },
+  )
+
+  // use dynamic import for ESM module
+  const { default: stripAnsi } = await import('strip-ansi')
```
raineorshine, line 36: If you add `jsonUpgraded` to the options, you can just `JSON.parse` the output and avoid having to strip ansi.

raineorshine (changes requested):
Thanks! Looks great. Just some minor changes requested.

src/cli-options.ts
```diff
@@ -146,6 +146,31 @@ Example:
     Saving partially upgraded package.json
 `
 
+/** Extended help for the filterResults option. */
+const extendedHelpFilterResults = (): string => {
+  return `Filters the results of update based on user provided function. Only available in .ncurc.js or when importing npm-check-updates as a module:
+
+    ${chalk.gray(`/**
+      @param packageName                   The name of the dependency.
```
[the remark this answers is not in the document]
mslowiak: Please take a look if I used the param definition and link correctly (I am not js dev, so had no experience with it).

src/cli-options.ts
```diff
@@ -146,6 +146,31 @@ Example:
...
+      @param currentVersion                Current version declaration (may be range)
+      @param currentVersionSemver          Current version declaration in semantic versioning format (may be range)
+      @param upgradedVersion               Upgraded version declaration (may be range)
+      @param upgradedVersionSemver         Upgraded version declaration in semantic versioning format (may be range)
```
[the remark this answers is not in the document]
mslowiak: fixed 🚀

src/cli-options.ts
mslowiak: fixed 🚀

src/lib/upgradePackageDefinitions.ts
mslowiak: fixed 🚀

test/filterResults.test.ts
mslowiak: fixed 🚀

test/filterResults.test.ts
mslowiak: I think I got what you described here, but please check this part ;)

src/types/FilterResultsFunction.ts
mslowiak: fixed 🚀

src/cli-options.ts
mslowiak: fixed 🚀

src/cli-options.ts
mslowiak: fixed 🚀

src/lib/upgradePackageDefinitions.ts
mslowiak: fixed 🚀

mslowiak:
@raineorshine 
I have made fixes according to your comments 👍 
Please do check the code for two that I left unresolved 🥺 
Thanks for solid review here!

raineorshine (approved):

raineorshine:
Perfect! Thanks for your contribution!

tvdijen:
I'm struggling to get this to work.. Whatever I try I get "TypeError: options.filterResults is not a function".. Could I get a working example of a `.ncurc.js` file?

raineorshine:
@tvdijen Here's the example from the README:

```
module.exports = {
  /** Filter out non-major version updates.
    @param {string} packageName               The name of the dependency.
    @param {string} currentVersion            Current version declaration (may be range).
    @param {SemVer[]} currentVersionSemver    Current version declaration in semantic versioning format (may be range).
    @param {string} upgradedVersion           Upgraded version.
    @param {SemVer} upgradedVersionSemver     Upgraded version in semantic versioning format.
    @returns {boolean}                        Return true if the upgrade should be kept, otherwise it will be ignored.
  */
  filterResults: (packageName, { currentVersion, currentVersionSemver, upgradedVersion, upgradedVersionSemver }) => {
    const currentMajorVersion = currentVersionSemver?.[0]?.major
    const upgradedMajorVersion = upgradedVersionSemver?.major
    if (currentMajorVersion && upgradedMajorVersion) {
      return currentMajorVersion < upgradedMajorVersion
    }
    return true
  }
}
```

https://github.com/raineorshine/npm-check-updates#filterresults

tvdijen:
Yeah, I tried to copy that into the file, but it's still throwing me the exact same error...
Am I using it the wrong way?  `npx npm-check-updates --filterResults=filterResults`

raineorshine:
npm-check-updates automatically reads from the `.ncurc.js` file in the project or user folder, so you don't need to specify `--filterResults` on the command line.

(In fact, it is not possible to use `--filterResults` on the command line because it requires defining a function in javascript. Options that take string, number, or boolean values can be provided on the command line.)

tvdijen:
Ah, I'm sorry, I'm starting to understand.. It _is_ working, it's just not filtering things the way I expected it to.

```
  filterResults: (packageName, { currentVersion, currentVersionSemver, upgradedVersion, upgradedVersionSemver }) => {
    const currentMajorVersion = currentVersionSemver?.[0]?.major
    const upgradedMajorVersion = upgradedVersionSemver?.major
    console.log(packageName, "; Current: ", currentMajorVersion, "; Upgrade: ", upgradedMajorVersion);
    if (upgradedMajorVersion > currentMajorVersion) {
      return false
    }
    return true
  }
```

This prints: `copy-webpack-plugin ; Current:  9 ; Upgrade:  11`
But the summary says `copy-webpack-plugin                 ^9.1.0  →  ^11.0.0`

If I move the `console.log` inside the if-statement it won't print, so for some reason the condition is not met even though 11 > 9  :/

raineorshine:
Good catch, I'll update the example to bring attention to this gotcha.

tvdijen:
Well, I'm not sure if this is the real solution since according to the doc-block the `currentVersion` can be a range, so in this case the `parseInt()` would return `NaN`.

raineorshine:
The `[0]` grabs the first version in the range, so it parseInt should work fine. For example, if the current version specified your your package.json is `>= 1.1.7 < 2.0.0`, then `currentVersionSemver?.[0]?.major` would equal `'1'`.

PR #1309, opened 2023-07-22
feat: bun as packageManager option

ImBIOS:
@Jarred-Sumner @oven-sh need your help and review


raineorshine:
Hi, thanks for the contribution. This looks good to me. Just need to get the tests passing.

raineorshine:
The minimum packageManager file just needs a function to return the latest version:

```js
{
  latest: (packageName: string,
  currentVersion: string) =>
    { error?: string, version?: number }
}
``` 

Additional targets (patch, minor, newest, greatest) can be added by defining their respective functions, which have the same signature. You may also consider defining a `list` function to add support for global mode. 

The full `PackageManager` type is here: https://github.com/raineorshine/npm-check-updates/blob/main/src/types/PackageManager.ts

ImBIOS:
@raineorshine I just implemented the basic bun spawner

src/package-managers/bun.ts
```diff
@@ -0,0 +1,42 @@
+import spawn from 'spawn-please'
+import { Index } from '../types/IndexType'
+import { NpmOptions } from '../types/NpmOptions'
+import { latest as npmLatest, list as npmList } from './npm'
+
+/**
+ * (Bun) Fetches the list of all installed packages.
+ */
+export const list = npmList
```
raineorshine, line 9: You should spawn bun for the list command, as npm may have a different global install directory than bun.

Hopefully bun has a `--json` option, otherwise you'll have to parse the results.

src/package-managers/bun.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+async function spawnBun(
+  args: string | string[],
+  bunOptions: NpmOptions = {},
+  spawnOptions: Index<any> = {},
+): Promise<any> {
+  // Bun not yet supported on Windows.
+  // @see https://github.com/oven-sh/bun/issues/43
+  if (process.platform === 'win32') {
+    throw new Error('Bun not yet supported on Windows')
```
raineorshine, line 32: Please use https://github.com/raineorshine/npm-check-updates/blob/main/src/lib/programError.ts, which has different error behavior for CLI mode and programmatic mode.

test/package-managers/bun/index.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+chai.should()
+chai.use(chaiAsPromised)
+chai.use(chaiString)
+
+process.env.NCU_TESTS = 'true'
+
+describe('bun', function () {
+  it('list', async () => {
+    const versionObject = await bun.list({ cwd: __dirname })
```
raineorshine, line 14: Though there are some old tests remaining that operate on test files in the project directory tree, I prefer a dynamic setup in a temp directory as seen here:

https://github.com/raineorshine/npm-check-updates/blob/c7a339c53a09caf1d6ed28f6e6e3c6e25eb26fec/test/bin.test.ts#L115-L128

It makes the tests slightly more verbose, but I've found it easier to manage by having everything related to a test contained in the test file. Plus cleanup is easier because you can just delete the temp directory rather than having to do a precise restore.

raineorshine (changes requested):

raineorshine:
Hi, have you had a chance to review the requested changes?

Just wanted to check in since some merge conflicts are accumulating as this sits.

ImBIOS:
@raineorshine I've fixed the conflict and delivered 2 change requests. But, I have no idea how to deliver the other 2 requests you made. Can I request for your help?

raineorshine:
I implemented the `list` command which gets the list of installed bun dependencies. The bun cli does not have a `--json` option so it parses the cli line output. Please see 15775ad to see how I did it.

Basic doctor mode works now. Also, I installed bun conditionally in the CI environment and have it skip the tests on Windows.

I think we can merge this as an initial implementation. Additional functionality can be added in separate PR's. Let me know if there's anything else you'd like to add before I merge. Thanks!

raineorshine (approved):

ImBIOS:
@raineorshine Nah, All good. Looks Good To Me. Thanks a lot for your help adding bun 🙏 😄

PR #1326, opened 2023-09-05
fix #1325 - setting errorLevel in workspaces doesn't work as expected

### LINKED ISSUE #1325: setting errorLevel in workspaces doesn't work as expected

---

## Steps to Reproduce

Steps:

- Create a mono repo, with 2 workspaces that have upgrades available to them (each)
- Run in the root of the monorepo
```bash
yarn ncu --cache --workspaces --root --enginesNode --target minor --upgrade --errorLevel 2
```

## Current Behavior

The script will exit with error code, after it upgrades the first workspace, without upgrading the 2nd workspace

## Expected Behavior

If no actual error occurred, need to run all workspaces, and decide on exist status at the end - if there are upgrades needed in any of the workspace, error, otherwise, not.
This allow to conditional run commands, if *any* of the workspaces requires an update

regevbr:
fix #1325

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -72,6 +72,15 @@ const getPackageManagerForInstall = async (options: Options, pkgFile: string) =>
   return pnpmDetected ? 'pnpm' : 'npm'
 }
 
+/** Returns if analysis contains upgrades */
+const someUpgraded = (pkgs: string[], analysis: Index<PackageFile> | PackageFile) => {
+  // deep mode analysis is of type Index<PackageFile>
+  // non-deep mode analysis is of type <PackageFile>, so we normalize it to Index<PackageFile>
+  const analysisNormalized: Index<PackageFile> =
+    pkgs.length === 1 ? { [pkgs[0]]: analysis as PackageFile } : (analysis as Index<PackageFile>)
+  return Object.values(analysisNormalized).some(upgrades => Object.keys(upgrades).length > 0)
+}
+
 /** Either suggest an install command based on the package manager, or in interactive mode, prompt to auto-install. */
 const npmInstall = async (
   pkgs: string[],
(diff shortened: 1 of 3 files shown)


regevbr:
@raineorshine I would appreciate your CR

src/index.ts
```diff
@@ -219,12 +223,15 @@ async function runUpgrades(options: Options, timeout?: NodeJS.Timeout): Promise<
...
               // convert Windows path to *nix path for consistency
               .replace(/\\/g, '/')
-          : pkgFile!]: await runLocal(pkgOptions, pkgData, pkgFile),
+          : pkgFile!]: runResult,
       }
     }, Promise.resolve({} as Index<PackageFile> | PackageFile))
     if (options.json) {
       printJson(options, analysis)
     }
+    if (totalUpgradedCount > 0 && options.errorLevel === 2) {
```
raineorshine, line 232: I think we can just check if the `analysis` is empty. Then we don't have to alter the return type or add other variables. 

See the `someUpgraded` variable in the `npmInstall` function.

raineorshine (changes requested):
Thanks for the PR!

src/index.ts
regevbr: Ok perfect changed it - it is a much smaller change now. I wasn't too sure about the structure of the response

src/index.ts
```diff
@@ -209,6 +209,7 @@ async function runUpgrades(options: Options, timeout?: NodeJS.Timeout): Promise<
         ...rcConfig,
         packageFile: packageInfo.filepath,
         workspacePackages,
+        errorLevel: 1,
```
raineorshine, line 212: Why set `errorLevel` here?

raineorshine (commented):
Thanks! That looks better. Just one question for you.

src/index.ts
regevbr: if we do not, it will throw if there are upgrades, which is the current behaviour we are trying to fix

raineorshine: Okay, thanks. It seems wrong that we have to hardcode this here when the default value of `errorLevel` is already 1. Maybe we can consolidate the workspace and non-workspace logic to avoid this? 

I will need to take a closer look at how it's handled to know. I'm not available as much right since I'm traveling, but I can check it out soon.

regevbr: the default is indeed 1, but in my use case, we pass 2, thus causing it to exist on the first workspace that needs updating. This will ensure that it will not error on first workspace.
If you want, we can instead pass a flag to runLocal that instructs to not exit on upgrades needed. What do you think?

raineorshine: If we need to override the  errorLevel passed to `runLocal`, then the exit logic is probably in the wrong place. Without having looked at the code yet, I'm thinking we need to return early from `runLocal` but only exit the process at the end.

regevbr: I think I see your point - I will make the change

regevbr: Done - cleaner code now

raineorshine (approved):
💯

PR #1327, opened 2023-09-12
Update filter type to allow mixed string and RegExp arrays

stephenwade:
This PR updates the FilterPattern type to accept mixed string and RegExp arrays. This allows the user to pass an array containing both strings and regular expressions when using npm-check-updates programmatically.

This PR only changes types. It doesn't change any runtime behavior.


raineorshine:
Thanks!

There was one small change of runtime behavior I needed to make (3aaf995). Luckily the unit test I wrote picked it up immediately.

raineorshine (approved):

Issue #961, opened 2021-10-26
Batch npm fetch


make-github-pseudonymous-again:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)
- [x] I am using the latest version of `npm-check-updates`
- [x] I am using `node >= 10.17`

---------------------------

## Steps to Reproduce

Run `ncu` with >100 dependencies and a slow internet connection.

## Current Behavior

It's slow.

## Expected Behavior

It's fast.

## Suggested solution

Ask `npm` to expose a route where one can query for more than one package at a time. It could even explicitly be an *updates available* route that only replies with a list of updates given current version, not with a list of last available versions. They will have to put limits on this route like maximum payload (`ncu` resolves wildcards locally), but on average I think it would lighten the load on their side and make it faster for us.

make-github-pseudonymous-again:
Hi @zkat, @isaacs, and  @wraithgar! Any idea who to contact to discuss a potential addition to npm's API?

wraithgar:
Changes to the npm registry itself would have to go through [support](https://www.npmjs.com/support). The npm cli interacts with that and many other registries (i.e. github packages registry, private registries, azure artifacts). If that route were added to any of those registries the cli could then be updated to use it.

make-github-pseudonymous-again:
Thanks @wraithgar.

@raineorshine Shall I open a ticket?

make-github-pseudonymous-again:
Hi @raineorshine! I would like to open a npm support ticket to request this API addition we would need. Do you give me permission to explicitly mention in this ticket the reason for this request is to improve the running time of `npm-check-updates`?

raineorshine:
Yes, go for it!

make-github-pseudonymous-again:
Here is what I sent:

> New API route for batched version update information fetching
>
> Dear support employee,
> 
> A tool I use to keep my npm packages (direct) dependencies up to date is
> relying on many "parallel" API queries to fetch relevant information. Currently
> this process running time scales quite badly with the number of dependencies.
> There are several ways this could be improved, and some involve having a better
> way to query the npm registry for information. For instance:
> 
>   - The tool currently makes one query per (direct) dependency. I think that
>     being able to query the registry for a reasonable amount of packages in a
>     single batch would be significantly more efficient.
>   - The tool currently uses `pacote.packument` which does not seem to allow to
>     query for only a subset of the fields. I think that being able to specify
[...]

PR #1328, opened 2023-09-13
Speed up requests with dist-tags endpoint

raineorshine:
Use the `/dist-tags` registry endpoint for fetching the latest version instead of the full packument.

e.g. https://registry.npmjs.org/-/package/npm-check-updates/dist-tags

Initial exploration of #961.


raineorshine:
Hit a block when I realized that the `deprecated` status is not available on that endpoint. npm-check-updates excludes deprecated versions by default, even if they are published to latest. Unfortunately this means the `/dist-tags` endpoint is insufficient.

PR #1329, opened 2023-09-13
Replace pacote with npm-registry-fetch

raineorshine:
Replace `pacote` with `npm-registry-fetch` to reduce the bundle size.

There is some risk that the request headers are not recreated accurately, so this should wait for a major version change to be safe. Consider releasing in a separate ~v17.0.1~ prerelease v17.0.0-1. This way the risk is the same but we have a checkpoint that can be tested separately from the bulk of v17.


raineorshine:
Rebased on [v17-next](https://github.com/raineorshine/npm-check-updates/tree/v17-next)

wmertens:
Once this merges, the node-gyp dep can go

PR #1344, opened 2023-11-03
fix: bad eslint-plugin-jsdoc version

Zamiell:
the current version of eslint-plugin-jsdoc prevent the repo from installing on node 20 (i.e. the current LTS), so I updated the dep to latest and now it works.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,7 +85,7 @@
         "eslint-config-standard": "^17.1.0",
         "eslint-plugin-fp": "^2.3.0",
         "eslint-plugin-import": "^2.28.1",
-        "eslint-plugin-jsdoc": "^41.1.2",
+        "eslint-plugin-jsdoc": "^46.8.2",
         "eslint-plugin-n": "^15.7.0",
         "eslint-plugin-promise": "^6.1.1",
         "husky": "^8.0.3",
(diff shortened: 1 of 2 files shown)


Zamiell:
eslint-plugin-jsdoc is a devDep, so it shouldn't be a breaking change for anybody, right?

raineorshine:
Hmmm. There is only one node engines field, so I don't think there is a way to have separate platform requirements for the devDependencies that I am aware of.

It would probably be best for you to use node v14–18 to fix the Windows bug until the node engine is bumped. 

Or we could focus on getting npm-check-updates v17 out the door since it is pretty close. It just needs to be rebased on main and then address a few breaking dependency changes.

Zamiell:
> There is only one node engines field, so I don't think there is a way to have separate platform requirements for the devDependencies that I am aware of.

oh, ok, I wasn't aware of that.

> Or we could focus on getting npm-check-updates v17 out the door since it is pretty close. It just needs to be rebased on main and then address a few breaking dependency changes.

ok, I'll close this PR in the meantime

Zamiell:
All the tests seem to be passing on the v17 branch, can you clarify exactly what needs to be done?

raineorshine:
Sure, I've created a discussion: #1346

PR #1345, opened 2023-11-03
v17

Zamiell:
Intro and discussion: https://github.com/raineorshine/npm-check-updates/discussions/1346

PR #1349, opened 2023-11-05
v17-next

raineorshine:
`v17` fell behind, so I rebased it on `main` and pushed it to `v17-next`. This should serve as the basis for the v17 release.

This was done to avoid force pushing `v17`.


raineorshine:
A lot of upgrades were made, but I broke something along the way. I'll take a closer look soon.

raineorshine:
Merged into #1345.

PR #1368, opened 2024-02-09
Support `semver` target with `gitTags` package manager

piranna:
Using `semver` target exit with error on git dependencies. This PR add support for it, by doing nothing since semver is not possible to enforce with git dependencies (would they be practical at all?), so no doesn't exit with error, just they are being ignored, the same as if they were explicit versions.

### DIFF
--- a/src/package-managers/gitTags.ts
+++ b/src/package-managers/gitTags.ts
@@ -1,7 +1,7 @@
 /** Fetches package metadata from Github tags. */
 import parseGithubUrl from 'parse-github-url'
 import remoteGitTags from 'remote-git-tags'
-import semver from 'semver'
+import { valid } from 'semver'
 import { print } from '../lib/logging'
 import * as versionUtil from '../lib/version-util'
 import { GetVersion } from '../types/GetVersion'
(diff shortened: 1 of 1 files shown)


raineorshine:
I agree we should add a `semver` implementation for git tags, though I don't think ignoring them is quite right. We already parse a semver-like version number from the tag name and use that in `latest`, `greatest`, `minor`, and `patch`. Non semver-like tags are ignored. `semver` can be implemented basically the same as in /package-managers/npm with [`const version = nodeSemver.maxSatisfying(versionsFiltered, currentVersion)`](https://github.com/raineorshine/npm-check-updates/blob/f873346504fae04645abc72a3bf3b1962d91c531/src/package-managers/npm.ts#L877).

piranna:
My first impression was to try to do a proper implementation, but found a problem: there's no way to add caret (`^`) or tilde (`~`) anywhere in the version when using git dependencies to define the acceptable version range level, so if you define a tag or a commit or a branch, it's almost the same as using a fixed dependency version or (according to semver spec) using a prerelease: you want to use THAT specific version / commit / tag / branch HEAD. That's why I decided to ignore them. I'm thinking on a way to properly implement semver, but I think that not only it's not possible to define a range in a git dependency, but also that it doesn't makes sense at all.

raineorshine:
Ah, I see. So all git tags are exact versions. Now I understand!

piranna:
> Ah, I see. So all git tags are exact versions. Now I understand!

Yes, git tags are "alias" of a specific commit, the same as using a branch name is an "alias" to the commit at the HEAD of that branch. So yes... they works as exact versions :-)

(Ok, tags can be reassigned to another commit, but it's not so common, and by definition, the HEAD of a branch is a moving target, because pushing a new commit change the HEAD of the branch, but from the point of view of the git dependency, they are exact versions, no matter if the content can change, so it's not possible to define ranges.)

piranna:
Cool, thanks for merging :-)

raineorshine:
Published to `v16.14.15`

PR #1372, opened 2024-03-07
Replace rimraf with built-in fs.rm

Torathion:
Hi, your most recent issues caught my attention and I wanted to contribute to this amazing project.

I noticed that you are already using `fs.rm` but kept using rimraf, so I fully replaced it with the node.js built-in. I would highly recommend using the built-in as `rimraf` is not a small package (277 KB) and you are still using the deprecated `@types/rimraf` package. 

Due to the frequent use of the same options object in the tests, I optimized it a bit, made it into a constant and placed it into a new file called `testContants.ts` so the test will run a tiny bit faster. It's just a micro-optimization and to make the code more readable.


raineorshine:
Hi! Thanks for your contribution. 

That would be great to remove rimraf and replace it with the built-in. Thank you. 

I actually follow a test design strategy that avoids abstractions in tests, preferring inline configurations when possible. This is unlike the application code in which redundancy is emphatically avoided. This approach makes the tests more self-contained and, in more complex setups, reduces the risk of bugs.

Torathion:
Ah, got it. That's why the CI also failed. I'll do it ASAP.

its-dibo:
is this fix included in `v16.14.15` yet?

raineorshine:
No, this is planned for v17.

PR #1373, opened 2024-03-08
Replace globby with fast-glob

Torathion:
I realized that [globby](https://www.npmjs.com/package/globby) uses [fast-glob](https://www.npmjs.com/package/fast-glob) under the hood. So, globby is just a wrapper package with some extra options.

Also, I micro-optimized two things:
- I made the options into a constant, so it won't rebuild the options object every time
- You unnecessarily copied the output of one glob function.

The other change is just prettier being prettier.


Torathion:
Oh, I haven't updated the package-lock.json. Uno momento

raineorshine (approved):
Looks good, thanks!

PR #1374, opened 2024-03-08
Replace minimatch with the faster alternative picomatch

Torathion:
Replaces [minimatch](https://www.npmjs.com/package/minimatch) with the smaller (434 KB -> 85 KB) and [faster](https://www.npmjs.com/package/picomatch#Benchmarks) alternative [picomatch](https://www.npmjs.com/package/picomatch).

Picomatch is also the faster alternative to micromatch and nanomatch that aren't maintained anymore. Picomatch uses the factory pattern to create Matchers to be used to match strings:

```ts
import pm from 'picomatch'

const matcher = pm('*.js')

console.log(matcher('a.js')) // true
console.log(matcher('b.ts')) // false
```

Written inline, it looks like this:

```ts
import picomatch from 'picomatch'

console.log(picomatch('*.js')('a.js')) // true
```

Some other things:

- You should refactor the `composeFilter` function, because every inline function, such as `predicate`, `matchUnscoped` and `matchScoped` are completely rebuild every time you call `composeFilter`. This pattern is also the reason why nested `forEach`-loops are incredibly slow, due to the constant function rebuilding.
- I replaced `filterPattern[filterPattern.length -1]` with `filterPattern.at(-1)` which is a [new prototype function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/at) that is twice as fast. You should integrate it into your project or use [eslint-plugin-unicorn](https://www.npmjs.com/package/eslint-plugin-unicorn) to fix all occurrences of it.


raineorshine (approved):
Thanks, a nice improvement. I didn't know about `.at`. 

I'll keep the compare function optimization in mind as a future area of improvement.

PR #1375, opened 2024-03-08
Replace get-stdin with built-in stream consumer

Torathion:
Since Node.js 16.7.0, node introduced stream consumers and with that the ability to parse a stream to [text](https://nodejs.org/api/webstreams.html#streamconsumerstextstream). I found out about this as Sindre updated his [README](https://github.com/sindresorhus/get-stdin?tab=readme-ov-file#tip) of get-stdin on his github, but not on NPM. This explains why he hasn't maintained the package since then.


raineorshine (approved):
Great!

PR #1383, opened 2024-03-21
Fix circular peer dependencies blocking updates

### LINKED ISSUE #1382: Unable to upgrade packages which are circular peer dependencies

---

## Steps to Reproduce

Use the following `package.json`, run `npm install` then `npm run test`.

```json
{
  "name": "npm-check-updates-circular-peer",
  "scripts": {
    "test": "npm-check-updates --peer"
  },
  "dependencies": {
    "@vitest/ui": "^1.3.1",
    "npm-check-updates": "^16.14.17",
    "vitest": "^1.3.1"
  }
}
```

The output will be the following:

```

 @vitest/ui  ^1.3.1  →  ^1.4.0  reason: vitest requires 1.3.1
 vitest      ^1.3.1  →  ^1.4.0  reason: @vitest/ui requires 1.3.1
```

Both of these vitest packages depend on each other (eg: their versions must be kept in sync). This makes it impossible to update one without the other, both need to be updated together.

It'd be great if the peer dependency check could see if the version that it wants to update package A to would then no longer cause a conflict with package B.

It seems like currently what is happening is:

* Get the latest versions of each package
* Check if the latest version of a package conflicts with the current version of every other package

I'm suggesting:

* Get the latest versions of each package
* Check if the latest version of a package conflicts with the proposed (new) version of every other package

CreativeTechGuy:
Fixes #1382 

This also fixes an issue (no clue how it never caused an error) where calling `packageManager.getPeerDependencies` with a version range (eg: `^1.2.3`) would cause the function to return an array instead of the expected object.


src/lib/isCircularPeer.ts
```diff
@@ -0,0 +1,38 @@
+import { Index } from '../types/IndexType'
```
raineorshine, line 1: Let's move all of the `isCircularPeer` code into getPeerDependenciesFromRegistry.ts` since that is the only place it is used. I prefer co-locating code unless it is used in different places.

src/lib/getPeerDependenciesFromRegistry.ts
```diff
@@ -22,14 +24,23 @@ async function getPeerDependenciesFromRegistry(packageMap: Index<VersionSpec>, o
...
+    async (accumPromise, [pkg, version]) => {
+      const dep = await packageManager.getPeerDependencies!(pkg, nodeSemver.coerce(version)?.version ?? version)
+      if (bar) {
+        bar.tick()
+      }
+      const accum = await accumPromise
+      const newAcc: Index<Index<string>> = { ...accum, [pkg]: dep }
+      const circularData = isCircularPeer(newAcc, pkg)
+      if (circularData.isCircular) {
+        delete newAcc[pkg][circularData.offendingPackage]
```
raineorshine, line 37: `isCircularPeer` looks good. My only concern is that it does not consider the peerDependency version itself. It assumes that if there is a cycle, then the package can be upgraded, regardless of what peerDependency is listed. Is that true?

raineorshine (changes requested):
Looks good! Just a few questions/comments.

src/lib/isCircularPeer.ts
CreativeTechGuy: Haha I originally had that but when looking around it looked like one function per file so I moved it. I'll move it back

test/peer.test.ts
```diff
@@ -1,52 +1,38 @@
...
-      await fs.rm(path.join(cwd, 'node_modules'), { recursive: true, force: true })
-      await fs.rm(path.join(cwd, 'package-lock.json'), { recursive: true, force: true })
-    }
+    const upgrades = await ncu({ cwd })
+    upgrades!.should.deep.equal({
+      'ncu-test-return-version': '2.0.0',
+    })
   })
 
   it('peer dependencies of installed packages are checked when using option peer', async () => {
```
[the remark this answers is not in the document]
CreativeTechGuy: Good catch!

test/peer.test.ts
```diff
@@ -1,52 +1,38 @@
-import fs from 'fs/promises'
 import path from 'path'
 import ncu from '../src/'
-import spawnNpm from '../src/package-managers/npm'
 import chaiSetup from './helpers/chaiSetup'
 
 chaiSetup()
 
 describe('peer dependencies', function () {
   it('peer dependencies of installed packages are ignored by default', async () => {
```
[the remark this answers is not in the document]
CreativeTechGuy: Good catch!

src/lib/getPeerDependenciesFromRegistry.ts
CreativeTechGuy: So honestly I don't know. All this does currently is remove a peer dependency if it's presence would have resulted in a circular dependency. It doesn't specifically force any updates, just allows the normal code to update if appropriate. In theory, this could allow an update which shouldn't have been allowed in a very specific situation where there's a circular peer dependency chain but the versions are configured in a certain way.

It's worth noting that it'll only remove one peer dependency, not all of them. So in the example of `vitest` and `@vitest/ui` which are peers of each other, it'll remove just one's peer of the other. So for example, `@vitest/ui` will still have the real peer of `vitest`, but `vitest`'s peer of `@vitest/ui` will be removed to break the cycle.

I think in most cases where there's circular peer dependencies it's because the versions of those packages need to be kept in-sync. So this should hopefully work. But it hurts my brain to think through this so I am struggling to think of all of the different use-cases where this could possibly not work.

src/lib/getPeerDependenciesFromRegistry.ts
```diff
@@ -22,14 +24,23 @@ async function getPeerDependenciesFromRegistry(packageMap: Index<VersionSpec>, o
...
-    const dep = await packageManager.getPeerDependencies!(pkg, version)
-    if (bar) {
-      bar.tick()
-    }
-    const accum = await accumPromise
-    return { ...accum, [pkg]: dep }
-  }, {})
+  const peerDependencies: Index<Index<string>> = Object.entries(packageMap).reduce(
+    async (accumPromise, [pkg, version]) => {
+      const dep = await packageManager.getPeerDependencies!(pkg, nodeSemver.coerce(version)?.version ?? version)
```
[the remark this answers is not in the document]
CreativeTechGuy: For the record, both `VersionSpec` and `Version` are just `string`. They are interchangeable. I'll update the type to be more accurate, but it won't actually protect any future mistakes.

And I realized why this wasn't an issue until now. It's because we are now calling `getPeerDependenciesFromRegistry` in `runLocal` and passing it the current versions from the package.json. Should I pre-transform the results of `getCurrentDependencies` before I pass it in or what would you like me to do?

CreativeTechGuy:
I resolved a many things as I could, a few follow up questions though.

src/lib/getPeerDependenciesFromRegistry.ts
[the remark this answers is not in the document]
raineorshine: > For the record, both `VersionSpec` and `Version` are just `string`. They are interchangeable. I'll update the type to be more accurate, but it won't actually protect any future mistakes.

Yes, I understand. Let me suggest a different way to look at it though. `VersionSpec` and `Version` are *not* actually interchangeable; it's just that the type system doesn't know this. So when it comes to `VersionSpec` and `Version`, we are thrown back into a weakly typed world. It's possible to narrow `Version` with a [template literal type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) or [branded primitive type](https://github.com/microsoft/TypeScript/wiki/FAQ#can-i-make-a-type-alias-nominal). In the mean time, we have to treat them as different types, even if this is not yet reflected in the type system.

> And I realized why this wasn't an issue until now. It's because we are now calling `getPeerDependenciesFromRegistry` in `runLocal` and passing it the current versions from the package.json. Should I pre-transform the results of `getCurrentDependencies` before I pass it in or what would you like me to do?

We could read the installed versions, as `getPeerDependencies` did, although I don't love the reliance on node_modules. Or we could use `semver.minVersion` to take the exact version at the bottom of the range, though this has the risk of getting the incorrect peerDependency if it has changed within the range. Maybe read from disk and fallback to `semver.minVersion` if the dependency is not installed. Open to your thoughts.

src/lib/isCircularPeer.ts
raineorshine: Sorry for the confusion! It's likely that the codebase is not consistent in this regard.

src/lib/getPeerDependenciesFromRegistry.ts
raineorshine: Thanks, that makes sense. Given the complexity, the difficulty in mapping out all cases, and the impossibility of knowing the real-world likelihood of these peerDependency combinations in the wild, I think it's best to move forward with the new logic as-is. I hate to introduce a possible regression, but it will be far easier to debug and reason about with a concrete use case in the future than speculating about it now.

src/lib/getPeerDependenciesFromRegistry.ts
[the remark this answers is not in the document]
CreativeTechGuy: I think since this library uses package.json as the source of truth for most things, it should probably take the versions from package.json and then do minVersion to remove any ranges and use that. That feels most appropriate.

raineorshine (approved):

raineorshine:
Thanks!

Published in `17.0.0-2` on the `next` tag.

(`main` is on track for a major version release on `latest`, but is currently blocked by a few PR's. v16.x is hotfixes only at this point.)

CreativeTechGuy:
Awesome! Thank you for your help here.

PR #1385, opened 2024-03-21
Loosen --format repo URL restrictions

### LINKED ISSUE #1384: [Feature] Allow fully resolved custom/non-repo urls to display with --format repo
Currently only a few common git hosts are effectively "allowlisted" and if a URL isn't recognized, nothing will be displayed. For companies which use private NPM registries and Git Hosts (or anyone doing something custom/different), it'd be great if any URL which is listed as `repository.url` in the `package.json` could be displayed.

The npm package.json spec allows for any URL (that leads to a repo) to be used, so it'd be great if npm-check-updates allowed that too.

With private, non-github repos internal to a company, they don't get to display links to repos/changelogs when using npm-check-updates which makes it harder to use.

CreativeTechGuy:
Closes #1384

### DIFF
--- a/src/lib/getRepoUrl.ts
+++ b/src/lib/getRepoUrl.ts
@@ -67,7 +67,7 @@ async function getRepoUrl(
       // It may already be a valid Repo URL
       const url = new URL(gitURL)
       // Some packages put a full URL in this field although it's not spec compliant. Let's detect that and use it if present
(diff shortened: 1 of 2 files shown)


raineorshine (approved):
Yes, I don't see any reason to restrict repo urls to github, gitlab, and bitbucket. Thanks!

PR #1394, opened 2024-04-09
Reduce the bundle size by removing some `lodash` usage

SukkaW:
Since `npm-check-updates@17`, the code and all the direct dependencies are now bundled together by the Vite.

Currently, the bundle size of the `npm-check-updates` is 7.1 MiB:

```shell
$ npm run build && du -sh build
# ....
7.1M	build
```

Since `npm-check-updates@17` now targets Node.js 18+, we can replace some lodash usage with the native ES6 syntax and features. By doing so, the PR reduces the final bundle size to `6.2 MiB`:

```sh
$ npm run build && du -sh build
# ....
6.2M	build
```


raineorshine (approved):
Thanks, much appreciated!

PR #1395, opened 2024-04-09
Chore: add bundle analyzer

SukkaW:
Follow up of #1394.

A bundle analyzer can assist us in reducing the size of the bundle in the areas that matter.

image


raineorshine:
Thanks!

PR #1399, opened 2024-04-24
fix: use HOME environment variable for npm configuration file lookup

plasticrake:
On a Windows machine I am forced to work on, my `$HOME` environment variable is different that what `os.homedir()` returns. This change first looks at `$HOME` to mimic how npm finds the user configuration file.

See: https://github.com/npm/cli/blob/dd39de7d1da743cbd33b671fa96f66667109b451/workspaces/config/lib/index.js#L313

Documentation: https://github.com/npm/cli/blob/7f4e66772ee631158b47fcfcd8e22b7b6b9b9cce/docs/lib/content/using-npm/config.md?plain=1#L48
```
per-user configuration file (defaults to `$HOME/.npmrc`; configurable via CLI
```

### DIFF
--- a/src/lib/libnpmconfig/index.js
+++ b/src/lib/libnpmconfig/index.js
@@ -23,14 +23,14 @@ const NpmConfig = figgyPudding(
 )
 
 const ConfigOpts = figgyPudding({
-  cache: { default: path.join(os.homedir(), '.npm') },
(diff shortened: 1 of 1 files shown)


raineorshine:
Looks good to me! Will publish tomorrow.

raineorshine:
Fixed and published in `v16.14.19`.

Issue #1419, opened 2024-06-01
yarn package manager is missing getPeerDependencies


rbnayax:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)

The yarn package manager is missing `getPeerDependencies` and as such can't perform some important operations when `pre` flag is set. This can be easily fixed by adding the npm functionality, just like pnpm does

PR #1420, opened 2024-06-01
fix: yarn package manager is missing getPeerDependencies #1419

rbnayax:
fix: #1419

### DIFF
--- a/src/package-managers/yarn.ts
+++ b/src/package-managers/yarn.ts
@@ -294,3 +294,5 @@ export const patch = withNpmConfigFromYarn(npm.patch)
 export const semver = withNpmConfigFromYarn(npm.semver)
 
 export default spawnYarn
+
+export { getPeerDependencies, packageAuthorChanged } from './npm'


rbnayax:
@raineorshine not sure why the test failed, it works locally on my ubuntu

raineorshine:
I re-ran the Node v18 ubuntu test twice, and unfortunately it continues to fail in GitHub Actions.

It looks like the doctor test is passing, but the success message gets written to `stderr` instead of `stdout` for some reason. I have seen this behavior occur with `bun` before, but not with npm. It seems to occur consistently in this branch and not in `main`.

rbnayax:
@raineorshine I fixed the unstable test. I think it has something to do with new npm behaviour on node 18 where it prints the commands it is about to run to stderr.

raineorshine (changes requested):
We can't completely skip the assertion, as it creates a hole in the test coverage if a future change breaks doctor mode on node v18. We have to assert that `stderr` is either empty, or just contains the name of the npm script.

I re-ran the tests on `main` and they failed at the same commit where they succeeded two weeks ago, so it seems that a minor or patch upgrade on node v18 or its bundled npm caused the break. Since this was not related to your PR, feel free to go back to the original one line change you made and ignore the errors. Or you can rewrite the assertion as described above, though that should go in a separate PR.

rbnayax:
> We can't completely skip the assertion, as it creates a hole in the test coverage if a future change breaks doctor mode on node v18. We have to assert that `stderr` is either empty, or just contains the name of the npm script.
> 
> I re-ran the tests on `main` and they failed at the same commit where they succeeded two weeks ago, so it seems that a minor or patch upgrade on node v18 or its bundled npm caused the break. Since this was not related to your PR, feel free to go back to the original one line change you made and ignore the errors. Or you can rewrite the assertion as described above, though that should go in a separate PR.

Done. separate PR in #1423

raineorshine (approved):

Issue #1418, opened 2024-06-01
Do not upgrade a package if its peer dependencies are not currently met


rbnayax:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)

---

## Steps to Reproduce

.ncurc.js:

```js
module.exports = {
  pre: false,
  interactive: true,
  enginesNode: true,
  cache: false,
  peer: true,
  upgrade: true,
  install: 'never',
  target: 'greatest'
}
```

Dependencies:

```json
{
    "eslint": "^8.57.0",
    "eslint-plugin-import": "^2.29.1",
    "eslint-plugin-unused-imports": "^3.2.0"
}
```

## Current Behavior

If you have the dependencies above, and you set `-pre` `eslint` package will not get upgraded to v9 because:
```
 eslint  ^8.57.0  →  ^9.4.0  reason: eslint-plugin-import requires ^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8
```

But `eslint-plugin-unused-imports` will be upgraded to 4.0.0. 
Version 4.0.0 has `eslint` set in its peer dependencies as 9. This renders the generated package.json file unusable.

## Expected Behavior

I would expect that with the `pre` flag set, if an upgraded package has unmet peer deps, it will not get upgraded

raineorshine:
Hi, thanks for reporting. I can confirm that this is a bug in https://github.com/raineorshine/npm-check-updates/blob/main/src/lib/upgradePackageDefinitions.ts. It seems that as `eslint-plugin-import` is removed, `eslint-plugin-unused-imports` accidentally gets added to the upgrades. Unfortunately I'm not super familiar with this file, since I wasn't the original implementator.

The next step would be to create a breaking unit test for this test case. Then we can attempt a solution. I don't have the bandwidth for this at the moment, but hopefully someone with some free time can take a stab at it.

PR #1421, opened 2024-06-02
fix: Do not upgrade if peerDependencies are not met

rbnayax:
fix: #1418

### DIFF
--- a/src/lib/upgradePackageDefinitions.ts
+++ b/src/lib/upgradePackageDefinitions.ts
@@ -1,5 +1,5 @@
 import { dequal } from 'dequal'
-import { satisfies } from 'semver'
+import { intersects, satisfies } from 'semver'
 import { parse, parseRange } from 'semver-utils'
 import { Index } from '../types/IndexType'
 import { Options } from '../types/Options'
@@ -11,6 +11,56 @@ import { pickBy } from './pick'
 import queryVersions from './queryVersions'
 import upgradeDependencies from './upgradeDependencies'
 
+type CheckIfInPeerViolationResult = {
+  violated: boolean
+  filteredUpgradedDependencies: Index<VersionSpec>
+  upgradedPeerDependencies: Index<Index<VersionSpec>>
+}
+
+/**
+ * Check if the peer dependencies constraints of each upgraded package, are in violation,
+ * thus rendering the upgrade to be invalid
+ *
+ * @returns Whether there was any violation, and the upgrades that are not in violation
+ */
+const checkIfInPeerViolation = (
+  currentDependencies: Index<VersionSpec>,
+  filteredUpgradedDependencies: Index<VersionSpec>,
+  upgradedPeerDependencies: Index<Index<VersionSpec>>,
+): CheckIfInPeerViolationResult => {
+  const upgradedDependencies = { ...currentDependencies, ...filteredUpgradedDependencies }
+  const filteredUpgradedDependenciesAfterPeers = pickBy(filteredUpgradedDependencies, (spec, dep) => {
+    const peerDeps = upgradedPeerDependencies[dep]
+    if (!peerDeps) {
+      return true
+    }
+    return Object.entries(peerDeps).every(
+      ([peer, peerSpec]) =>
+        upgradedDependencies[peer] === undefined || intersects(upgradedDependencies[peer], peerSpec),
+    )
+  })
+  const violated =
+    Object.keys(filteredUpgradedDependencies).length > Object.keys(filteredUpgradedDependenciesAfterPeers).length
+  let filteredUpgradedPeerDependencies = upgradedPeerDependencies
+  if (violated) {
+    filteredUpgradedPeerDependencies = pickBy(
+      upgradedPeerDependencies,
+      (spec, dep) => filteredUpgradedDependenciesAfterPeers[dep] || !filteredUpgradedDependencies[dep],
+    )
+  }
+  return {
+    violated,
+    filteredUpgradedDependencies: filteredUpgradedDependenciesAfterPeers,
+    upgradedPeerDependencies: filteredUpgradedPeerDependencies,
+  }
+}
+
+export type UpgradePackageDefinitionsResult = [
+  upgradedDependencies: Index<VersionSpec>,
+  latestVersionResults: Index<VersionResult>,
+  newPeerDependencies?: Index<Index<VersionSpec>>,
+]
+
 /**
  * Returns a 3-tuple of upgradedDependencies, their latest versions and the resulting peer dependencies.
  *
@@ -21,7 +71,7 @@ import upgradeDependencies from './upgradeDependencies'
 export async function upgradePackageDefinitions(
   currentDependencies: Index<VersionSpec>,
   options: Options,
-): Promise<[Index<VersionSpec>, Index<VersionResult>, Index<Index<VersionSpec>>?]> {
+): Promise<UpgradePackageDefinitionsResult> {
   const latestVersionResults = await queryVersions(currentDependencies, options)
 
   const latestVersions = keyValueBy(latestVersionResults, (dep, result) =>
@@ -47,22 +97,68 @@ export async function upgradePackageDefinitions(
 
   const filteredLatestDependencies = pickBy(latestVersions, (spec, dep) => filteredUpgradedDependencies[dep])
 
+  let result: UpgradePackageDefinitionsResult = [
+    filteredUpgradedDependencies,
+    latestVersionResults,
+    options.peerDependencies,
+  ]
+
+  if (!options.peer || Object.keys(filteredLatestDependencies).length === 0) {
+    return result
+  }
+
   if (options.peer && Object.keys(filteredLatestDependencies).length > 0) {
     const upgradedPeerDependencies = await getPeerDependenciesFromRegistry(filteredLatestDependencies, options)
-    const peerDependencies = { ...options.peerDependencies, ...upgradedPeerDependencies }
-    if (!dequal(options.peerDependencies, peerDependencies)) {
-      const [newUpgradedDependencies, newLatestVersions, newPeerDependencies] = await upgradePackageDefinitions(
-        { ...currentDependencies, ...filteredUpgradedDependencies },
-        { ...options, peerDependencies, loglevel: 'silent' },
+
+    let checkPeerViolationResult: CheckIfInPeerViolationResult
+
+    if (
+      dequal(options.peerDependencies, {
+        ...options.peerDependencies,
+        ...upgradedPeerDependencies,
+      })
+    ) {
+      checkPeerViolationResult = checkIfInPeerViolation(
+        currentDependencies,
+        filteredUpgradedDependencies,
+        options.peerDependencies!,
       )
-      return [
-        { ...filteredUpgradedDependencies, ...newUpgradedDependencies },
+      if (!checkPeerViolationResult.violated) {
+        return result
+      }
+    } else {
+      checkPeerViolationResult = {
+        violated: false,
+        filteredUpgradedDependencies,
+        upgradedPeerDependencies,
+      }
+    }
+    let runCount = 0
+    do {
+      if (runCount++ > 6) {
+        throw new Error(`Stuck in a while loop. Please report an issue`)
+      }
+      const peerDependenciesAfterUpgrade = {
+        ...options.peerDependencies,
+        ...checkPeerViolationResult.upgradedPeerDependencies,
+      }
+      if (dequal(options.peerDependencies, peerDependenciesAfterUpgrade)) {
+        // We can't find anything to do, will not upgrade anything
+        return [{}, latestVersionResults, options.peerDependencies]
+      }
+      const [newUpgradedDependencies, newLatestVersions, newPeerDependencies] = await upgradePackageDefinitions(
+        { ...currentDependencies, ...checkPeerViolationResult.filteredUpgradedDependencies },
+        { ...options, peerDependencies: peerDependenciesAfterUpgrade, loglevel: 'silent' },
+      )
+      result = [
+        { ...checkPeerViolationResult.filteredUpgradedDependencies, ...newUpgradedDependencies },
         { ...latestVersionResults, ...newLatestVersions },
         newPeerDependencies,
       ]
-    }
+      checkPeerViolationResult = checkIfInPeerViolation(currentDependencies, result[0], result[2]!)
+    } while (checkPeerViolationResult.violated)
   }
-  return [filteredUpgradedDependencies, latestVersionResults, options.peerDependencies]
+  return result
 }
 
 export default upgradePackageDefinitions
--- a/src/lib/getIgnoredUpgradesDueToPeerDeps.ts
+++ b/src/lib/getIgnoredUpgradesDueToPeerDeps.ts
@@ -1,9 +1,10 @@
-import { satisfies } from 'semver'
+import { intersects, satisfies } from 'semver'
 import { IgnoredUpgradeDueToPeerDeps } from '../types/IgnoredUpgradeDueToPeerDeps'
 import { Index } from '../types/IndexType'
 import { Options } from '../types/Options'
 import { Version } from '../types/Version'
 import { VersionSpec } from '../types/VersionSpec'
+import getPeerDependenciesFromRegistry from './getPeerDependenciesFromRegistry'
 import upgradePackageDefinitions from './upgradePackageDefinitions'
 
 /** Get all upgrades that are ignored due to incompatible peer dependencies. */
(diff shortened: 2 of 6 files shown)


rbnayax:
Hi @raineorshine 

The CI here fails as well I will investigate soon.

Regarding the fix - this works according to my tests but it is very complicated. Would appreciate you're review.

overview::

Current process:

1. upgrade deps, while making sure current peer deps are kept
2. check if peer deps are updated (due to upgrades)
3. rereun the upgrade process, with updated pacakges and peers (step 1) as the upgraded peers might allow more upgrades


Suggested:

1. upgrade deps, while making sure current peer deps are kept
2. If updates causes the new packages list to be in peer violation due to peers update:
  2.1. rereun the upgrade process with all the upgrades as it might sort itself and check if in violation again
  2.2. rerun the upgrae process without upgrades that cause violation (only if peers udpated after removal)
  2.3. return the results with the most new updates that is valid
3. else, check if peer deps are updated (due to upgrades)
  3.1. rereun the upgrade process, with updated pacakges and peers (step 1) as the upgraded peers might allow more upgrades

rbnayax:
Hi @raineorshine this is ready for initial review. It is not a straightforward fix and I would appreciate your input

src/lib/upgradePackageDefinitions.ts
```diff
@@ -11,6 +11,77 @@ import { pickBy } from './pick'
...
+  let wereUpgradedDependenceFiltered = false
+  const filteredUpgradedDependenciesAfterPeers = pickBy(filteredUpgradedDependencies, (spec, dep) => {
+    const peerDeps = filteredUpgradedPeerDependencies[dep]
+    if (!peerDeps) {
+      return true
+    }
+    const valid = Object.entries(peerDeps).every(
+      ([peer, peerSpec]) =>
+        upgradedDependenciesVersions[peer] === undefined ||
+        semver.satisfies(upgradedDependenciesVersions[peer], peerSpec),
```
raineorshine, line 40: See other comment about `intersects` vs `satisfies`.

src/lib/upgradePackageDefinitions.ts
```diff
@@ -11,6 +11,77 @@ import { pickBy } from './pick'
...
+const rerunUpgradeIfChangedPeers = async (
+  currentDependencies: Index<VersionSpec>,
+  filteredUpgradedDependencies: Index<VersionSpec>,
+  upgradedPeerDependencies: Index<Index<VersionSpec>>,
+  latestVersionResults: Index<VersionResult>,
+  options: Options,
+): Promise<void | [Index<VersionSpec>, Index<VersionResult>, Index<Index<VersionSpec>>?]> => {
+  if (Object.keys(filteredUpgradedDependencies).length === 0) {
+    return [{}, latestVersionResults, options.peerDependencies]
+  }
```
raineorshine, line 69: Any reason for this special case? It seems to work fine without it.

src/lib/upgradePackageDefinitions.ts
```diff
@@ -11,6 +11,77 @@ import { pickBy } from './pick'
...
+ *
+ * @returns
+ */
+const rerunUpgradeIfChangedPeers = async (
+  currentDependencies: Index<VersionSpec>,
+  filteredUpgradedDependencies: Index<VersionSpec>,
+  upgradedPeerDependencies: Index<Index<VersionSpec>>,
+  latestVersionResults: Index<VersionResult>,
+  options: Options,
+): Promise<void | [Index<VersionSpec>, Index<VersionResult>, Index<Index<VersionSpec>>?]> => {
```
raineorshine, line 66: Avoid implicit return undefined, as it makes it more difficult to follow all code paths. Instead, explicitly return `null`. In addition, the comments should make it clear what the semantics of returning `null` are.

src/lib/upgradePackageDefinitions.ts
```diff
@@ -11,6 +11,77 @@ import { pickBy } from './pick'
...
+  if (Object.keys(filteredUpgradedDependencies).length === 0) {
+    return [{}, latestVersionResults, options.peerDependencies]
+  }
+  const peerDependencies = { ...options.peerDependencies, ...upgradedPeerDependencies }
+  if (!dequal(options.peerDependencies, peerDependencies)) {
```
raineorshine, line 71: This condition would be better inverted:

```ts
if (dequal(options.peerDependencies, peerDependencies)) return null

...
```

src/lib/upgradePackageDefinitions.ts
```diff
@@ -49,17 +120,55 @@ export async function upgradePackageDefinitions(
...
+    )
+    // Try another run to see if we can upgrade some more
+    let rerunResult = await rerunUpgradeIfChangedPeers(
+      currentDependencies,
+      filteredUpgradedDependencies,
+      upgradedPeerDependencies,
+      latestVersionResults,
+      options,
+    )
+    if (!rerunResult) {
```
raineorshine, line 136: There is a lot of duplication between the second run (step 2) and the while loop (step 3). Can you combine this into a single while or do-while loop? You may be able to prepopulate `checkPeerViolationResult` to combine the logic.

src/lib/upgradePackageDefinitions.ts
```diff
@@ -11,6 +11,77 @@ import { pickBy } from './pick'
 import queryVersions from './queryVersions'
 import upgradeDependencies from './upgradeDependencies'
 
+/**
+ * check if in peer violation
```
raineorshine, line 15: Let's come up with a more descriptive JSDOC comment for this function. 

What does it mean for a peerDependency to be in violation? What are the different conditions that this function checks?

src/lib/upgradePackageDefinitions.ts
```diff
@@ -11,6 +11,77 @@ import { pickBy } from './pick'
...
+  upgradedPeerDependencies: Index<Index<VersionSpec>>,
+) => {
+  const upgradedDependencies = { ...currentDependencies, ...filteredUpgradedDependencies }
+  const upgradedDependenciesVersions = Object.fromEntries(
+    Object.entries(upgradedDependencies).map(([packageName, versionSpec]) => {
+      return [packageName, semver.minVersion(versionSpec)?.version ?? versionSpec]
+    }),
+  )
+  const filteredUpgradedPeerDependencies = { ...upgradedPeerDependencies }
+  let wereUpgradedDependenceFiltered = false
```
raineorshine, line 31: Let's call variable this `violated` to be consistent with the function name.

src/lib/upgradePackageDefinitions.ts
```diff
@@ -11,6 +11,77 @@ import { pickBy } from './pick'
...
+        semver.satisfies(upgradedDependenciesVersions[peer], peerSpec),
+    )
+    if (!valid) {
+      wereUpgradedDependenceFiltered = true
+      delete filteredUpgradedPeerDependencies[dep]
+    }
+    return valid
+  })
+  return {
+    issuesFound: wereUpgradedDependenceFiltered,
```
raineorshine, line 49: Let's call this key `violated` for consistency.

raineorshine (changes requested):
Thanks for your contribution. I did my best to review, but it is admittedly complicated, as you pointed out. The unit test helps a lot though, so thanks for that. If there are any other test cases you want to add, that would help make this more robust.

In general, the code deserves some more thorough commenting. Otherwise it's going to be challenging to come back to this code in the future if any issues arise, or if the functionality needs to be extended.

src/lib/upgradePackageDefinitions.ts
rbnayax: done

src/lib/upgradePackageDefinitions.ts
rbnayax: agree, done

src/lib/upgradePackageDefinitions.ts
rbnayax: done

src/lib/upgradePackageDefinitions.ts
rbnayax: done

src/lib/upgradePackageDefinitions.ts
rbnayax: Done

src/lib/upgradePackageDefinitions.ts
rbnayax: was hard :-) but done

src/lib/upgradePackageDefinitions.ts
rbnayax: The code does reach there (I added a test) but it is not really needed so I removed it

src/lib/getIgnoredUpgrades.ts
```diff
@@ -13,36 +14,54 @@ export async function getIgnoredUpgrades(
...
+        )
+        .reduce((accumReason, [peerPkg, peers]) => ({ ...accumReason, [peerPkg]: peers[pkgName] }), {} as Index<string>)
+      if (Object.keys(reason).length === 0) {
+        const peersOfPkg = upgradedPeerDependenciesLatest?.[pkgName] || {}
+        reason = Object.entries(peersOfPkg)
+          .filter(
+            ([peer, peerSpec]) =>
+              upgradedPackagesWithPeerRestriction[peer] &&
+              !satisfies(upgradedPackagesWithPeerRestriction[peer], peerSpec),
+          )
```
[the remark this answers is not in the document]
rbnayax: I'm using the same logic as in:
https://github.com/raineorshine/npm-check-updates/blob/0f17df28573aea128f40335e9856526f619db9e3/src/lib/runLocal.ts#L183-L192

I don't think intersect is the right way to go as it is not sufficient to make sure that peer dependency is met. For example, if package P1 spec is `^1.0.0` and P2 has P1 as peer with spec `^1.2.0`, the 2 ranges will intersect  but it will not mean that the peer dependency is met...

`current` is not sufficient, as we need to check against the versions that are currently upgraded.

Thoughts?

src/lib/upgradePackageDefinitions.ts
rbnayax: see my reply there

rbnayax:
@raineorshine I pushed all the fixes, looking forward to your next review

src/lib/getIgnoredUpgrades.ts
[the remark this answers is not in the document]
raineorshine: But if the P1 spec is `^1.0.0` (not `1.0.0`) then npm will install `1.2.0` or later and the peer dependency will be met, no?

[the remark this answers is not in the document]
rbnayax: That will not be the case if your lcokfile was not updated and P1 was not upgraded due to a filter. But if you think that will be better I will make the change, looking forward to your opinion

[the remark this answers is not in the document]
raineorshine: npm-check-updates is generally just concerned with package.json versions, and assumes that the user will run `npm install` and manage the lockfile through npm. The lockfile is basically always out of date after running `ncu -u`, so I don't see that as a problem. It would be a problem if npm-check-updates upgraded a version range to something that was incompatible though. If you think that's the case, please let me know. I didn't notice that the existing code uses the same `minVersion` trick to get the bottom of the range.

I'm having a hard time picturing the situation you described with a filter on P1. If you think that's a problem, we should add a unit test that covers that. It is admittedly difficult to reason about this abstractly, at least for me.

src/lib/upgradePackageDefinitions.ts
```diff
@@ -49,18 +127,47 @@ export async function upgradePackageDefinitions(
...
+        checkPeerViolationResult = checkIfInPeerViolation(currentDependencies, rerunResult[0], rerunResult[2]!)
+        if (!checkPeerViolationResult.violated) {
+          // We found a stable solution
+          return rerunResult
+        }
+      }
+      if (runIndex++ > 5) {
+        throw new Error(`Stuck in a while loop. Please report an issue`)
+      }
+    } while (true)
```
raineorshine, line 170: This would make more sense as `while (checkPeerViolationResult.violated)` since the dependencies will keep getting checked until there are no more violations.

src/lib/upgradePackageDefinitions.ts
```diff
@@ -49,18 +127,47 @@ export async function upgradePackageDefinitions(
...
+        }
+        checkPeerViolationResult = checkIfInPeerViolation(
+          currentDependencies,
+          filteredUpgradedDependencies,
+          upgradedPeerDependencies,
+        )
+        if (!checkPeerViolationResult.violated) {
+          // No issues were found, return
+          return [filteredUpgradedDependencies, latestVersionResults, options.peerDependencies]
+        }
```
raineorshine, line 159: This if statement can be removed once the loop condition changes to `while (checkPeerViolationResult.violated)`.

src/lib/upgradePackageDefinitions.ts
```diff
@@ -49,18 +127,47 @@ export async function upgradePackageDefinitions(
...
+      } else {
+        checkPeerViolationResult = checkIfInPeerViolation(currentDependencies, rerunResult[0], rerunResult[2]!)
```
raineorshine, line 161: I suspect that the two calls to `checkIfInPeerViolation` but it's a little hard to see now. Once the other refactoring is completed I will reevaluate.

raineorshine (changes requested):
Getting closer. Thanks for your effort!

src/lib/upgradePackageDefinitions.ts
```diff
@@ -11,6 +11,84 @@ import { pickBy } from './pick'
...
+
+type RerunUpgradeResult = null | [Index<VersionSpec>, Index<VersionResult>, Index<Index<VersionSpec>>?]
+
+/**
+ * rerun upgrade if changed peers
+ *
+ * @returns `null` if no changes were made to the peer dependence as a result of the recent upgrade
+ * else returns the upgrade result
+ */
+const rerunUpgradeIfChangedPeers = async (
```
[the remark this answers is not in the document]
rbnayax: Done

src/lib/upgradePackageDefinitions.ts
rbnayax: Done

src/lib/upgradePackageDefinitions.ts
rbnayax: Done

src/lib/upgradePackageDefinitions.ts
rbnayax: Done

src/lib/upgradePackageDefinitions.ts
```diff
@@ -11,6 +11,84 @@ import { pickBy } from './pick'
...
+  filteredUpgradedDependencies: Index<VersionSpec>,
+  upgradedPeerDependencies: Index<Index<VersionSpec>>,
+): CheckIfInPeerViolationResult => {
+  const upgradedDependencies = { ...currentDependencies, ...filteredUpgradedDependencies }
+  const upgradedDependenciesVersions = Object.fromEntries(
+    Object.entries(upgradedDependencies).map(([packageName, versionSpec]) => {
+      return [packageName, minVersion(versionSpec)?.version ?? versionSpec]
+    }),
+  )
+  const filteredUpgradedPeerDependencies = { ...upgradedPeerDependencies }
```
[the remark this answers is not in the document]
rbnayax: Honestly, I can't see how you can filter down both at the same time. Their keys are overlapping and not contained in one another and as such there is no single set of keys I can reduce on. If you can post a snippet showing how, I would appreciate it.

That being said, I did change it a bit to be cleaner without deleting keys, let me know what you think

src/lib/getIgnoredUpgrades.ts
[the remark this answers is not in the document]
rbnayax: Im not sure I can write a test for it. I changed to intresect in both places where I used minVersion

rbnayax:
@raineorshine I pushed all the fixes, looking forward to your next review

raineorshine:
Thanks! I have some work responsibilities right now, but I'll review as soon as I can.

rbnayax:
> Thanks! I have some work responsibilities right now, but I'll review as soon as I can.

Thanks I appreciate it

rbnayax:
> Thanks! I have some work responsibilities right now, but I'll review as soon as I can.

Hi @raineorshine just a ping to remind you in case you forgot :-)

src/lib/upgradePackageDefinitions.ts
```diff
@@ -49,18 +100,42 @@ export async function upgradePackageDefinitions(
...
-      ]
+
+    let checkPeerViolationResult: CheckIfInPeerViolationResult = {
+      violated: false,
+      filteredUpgradedDependencies,
+      upgradedPeerDependencies,
     }
+    let rerunResult: UpgradePackageDefinitionsResult
+    let runIndex = 0
+    do {
```
raineorshine, line 111: I'm struggling to understand this do-while loop. Something about the `runIndex > 1` "base case" and the constant reassignment of `rerunResult. 

Any ideas how to clean this up and make it easier to understand?

raineorshine (changes requested):

src/lib/upgradePackageDefinitions.ts
rbnayax: I gave it another shot

src/lib/upgradePackageDefinitions.ts
```diff
@@ -11,6 +11,57 @@ import { pickBy } from './pick'
...
+    )
+    if (!valid) {
+      violatedDependencies.add(dep)
+    }
+    return valid
+  })
+  const violated = violatedDependencies.size > 0
+  let filteredUpgradedPeerDependencies = upgradedPeerDependencies
+  if (violated) {
+    filteredUpgradedPeerDependencies = pickBy(upgradedPeerDependencies, (spec, dep) => !violatedDependencies.has(dep))
```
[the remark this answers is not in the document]
rbnayax: not quire but I changed it

raineorshine:
Thanks, I'll take a look soon.

Would you mind rebasing on `main`? I just merged your other PR and there are some merge conflicts. Thank you.

rbnayax:
> Thanks, I'll take a look soon.
> 
> Would you mind rebasing on `main`? I just merged your other PR and there are some merge conflicts. Thank you.

of course already on it

rbnayax:
> Thanks, I'll take a look soon.
> 
> Would you mind rebasing on `main`? I just merged your other PR and there are some merge conflicts. Thank you.

Done

raineorshine:
Sorry about the delay! I've had work responsibilities and have not found time to review this yet.

rbnayax:
@raineorshine how are you? Can we expect this to be reviewed soon? It will help my team a lot to have that feature in production

raineorshine:
Yes. I've been at a conference this week, but I will try to get to it soon.

raineorshine (approved):
Looks good! Thanks for your patience. 

Will release tonight or tomorrow.

PR #1423, opened 2024-06-08
fix: ci flakiness

rbnayax:


rbnayax:
@raineorshine falkiness started appearing on node 20 as well. I think the fix here is the best, looking forward to your commnets

PR #1430, opened 2024-07-16
fix: remove colon from git+ssh paths

jacktuck:
```
    "ncu-test-v2": "git+ssh://git@github.com/raineorshine/ncu-test-v2#semver:^2.0.0"
```
works fine

but
```
    "ncu-test-v2": "git+ssh://git@github.com:raineorshine/ncu-test-v2#semver:^2.0.0"
``` 

does not

This should work with both now.


raineorshine:
Thank you

PR #1431, opened 2024-07-17
fix: git tags

jacktuck:
Hi

Thanks for merging the PR from yesterday. I just noticed I updated the http:// promise rather than the ssh:// by mistake.

Sorry about that. 

Is it possible to make a new release once this is merged? (prerelease would be fine)


jacktuck:
Thank you :)

raineorshine:
Thanks for the catch.

Published to `v17.0.0-6`

PR #1443, opened 2024-08-07
fix: npm get peer deps returns list and not object when using range

rbnayax:
Running the test without the fix results in:
```
AssertionError: expected { 'eslint-plugin-unused-imports': [ { eslint: '9', '@typescript-eslint/eslint-plugin': '8' }, { '@typescript-eslint/eslint-plugin': '^8.0.0-0', eslint: '^9.0.0' } ] } to deeply equal { 'eslint-plugin-unused-imports': { '@typescript-eslint/eslint-plugin': '^8.0.0-0', eslint: '^9.0.0' } }
```

### DIFF
--- a/test/getPeerDependenciesFromRegistry.test.ts
+++ b/test/getPeerDependenciesFromRegistry.test.ts
@@ -15,6 +15,17 @@ describe('getPeerDependenciesFromRegistry', function () {
     })
   })
 
+  it('single package with range', async () => {
+    await chalkInit()
+    const data = await getPeerDependenciesFromRegistry({ 'eslint-plugin-unused-imports': '^4' }, {})
+    data.should.deep.equal({
+      'eslint-plugin-unused-imports': {
+        '@typescript-eslint/eslint-plugin': '^8.0.0-0',
+        eslint: '^9.0.0',
+      },
+    })
+  })
+
   it('single package empty', async () => {
     await chalkInit()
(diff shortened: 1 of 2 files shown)


test/getPeerDependenciesFromRegistry.test.ts
```diff
@@ -15,6 +15,17 @@ describe('getPeerDependenciesFromRegistry', function () {
     })
   })
 
+  it('single package with range', async () => {
+    await chalkInit()
+    const data = await getPeerDependenciesFromRegistry({ 'eslint-plugin-unused-imports': '^4' }, {})
```
raineorshine, line 20: `getPeerDependenciesFromRegistry` is currently typed to take an exact version (`Version`), not a version range (`VersionSpec`). They both resolve to `string`, so the type checker does not catch this yet.

Is there somewhere in the code that is calling `getPeerDependenciesFromRegistry` with a version range instead of an exact version? I think we need to fix it there. Getting the peer dependencies of an entire version range doesn't really make sense to me.

raineorshine (changes requested):

test/getPeerDependenciesFromRegistry.test.ts
rbnayax: You are correct, but for the life of me, I can't remember how I came to find that "issue". I can't reproduce it locally and I can't see how version spec was passed here. So I will close this PR for now

PR #1447, opened 2024-08-07
fix: check if upgrade ignored due to engines fails due to usage of ranges

rbnayax:
@raineorshine this is a show stopper for us, I would appreciate your input

### DIFF
--- a/src/lib/getEnginesNodeFromRegistry.ts
+++ b/src/lib/getEnginesNodeFromRegistry.ts
@@ -1,6 +1,7 @@
 import ProgressBar from 'progress'
 import { Index } from '../types/IndexType'
 import { Options } from '../types/Options'
+import { VersionResult } from '../types/VersionResult'
 import { VersionSpec } from '../types/VersionSpec'
 import getPackageManager from './getPackageManager'
 
@@ -11,7 +12,7 @@ import getPackageManager from './getPackageManager'
  * @param [options={}] Options.
  * @returns Promised {packageName: engines.node} collection
  */
-async function getEnginesNodeFromRegistry(packageMap: Index<VersionSpec>, options: Options) {
+async function getEnginesNodeFromRegistry(packageMap: Index<VersionResult>, options: Options) {
   const packageManager = getPackageManager(options, options.packageManager)
   if (!packageManager.getEngines) return {}
 
@@ -23,7 +24,9 @@ async function getEnginesNodeFromRegistry(packageMap: Index<VersionSpec>, option
   }
 
   return Object.entries(packageMap).reduce(async (accumPromise, [pkg, version]) => {
-    const enginesNode = (await packageManager.getEngines!(pkg, version, options)).node
+    const enginesNode = version.version
+      ? (await packageManager.getEngines!(pkg, version.version, options)).node
+      : undefined
     if (bar) {
       bar.tick()
     }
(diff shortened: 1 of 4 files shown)


src/lib/getEnginesNodeFromRegistry.ts
```diff
@@ -11,7 +12,7 @@ import getPackageManager from './getPackageManager'
  * @param [options={}] Options.
  * @returns Promised {packageName: engines.node} collection
  */
-async function getEnginesNodeFromRegistry(packageMap: Index<VersionSpec>, options: Options) {
+async function getEnginesNodeFromRegistry(packageMap: Index<VersionResult>, options: Options) {
```
raineorshine, line 15: The function only uses version, so why pass in more information than what is used?

src/lib/getIgnoredUpgradesDueToEnginesNode.ts
```diff
@@ -21,13 +21,13 @@ export async function getIgnoredUpgradesDueToEnginesNode(
...
   if (!optionsEnginesNodeMinVersion) return {}
-  const [upgradedLatestVersions] = await upgradePackageDefinitions(current, {
+  const [upgradedLatestVersions, latestVersionResults] = await upgradePackageDefinitions(current, {
     ...options,
     enginesNode: false,
     nodeEngineVersion: undefined,
     loglevel: 'silent',
   })
-  const enginesNodes = await getEnginesNodeFromRegistry(upgradedLatestVersions, options)
+  const enginesNodes = await getEnginesNodeFromRegistry(latestVersionResults, options)
```
raineorshine, line 30: Can you explain this change for me?

raineorshine (changes requested):

src/lib/getIgnoredUpgradesDueToEnginesNode.ts
rbnayax: upgradedLatestVersions contains specs (`^4.0.0`) which can't be used to get the info from npm registry. So we use the actual versions instead

src/lib/getEnginesNodeFromRegistry.ts
rbnayax: done

src/lib/getIgnoredUpgradesDueToEnginesNode.ts
raineorshine: First, in that case, let's change the first param of `getEnginesNodeFromRegistry` to use `Version` instead of `VersionSpec`.

i.e.:

```ts
async function getEnginesNodeFromRegistry(packageMap: Index<Version>, options: Options)
```

Eventually we should change `Version` to a branded type so that the type system can enforce the distinction between an exact version and a version range.

---

Second, `latestVersionResults` is not filtered the way that `upgradedLatestVersions` is. Is that desired? I'm thinking we probably don't need to look up the engines node version for dependencies that have already been filtered out due to the `minimal` or `filterResults` options.

This filters down `latestVersionResults` to just the keys in `upgradedLatestVersions`:

```ts
const filteredLatestVersionResults = pickBy(latestVersionResults, (spec, dep) => upgradedLatestVersions[dep])
```

rbnayax: I already changed getEnginesNodeFromRegistry to use Version.
Yes, branded types might be the way thought it comes with annoying boilerplate :-(
Yes you are correct, applied a fix

raineorshine (approved):

PR #1449, opened 2024-08-21
Add installedVersion format option

jakubdrozdek:
A new format option called exactCurrentVersion has been added. This option replaces the usual version range copied from the project's package.json file with the exact version of the installed package. For instance, this feature might be helpful if you need to generate a release changelog gist for the updated package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ Options that take no arguments can be negated by prefixing them with `--no-`, e.
   </tr>
   <tr>
     <td><a href="#format">--format &lt;value&gt;</a></td>
-    <td>Modify the output formatting or show additional information. Specify one or more comma-delimited values: group, ownerChanged, repo, time, lines. (default: [])</td>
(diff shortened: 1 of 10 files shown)


jakubdrozdek:
@raineorshine Seems like the tests for `getIgnoredUpgradesDueToEnginesNode` rely on the actual data returned from the NPM registry, and are flaky - when you run them again in a month, they might fail with no changes in the code.

raineorshine (changes requested):
> A new format option called exactCurrentVersion has been added. This option replaces the usual version range copied from the project's package.json file with the exact version of the installed package. For instance, this feature might be helpful if you need to generate a release changelog gist for the updated package.

Thanks! Let's rename this to `installedVersion` if you don't mind. It's shorter and it makes it clear that this feature depends on the installed packages in node_modules, which not every package manager supports.

> @raineorshine Seems like the tests for `getIgnoredUpgradesDueToEnginesNode` rely on the actual data returned from the NPM registry, and are flaky - if you'd run them again in a month, they might fail with no changes in the code.

Thanks for bringing this to my attention. We should definitely mock those requests.

PR #1457, opened 2024-09-21
Refactor .ncurc loading 

### LINKED ISSUE #1452: `format` remains a string when set via `ncurc.js`

---

## Steps to Reproduce

.ncurc.js:

```js
module.exports = {
  interactive: true,
  format: 'group' ,// <-- Where the problem starts
  deep: true // <-- What manifests the problem
};
```

Dependencies: _Not relevant_

Steps:

1. Run `ncu` with the above config.

## Current Behavior

When the config is loaded, `format: 'group'` is loaded and kept as a string. Later, after selecting dependencies, ncu crashes with an error: `TypeError: (r.format || []).filter is not a function`

I did a little digging and determined that the error comes from [runLocal.ts#L251](https://github.com/raineorshine/npm-check-updates/blob/27156434c2936568b2d3be929c9eaa4c084ecc8d/src/lib/runLocal.ts#L251).  Since `'group'`  is a string, it doesn't have a `.format` method.

This probably hasn't been noticed before because ncu works just fine with `format: 'group'`, so long as `deep: true` isn't _also_ enabled. Other parts of the code -- such as `options.format?.includes('group')` on [runLocal.ts#L93](https://github.com/raineorshine/npm-check-updates/blob/27156434c2936568b2d3be929c9eaa4c084ecc8d/src/lib/runLocal.ts#L93) -- work largely as expected because strings also happen to have a mostly-compatible `.includes` method.

Specifying `format: ['group']` results in the correct behavior.

## Expected Behavior

I would expect one of two behaviors:
1. A validation error on load, stating that `format` must be an array.
2. Or, automatically convert the string to an array [like the CLI does](https://github.com/raineorshine/npm-check-updates/blob/27156434c2936568b2d3be929c9eaa4c084ecc8d/src/cli-options.ts#L669)

I believe the latter option is the expected behavior, as [the project's own `.ncurc.js` file](https://github.com/raineorshine/npm-check-updates/blob/27156434c2936568b2d3be929c9eaa4c084ecc8d/.ncurc.js#L2) uses `format: 'group'`.

jftanner:
My primary intent here was to fix #1452. Internally, `format` is expected to be an array. However, when it's loaded outside the cli (in `runUpgrades`) it's not parsed as such. The _correct_ answer probably would have been to validate it and throw an error. However, since even this repo's own `.ncurc` file uses a string, that would definitely be a breaking change.

While this fixes the problem, I recommend refactoring the app to validate the loaded `.ncurc` file every time one is loaded, and avoid passing into commander as arguments. There remain a number of type-safety issues with invalid configs that are outside the scope of this PR.

The code to set `args` didn't _need_ to change, but I'd already cleaned it up while chasing a red herring, so I figured I'd leave it.

I didn't see commit guidelines anywhere, but let me know if you'd prefer a semantic-commit format and I can change it.

Oh, and thanks for this incredibly helpful tool!

### DIFF
--- a/src/lib/getNcuRc.ts
+++ b/src/lib/getNcuRc.ts
@@ -37,9 +37,9 @@ async function getNcuRc({
   }
 
   // convert the config to valid options by removing $schema and parsing format
-  const { $schema: _, format, ...rawConfig } = rawResult?.config || {}
+  const { $schema: _, ...rawConfig } = rawResult?.config || {}
   const config: Options = rawConfig
-  if (typeof format === 'string') config.format = format.split(',')
+  if (typeof config.format === 'string') config.format = cliOptionsMap.format.parse!(config.format)
 
   // validate arguments here to provide a better error message
   const unknownOptions = Object.keys(config).filter(arg => !cliOptionsMap[arg])
--- a/src/lib/getNcuRc.ts
+++ b/src/lib/getNcuRc.ts
@@ -3,6 +3,7 @@ import path from 'path'
 import { rcFile } from 'rc-config-loader'
 import { cliOptionsMap } from '../cli-options'
 import { Options } from '../types/Options'
+import { RcOptions } from '../types/RcOptions'
 import programError from './programError'
 
 /** Loads the .ncurc config file. */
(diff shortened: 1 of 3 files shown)


jftanner:
Ah... the tests run against `build`. :(
I'm used to using `tsx` to run typescript directly, so I didn't even think to build it before testing.
I'll need to debug this over the weekend.

jftanner:
It's because I was setting `{ format: undefined }`. :)

src/lib/getNcuRc.ts
```diff
@@ -23,52 +24,47 @@ async function getNcuRc({
...
-
-  const result = {
-    filePath: rawResult?.filePath,
-    // Prevent the cli tool from choking because of an unknown option "$schema"
-    config: rawConfigWithoutSchema,
-  }
+  // convert the config to valid options by removing $schema and parsing format
+  const { $schema: _, format, ...rawConfig } = rawResult?.config || {}
+  const config: Options = rawConfig
+  if (typeof format === 'string') config.format = format.split(',')
```
raineorshine, line 42: Let's import and use [formatOption.parse](https://github.com/raineorshine/npm-check-updates/blob/8a5967a354e6b21ef7ae7180b929a93e0401cb77/src/cli-options.ts#L669) rather than duplicate the parsing logic. Having the same parsing logic in multiple places is more likely to introduce regressions in the future if one is changed and not the other.

Ideally we would generalize this to all options rather than only `format`, but I'm okay with it for the sake of limiting the scope of this PR. Thanks!

raineorshine (changes requested):
Thanks, looks good!

> While this fixes the problem, I recommend refactoring the app to validate the loaded `.ncurc` file every time one is loaded, and avoid passing into commander as arguments. There remain a number of type-safety issues with invalid configs that are outside the scope of this PR.

Yes, good point.

> Ah... the tests run against `build`. :(
> I'm used to using `tsx` to run typescript directly, so I didn't even think to build it before testing.

That's not a great test setup on my part. `npm test` should not depend on anything other than dependencies being installed to work!

src/lib/getNcuRc.ts
jftanner: > Ideally we would generalize this to all options

I considered that (and, in fact, an early version did exactly that), but I looked through all the options and not all the parsers would apply correctly. Some of the array ones would even break. In all other cases, the documented types match the actual, internal types. Format is the only weird one. 

Regardless, you're spot on about loading the parser function. I meant to do that after my sanity test and forgot. I'll get that changed.

raineorshine: Very good points

jftanner: It's a good thing I went back! There was a pretty glaring bug in which it would drop any format option that _isn't_ a string. Clearly, I was getting tired by the time I finished this. 😭

jftanner: New commit should be better, but take a look to be sure.

raineorshine: Good catch! Surprised that wasn't covered by the unit tests. 

I'll take a look soon.

jftanner: Me too! They seem so thorough, I got lazy. Embarrassing noob mistake on my part. :'(

Speaking of the tests, it's worth noting that they are impacted by the dev environment. When I first started, they were broken by my `~/.npmrc` having a default private registry. And they ask for my SSH key when testing against GitHub. Mocking those away would probably be a nightmare, but worth mentioning. Unfortunately, I don't have the bandwidth to contribute anything meaningful in that space.

raineorshine: Yes, I agree. There are some live tests to intentionally cover the data that comes back from the npm registry, and some simply due to exceeding the current mocking capabilities of the project, but there is probably a better way to do this.

jftanner: If you have lots of contributors with this issue, you might consider putting un-mocked tests in a standardized container. But since most folks aren't going to have private registries configured, it's probably not worth the docker dependency. `mv ~/.npmrc ~/.npmrc.bak` was an easy workaround.

raineorshine (approved):

Issue #1429, opened 2024-07-08
Calls to spawn that include '.cmd' or '.bat' on windows should include '{shell: true}' argument


phaze-ZA:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)

This is due to the recent security patch released by node.js: https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2

---

## Steps to Reproduce

- Be on nodejs > 20.12.1
- Be on windows
- run npm-check-updates
- notice EINVAL error

## Current Behavior
EINVAL error is thrown

## Expected Behavior
Command is executed

raineorshine:
Hi, thanks for reporting. 

If there is a Windows user that would like to test a fix and open a PR that would be helpful.

phaze-ZA:
Will set something up locally and do a thingy

cakidnyc:
Any update? I can't use ncu on windows after updating node above v18.12.0.

phaze-ZA:
Apologies work got intense and I forgot about this. Busy testing a fix now and will submit a PR once I confirm that it works on both windows and non-windows

phaze-ZA:
@cakidnyc would you mind testing this on your side? Seems to be working fine on Mac and Windows here

cakidnyc:
It's good for me now on Windows 11 w/ node 20.16.0, thanks!    It does turn out that I had to manually delete ncu first ( npm install -g npm-check-updates didn't seem to upgrade it and uninstall -g wasn't clearing it out either).  After the manual delete and reinstall, then it starting working. So, it might be that it actually was working earlier, my install was just messed up.

PR #1465, opened 2024-10-16
Sanitize spawn calls which contain .cmd for win32

phaze-ZA:
See https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2 and 
https://github.com/raineorshine/npm-check-updates/issues/1429

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -174,6 +174,7 @@ const install = async (
               // See: https://github.com/raineorshine/npm-check-updates/issues/1191
               ...(packageManager === 'pnpm' ? { npm_config_strict_peer_dependencies: false } : null),
             },
+            shell: process.platform === 'win32',
           },
         )
         print(options, stdout)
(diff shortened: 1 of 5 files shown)


raineorshine:
There was a broken test in `main` that snuck in because the test was incorrectly written against live data from the npm registry that changed. I disabled it in 3f5f58af and rebased the PR branch on `main` so the CI result should be accurate now.

raineorshine (approved):
Thanks!

phaze-ZA:
Did some more digging, I'm thinking it may make sense to move away from  [cross spawn](https://github.com/moxystudio/node-cross-spawn) at some point. It hasn't been updated in the last 4 years and doesn't look like the owner is actually responding to issues or PR's. The reason this was failing is probably due to the path in `__dirname` is posix, which windows does not fancy. `cross-spawn` does some handling for that when it parses the arguments, but only if `shell` is not set to `true` (which is exactly what we want it to be). So solution might be to sanitize the paths before we call `spawn`. Thing is, once we go down this rabbit hole, we're effectively doing what `cross-spawn` does so it makes having that as a dependency less useful

phaze-ZA:
TLDR; reason it failed with my fix was due to `cross-spawn` not parsing the options (due to `{shell: true}`.

PR #1466, opened 2024-10-16
Revert "Sanitize spawn calls which contain .cmd for win32"

phaze-ZA:
Reverts raineorshine/npm-check-updates#1465

We need to revert this otherwise Windows machines will break - https://github.com/raineorshine/npm-check-updates/pull/1465#issuecomment-2417958200

I suspect the fix may take a bit more time. I wouldn't mind spending a weekend on replacing `cross-spawn` if you'd like.


raineorshine:
I would be grateful to have your help!

Issue #1470, opened 2024-10-27
yarn getPeerDependencies doesnt work on private reporsitories when credentials are only set for yarn and not npm


rbnayax:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)

---

## Steps to Reproduce

use only yarn with private repo on your company scope and authenticate only in yarn
run upgrade with --peers that should skip an upgrade due to peer dependencies in one of your company's dependencies

## Current Behavior

the upgrade is not skipped
reason: getPeerDependencies is using npm cli, which is not authenticated

## Expected Behavior

skip the dependency upgrade, while using yarn to realize the peer dependencies information

raineorshine:
Published in `v17.1.7`.

PR #1471, opened 2024-10-27
fix: yarn getPeerDependencies doesnt work on private reporsitories wh…

rbnayax:
…en credentials are only set for yarn and not npm 

fix: #1470

### DIFF
--- a/test/package-managers/npm/index.test.ts
+++ b/test/package-managers/npm/index.test.ts
@@ -26,8 +26,8 @@ describe('npm', function () {
   })
 
   it('getPeerDependencies', async () => {
-    await npm.getPeerDependencies('ncu-test-return-version', '1.0').should.eventually.deep.equal({})
-    await npm.getPeerDependencies('ncu-test-peer', '1.0').should.eventually.deep.equal({
+    await npm.getPeerDependencies('ncu-test-return-version', '1.0.0').should.eventually.deep.equal({})
+    await npm.getPeerDependencies('ncu-test-peer', '1.0.0').should.eventually.deep.equal({
       'ncu-test-return-version': '1.x',
     })
   })
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -298,7 +298,7 @@ const npmConfig = findNpmConfig()
  * @param data
  * @returns
  */
-function parseJson<R>(result: string, data: { command?: string; packageName?: string }): R {
+export function parseJson<R>(result: string, data: { command?: string; packageName?: string }): R {
   let json
   try {
     json = JSON.parse(result)
--- a/src/package-managers/yarn.ts
+++ b/src/package-managers/yarn.ts
@@ -181,6 +181,8 @@ function parseJsonLines(result: string): Promise<{ dependencies: Index<ParsedDep
   })
 }
 
+const cmd = process.platform === 'win32' ? 'yarn.cmd' : 'yarn'
+
 /**
  * Spawn yarn requires a different command on Windows.
  *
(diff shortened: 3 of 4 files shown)


test/package-managers/npm/index.test.ts
```diff
@@ -26,8 +26,8 @@ describe('npm', function () {
   })
 
   it('getPeerDependencies', async () => {
-    await npm.getPeerDependencies('ncu-test-return-version', '1.0').should.eventually.deep.equal({})
-    await npm.getPeerDependencies('ncu-test-peer', '1.0').should.eventually.deep.equal({
+    await npm.getPeerDependencies('ncu-test-return-version', '1.0.0').should.eventually.deep.equal({})
```
rbnayax, line 29: @raineorshine as far as I understand, this method will only be called with specific versions.
If that is not the case, please let me know as the yarn version doesn't support specs currently

rbnayax:
@raineorshine would appreciate your review :-)

test/package-managers/npm/index.test.ts
raineorshine: `getPeerDependenciesFromRegistry` is presumed to only get exact versions numbers:

https://github.com/raineorshine/npm-check-updates/blob/264b7e600db157c689ac7a167932a7e12dab2209/src/lib/getPeerDependenciesFromRegistry.ts#L51

However, `runLocal` incorrectly passes a version spec:

https://github.com/raineorshine/npm-check-updates/blob/264b7e600db157c689ac7a167932a7e12dab2209/src/lib/runLocal.ts#L195

I'm not actually sure what npm's behavior is when given a spec. Does it get the peer dependencies at the bottom of the range? Top of the range?

Open to your suggestion of how to best handle this.

rbnayax: From docs at https://docs.npmjs.com/cli/v7/commands/npm-view
```
If the version range matches multiple versions then each printed value will be prefixed with the version it applies to.
```
Running this locally with the json flag returns many json objects, for each version in the spec.
Running it in the code, this returns an array.
In case the sepc is only resolved to a single version (like in the test) this will work as it will not return an array.
So seems like this should not work at all if it is given a spec that resolves to multiple versions (but it will not fail)
I think that we need to fix the runLocal behaviour, thoughts?

rbnayax: yarn v2 works by giving latest version of the spec
we can make the same behavior for npm where we take the last element from the array, if such exists
the issue lies with yarn v1 which doesn't support spec at all. Let me check if that can be handled

raineorshine: Ah, never mind. It was staring me right in the face; it already converts to `minVersion` of the range, that's why it works. We don't need to change anything.

https://github.com/raineorshine/npm-check-updates/blob/264b7e600db157c689ac7a167932a7e12dab2209/src/lib/runLocal.ts#L201

The behavior is indeterminate of the spec is not a valid version range, but there's no need to address that at this point.

raineorshine (approved):

test/package-managers/npm/index.test.ts
rbnayax: I see but we have that in another place:
https://github.com/raineorshine/npm-check-updates/blob/264b7e600db157c689ac7a167932a7e12dab2209/src/lib/getIgnoredUpgradesDueToPeerDeps.ts#L27
Which as far as I understand does provide a spec. Should I fix that with same mechanism of minVersion?

rbnayax: haha Ok on it in another PR

PR #1472, opened 2024-10-27
fix: yarn getPeerDependencies doesnt work on private reporsitories wh…

rbnayax:
…en credentials are only set for yarn and not npm #1470

fix: #1470

completion of #1471


rbnayax:
@raineorshine would appreciate your review :-)

raineorshine (approved):

Issue #1414, opened 2024-05-22
interactive CLI has now quit or exit command.


vassudanagunta:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)

---

Currently the only way, AFAIK, is to issue a terminal kill command.

Ideally `q`  or `x` to exit with nothing upgraded or changed.

If agreed, I can take a stab at a PR when I get some free time.

raineorshine:
Thanks for the suggestion. I'd be happy with `q` for quit.

ZaLiTHkA:
just thinking out loud here, but perhaps this could be "fixed" by simply updating the input desrciptions for interactive mode?

pressing "escape" cancels the interactive mode prompt, returning the user to their terminal, as (I imagine) the OP would expect `q` or `x` to do.

vassudanagunta:
@ZaLiTHkA Good to know. But not very obvious or intuitive: In my experience, `ESC` is a standard idiom for cancellation in GUI apps. As far as I can remember, all the CLIs I've used exit on `q` or `x`. Either works for me, but I would suggest that it match most user expectations / CLI conventions.

ZaLiTHkA:
@vassudanagunta I agree with you on that point, it's definitely not very "intuitive" as it is right now.

on the other hand, the fact that there is already a mechanism in place to "cancel" interactive mode also means that adding support for a `q` or `x` shortcut to do the same thing _shouldn't_ need entirely new functionality, just a small adjustment to which key actually triggers it.

either way, a short message in the initial usage notes would clear this up for everyone.

PR #1476, opened 2024-11-12
Interactive page size and hint how to quit

edlerd:
# Done
- improve interactive mode to have dynamic page size. This helps if the terminal is very small
- add hint to interactive mode, that you can cancel with ESC key to address #1414

### DIFF
--- a/src/lib/runLocal.ts
+++ b/src/lib/runLocal.ts
@@ -106,13 +106,17 @@ const chooseUpgrades = async (
         ]
       })
 
+      const optionsPerPage = process.stdout.rows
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Thanks for your contribution!

The dynamic page size looks good.

Thank you for the suggestion for the interactive mode help, although I think in this case I would like to keep the instructions minimal and not have to print something that should be fairly obvious to most people. Interactive mode can be cancelled with Escape or Ctrl + C which are both highly conventional, and per #1414 I would be open to the addition of `q` to quit. Again just a personal preference to keep the instructions as terse as possible.

PR #1498, opened 2025-02-23
fix: readonly arrays on input options

Zamiell:
Today I discovered that npm-check-updates would not take in my immutable filterVersion array. So, I fixed it in this PR, as well as the surrounding arrays that were also incorrectly marked as mutable.

More info: In TypeScript libraries it is idiomatic to take in read-only data structures. That way, end-users do not have to cast their immutable data structures to be mutable simply to satisfy the TypeScript compiler (which is both ugly and reduces safety in using the library).

In case it isn't obvious, this PR is backwards compatible in that end-users can now use either mutable OR immutable arrays.

### DIFF
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -615,7 +615,7 @@ const cliOptions: CLIOption[] = [
       'Check one or more sections of dependencies only: dev, optional, peer, prod, or packageManager (comma-delimited).',
     default: ['prod', 'dev', 'optional', 'packageManager'],
     parse: value => (value && typeof value === 'string' ? value.split(',') : value),
-    type: 'string | string[]',
+    type: 'string | readonly string[]',
   },
   {
     long: 'deprecated',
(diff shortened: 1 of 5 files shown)


raineorshine (changes requested):
Thanks, seems fine!

Despite its location in `/src`, the file `RunOptions.ts` is actually the output of an automated process that generates the types and documentation for the API. See this note at the top of the file (which is not entirely obvious):

> /** This file is generated automatically from the options specified in /src/cli-options.ts. Do not edit manually. Run "npm run build" or "npm run build:options" to build. */

So instead of modifying `RunOptions.ts` directly, you will need to change https://github.com/raineorshine/npm-check-updates/blob/main/src/cli-options.ts and possibly https://github.com/raineorshine/npm-check-updates/blob/main/src/scripts/build-options.ts and then run `build:options` to achieve the desired change.

Zamiell:
Makes sense, please review the updated PR.

raineorshine (approved):

Issue #1490, opened 2025-01-16
Getting error while running npc in interactive mode


Ajin998:
<img width="1037" alt="Screenshot 2025-01-16 at 3 21 32 PM" src="https://github.com/user-attachments/assets/85122d25-9ccc-4e3e-9b5c-e6795d940b57" />

command executed:- `ncu -i`

raineorshine:
Hi, thanks for reporting.

Please fill out the full bug report. I don't currently have enough information to reproduce the error.

PR #1500, opened 2025-03-06
fix: Issue templates are not loading on Github

Deanmv:
Github has updated their [issue templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) and the existing one no longer loads when creating an issue.  This is a copy of the current one but into the new format and file location 

It looks like this is what happened here https://github.com/raineorshine/npm-check-updates/issues/1490


raineorshine:
Thank you!

PR #1502, opened 2025-03-08
Replace `remote-git-tags` with micro-optimized implementation.

Torathion:
I've come to notice that [`remote-git-tags`](https://www.npmjs.com/package/remote-git-tags?activeTab=readme) is an old package with statically declining weekly downloads. This might show that people are moving on from this package, either because it's not needed anymore or the packages using this are maintaining their own versions of it.

Since this package is a [very small](https://github.com/sindresorhus/remote-git-tags/blob/main/index.js) utility function with 22 lines code and fairly unoptimized, I decided to throw in my own micro-optimized version of it. This also gave me the chance to optimize the entire function of `getSortedVersions`.

This change reduces the bundle size by ~110 bytes and removes two dependencies.

I also tried to replace the ancient package [`parse-github-url`](https://www.npmjs.com/package/parse-github-url) with a newer, more stable package called [`git-url-parse`](https://www.npmjs.com/package/git-url-parse) that adds support for a ton of other git hosts and sources like azure, gists and gitlab, but it doesn't return the branch name. I've opened an [issue](https://github.com/IonicaBizau/git-url-parse/issues/176) on there, so there could still be a possibility of migrating to it.


raineorshine:
Hi, thanks for the contribution. 

Replacing `remote-git-tags` seems reasonable and I'm open to that. 

Replacing the functional programming style with procedural programming is out of scope for this project. Please follow the existing conventions, avoid mutations, use map, filter, reduce, etc and I will be happy to consider your submission.

Torathion:
I'm usually extremely against using array comfort functions as they cripple the performance of the package a lot, but I'll change the loop back to the original state.

raineorshine:
I respect that, but it's not the philosophy of this project. 

If you find a measurable performance issue, I'm happy to investigate. Typically it loops through only a few hundred packages/tags, which is significantly outweighed by network latency and download times.

Torathion:
If there is anything I should change, just mark it in your review.. I don't use functional programming, so I'm not familiar with it.

src/package-managers/gitTags.ts
```diff
@@ -1,53 +1,72 @@
...
-    tagsPromise = tagsPromise
-      .then(() => remoteGitTags(`ssh://git@${host}/${path?.replace(/^:/, '')}`))
-      .catch(() => remoteGitTags(`https://${auth ? auth + '@' : ''}${host}/${path}`))
-  }
+  let tags: Index<string>
 
-  // fetch remote tags
   try {
-    tagMap = await tagsPromise
+    if (protocol !== null) {
```
raineorshine, line 45: Let's preserve the functionality of `protocol != null` in order to include `undefined`.

raineorshine (changes requested):
Looks good, thanks!

src/package-managers/gitTags.ts
Torathion: `parse-github-url` is a extension to the deprecated utility `url.parse` from node.js It strictly uses `null` and not `undefined`.

You can test it yourself:

```js
import url from 'node:url'

console.log(url.parse('/')
```

That should return this:

```powershell
Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: '/',
  path: '/',
  href: '/'
}
```

Torathion: I could change it to `!protocol` since it doesn't seem to return empty strings.

PR #1507, opened 2025-04-07
remove incorrect comment about filterResults example

alberto:
`--target semver` filters **out** major version changes, while the example does the opposite, it lists only major version changes.


raineorshine (commented):
Thank you for catching this!

I have decided to correct the code example rather than changing the comment. Also changed the "filters out" language to hopefully be a bit clearer.

PR #1508, opened 2025-04-09
Add hideHint option for short terminals

clementallen:
When using `npm-check-updates` in short terminals (VS Code for example), the interactive mode hint takes up a lot of the terminal space leaving very little room for the scrolling list of available package updates.

`--hideHint` hides the interactive hint leaving more space to view available updates.

### Without (current behaviour)

image

### With `--hideHint`

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,10 @@ Options that take no arguments can be negated by prefixing them with `--no-`, e.
     <td>Customize how packages are divided into groups when using <code>--format group</code>.</td>
   </tr>
   <tr>
+    <td>--hideHint</td>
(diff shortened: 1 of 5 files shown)


clementallen:
Thanks for the feedback @raineorshine, agree that using `process.stdout.rows` to dynamically control hint displaying is a better approach.

I've now implemented only showing the hint when there are `> 18` rows which is a little bit higher than the default in VS Code and I think gives enough rows when larger than this.  Welcome for any thoughts on this value!

raineorshine (approved):
Looks good, thanks!

Issue #1515, opened 2025-04-23
ncu 18 uses Node/npm, even when installed via Bun


GingerAdonis:
- [X] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)

## Steps to Reproduce
- Install Node on Linux *without* npm.
- Install Bun on Linux.
- Globally install `npm-check-updates` via `bun install -g npm-check-updates`
- Run `ncu -g`

## Current Behavior

```
node:events:496
      throw er; // Unhandled 'error' event
      ^

Error: spawn npm ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess._handle.onexit (node:internal/child_process:291:12)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn npm',
  path: 'npm',
  spawnargs: [ '--global', '--json', 'ls', '--depth=0' ]
}

Node.js v22.14.0
```

## Expected Behavior
ncu is executed using Bun and thus recognizes it shouldn't use `npm`.

## Workaround
Command: `ncu -g --packageManager bun`

raineorshine:
Hi, thanks for reporting.

@LuisFerLCC It looks like #1514 had a regression. Could you look into this?

LuisFerLCC:
I think I already know what the problem is. Sadly, bun doesn't have a good way to test `bunx` with local packages, so I'll see what I can do later today.

LuisFerLCC:
So it was **not** what I expected (that bun had a `userAgent` check but not an `exepath` check). More details about the actual reason in the PR description.

#1516

GingerAdonis:
Thank you for the quick replies and PR.

XhmikosR:
This was fixed by #1516.

PR #1516, opened 2025-04-23
Check __dirname when getting running package manager

LuisFerLCC:
# Fixes
* #1515

## Explanation

When running a global package binary without `bunx`, bun runs the corresponding JavaScript file **directly**, without adding any of the environment variables that are usually added (yeah, I didn't know this). Yarn and PNPM do the same thing.

However, we can check the `__dirname` to see where the package is installed, and thus determine which package manager is being used.


GingerAdonis:
Is this a Bun bug in a way? I assume running a package via bunx or directly should result in the same env.

LuisFerLCC:
> Is this a Bun bug in a way?

It's not really a bug, it's a decision that all JavaScript package managers took. Because again, this also happens with npm, yarn and pnpm, and they were clearly very intentional about this (if you look up the path that the `ncu` command redirects to after installing the package globally, it's the JavaScript file for the package CLI, which has a "shebang" at the top telling the system that it must be run with vanilla Node.js).

Personally, I think this decision to execute the files directly with no env is just wrong. It does make things slightly faster, but also very inconsistent. I literally just figured this out **today**. I had no idea that package managers were this lazy.

raineorshine (approved):
Thanks @LuisFerLCC!

PR #1551, opened 2025-09-18
feat: add defineConfig helper

danielpza:
Add `defineConfig` helper similar to vite's and eslint's defineConfig 

- https://vite.dev/config/#config-intellisense
- https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file

This provides better intellisense support when defining a custom configuration in a ncurc config file.


README.md
```diff
@@ -818,10 +818,12 @@ Some options offer more advanced configuration using a function definition. Thes
 For example, `.ncurc.js`:
 
 ```js
-module.exports = {
```
raineorshine, line 821: Let's add this as a separate example so that users know its optional.

raineorshine (changes requested):
Looks reasonable, thanks!

raineorshine (approved):

PR #1553, opened 2025-09-18
fix(cooldown): support newest, minor, patch, and semver targets

SebastianSedzik:
## Issue
In the current implementation, the cooldown parameter functions only with the latest and greatest targets. For other targets (`newest`, `minor`, `patch`, or `semver`), it is broken - when used together, no updates are ever applied - despite being listed as supported in the documentation:

```
With other targets:

$ ncu --cooldown 5 --target greatest|newest|minor|patch|semver

Each target will select the best version that is at least 5 days old:

    greatest → 1.2.0        (highest version number outside cooldown)
    newest   → 2.0.0-beta.1 (most recently published version outside cooldown)
    minor    → 1.2.0        (highest minor version outside cooldown)
    patch    → 1.1.1        (highest patch version outside cooldown)
```
> [source](https://github.com/raineorshine/npm-check-updates#cooldown)

This PR is a fix for supporting rest of the targets

### DIFF
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -27,7 +27,7 @@ import { SpawnPleaseOptions } from '../types/SpawnPleaseOptions'
 import { Version } from '../types/Version'
 import { VersionResult } from '../types/VersionResult'
 import { VersionSpec } from '../types/VersionSpec'
-import { filterPredicate, satisfiesNodeEngine } from './filters'
+import { filterPredicate, satisfiesNodeEngine, satisfiesCooldownPeriod } from './filters'
 
 const EXPLICIT_RANGE_OPS = new Set(['-', '||', '&&', '<', '<=', '>', '>='])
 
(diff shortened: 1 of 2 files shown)


src/package-managers/npm.ts
```diff
@@ -918,6 +918,19 @@ export const newest: GetVersion = async (
...
 
+  if (options.cooldown) {
+    const versionsSatisfiesfyingCooldownPeriod = versionsSortedByTime.filter(version =>
+      satisfiesCooldownPeriod(
+        decorateTagPackumentWithTime((result as Packument).versions[version], result as Packument),
+        options.cooldown,
+      ),
+    )
+
+    console.log({ versionsSatisfiesfyingCooldownPeriod, result })
```
raineorshine, line 929: `console.log`

raineorshine (changes requested):
Good catch, thanks!

jerry23 (approved):

PR #1554, opened 2025-09-20
Enhance performance of get peer deps

rbnayax:
We have expeiricned degregaded performance for a while now, and I finally got around to investingating it. Appreantly, the get peers deps did not have any conccurency controls, and when you have 150 packages, it spwans 150 processes at the same time, causing my CPU to choke for a good minute.
I fixed that and also added a caching mechanism

### DIFF
--- a/src/lib/getPeerDependenciesFromRegistry.ts
+++ b/src/lib/getPeerDependenciesFromRegistry.ts
@@ -61,34 +61,41 @@ async function getPeerDependenciesFromRegistry(packageMap: Index<Version>, optio
   }
 
   const packageEntries = Object.entries(packageMap)
-  const accum: Index<Index<string>> = {}
 
   /**
-   * Processes peer dependencies for a package and updates the accumulator
+   * Fetches peer dependencies for a package
    * @param pkg - The package name
    * @param version - The package version
-   * @returns Promise that resolves when peer dependencies are processed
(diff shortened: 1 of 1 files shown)


src/lib/getPeerDependenciesFromRegistry.ts
```diff
@@ -59,19 +60,40 @@ async function getPeerDependenciesFromRegistry(packageMap: Index<Version>, optio
...
+  const packageEntries = Object.entries(packageMap)
+  const accum: Index<Index<string>> = {}
+
+  /**
+   * Processes peer dependencies for a package and updates the accumulator
+   * @param pkg - The package name
+   * @param version - The package version
+   * @returns Promise that resolves when peer dependencies are processed
+   */
+  const processPeerDependencies = async ([pkg, version]: [string, Version]) => {
```
raineorshine, line 72: `processPeerDependencies` currently relies on a mutation side effect to return its results. Using a typical return statement so that this is a pure function with no side effects would be preferable.

raineorshine (changes requested):
Thank you! This is a great improvement.

raineorshine (approved):

rbnayax:
@raineorshine when can we expect a release of this fix?

raineorshine:
Published to `v18.3.1`!

PR #1555, opened 2025-09-20
add support for skipping pinned vesrions

rbnayax:
If someone pinned a version, they might not want to update it. In that case, if using a new flag, it will be skipped

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,10 @@ Options that take no arguments can be negated by prefixing them with `--no-`, e.
     <td>Don't output anything. Alias for <code>--loglevel</code> silent.</td>
   </tr>
   <tr>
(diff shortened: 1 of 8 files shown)


rbnayax:
I agree that it can be achieved by that, and we are already using it (filterResults actually) but this requires from the user to write custom code and use the js config. I believe that this feature should be easier to achieve via a flag. The filter logic should be used for more complex logic

raineorshine:
That's reasonable, but ultimately it's not in alignment with the design philosophy of npm-check-updates. We already have a ton of options, so I'm not willing to add another option for something that can be accomplished with a custom filter function. 

We also have `--target semver` which respects pinned versions.

PR #1568, opened 2025-10-20
add interactive mode keyboard controls to docs

karladler:
Makes it better available to users that do not read the CLI comments (like me) but also makes this information available for search engines and LLMs.


lb0925788-sudo:
1569

Issue #1576, opened 2025-12-01
Add support for yarn catalog: feature


cctidal:
Similar to pnpm catalog: functionality, yarn has added support for the catalog: protocol https://yarnpkg.com/features/catalogs.

In yarn, the catalog configuration is stored in the .yarnrc.yml config file.

Examples:

```
// default catalog
catalog:
  lodash: ^4.17.21

// named catalogs
catalogs:
  react18:
    react: ^18.3.1
    react-dom: ^18.3.1
  react17:
    react: ^17.0.2
    react-dom: ^17.0.2
```

MKruschke:
@raineorshine will check how it was done for pnpm and yes that shouldn't be much different except where the catalog is configured.

MKruschke:
@raineorshine please have a look on the PR #1582

PR #1582, opened 2025-12-26
feat: support yarn catalogs

MKruschke:
see #1576 

additional fixed the unwanted behaviour to create an catalog entry if a named catalog `default` is present

### DIFF
--- a/src/lib/upgradePackageData.ts
+++ b/src/lib/upgradePackageData.ts
@@ -101,11 +101,6 @@ async function upgradePackageData(
           }
         }
 
-        // For pnpm, also expose the 'default' catalog as a top-level 'catalog' property
-        if (yamlData.catalogs?.default) {
-          yamlData.catalog = yamlData.catalogs.default
-        }
-
         return JSON.stringify(yamlData, null, 2)
       }
 
--- a/test/workspaces.test.ts
+++ b/test/workspaces.test.ts
@@ -550,7 +550,6 @@ catalogs:
           output.should.deep.equal({
             'pnpm-workspace.yaml': {
               packages: ['packages/**'],
-              catalog: { 'ncu-test-v2': '2.0.0' },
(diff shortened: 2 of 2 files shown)


src/lib/upgradePackageData.ts
```diff
@@ -100,11 +101,6 @@ async function upgradePackageData(
           }
         }
 
-        // For pnpm, also expose the 'default' catalog as a top-level 'catalog' property
-        if (yamlData.catalogs?.default) {
```
MKruschke: this doesn't make really sense to me and would create an catalog entry which wasn't defined before (see also workspace.test.ts)

test/workspaces.test.ts
```diff
@@ -550,7 +550,6 @@ catalogs:
           output.should.deep.equal({
             'pnpm-workspace.yaml': {
               packages: ['packages/**'],
-              catalog: { 'ncu-test-v2': '2.0.0' },
```
MKruschke: this shouldn't happen because `catalog` wasn't defined in the config before so it shouldn't be created by ncu

src/lib/upgradePackageData.ts
MKruschke: Hi @raineorshine, Hi @jakeboone02 ,
yes I'm pretty sure that this is the wrong behaviour and needs to be removed. Would even classify it as bug and not as breaking change.

### Details: 
For everyone that explicit is using a named catalog with the name `default` (e.g. `catalogs.default` it will also break the usage of pnpm. (see provide screenshot) because it will automatically introduce the default catalog `catalog` (in the current ncu implementation called singular catalog) after running ncu. I tested it with v10.13.1 as well as v10.26.2 of pnpm.

image

MKruschke: @raineorshine I hope the explanation above is sufficient enough and we can going forward with the PR.

raineorshine: In the absence of any objections, I'm fine to proceed.

Issue #1540, opened 2025-09-08
Overwrites pnpm-workspace.yaml into json


zhm:
I see there was support added for pnpm workspaces, which is awesome. A couple behaviors that were unexpected for me:

- It overwrites `pnpm-workspace.yaml` and I end up with a `pnpm-workspace.yaml` that's actually a JSON file. Is this intentional?
- It doesn't seem to respect "named catalogs".

Example, I have this at the bottom of my `pnpm-workspace.yaml` for an expo/mobile project where I want these dependencies. In package.json I use `catalog:mobile`, and other projects like the web app I just use `catalog:`.

```
catalogs:
  mobile:
    query-string: 7.1.3
    tailwindcss: ^3.4.17
```

But it upgrades those dependencies anyways. (e.g. upgrades tailwind to v4)

jakeboone02:
Right, I definitely want to make the fewest changes possible to the file. My question was whether you were ok with me replacing the `js-yaml` package with `yaml`. I'll go ahead and do it in my PR to fix this issue and you can review it then.

raineorshine:
Yes, I'm okay with that.

jakeboone02:
@raineorshine I've been working on this off and on, but could we possibly have a call sometime to work through the logic? I'm struggling a little with determining the right strategy. Probably just need another pair of eyes on it. We can address these, too:

- #1543
- #1560

MKruschke:
Hey @raineorshine, 
please have a look on the fix https://github.com/raineorshine/npm-check-updates/pull/1589

XhmikosR:
Is this issue still valid with the latest versions?

afonsojramos:
Retested with npm-check-updates 22.2.9 (Node 24, pnpm detected via lockfile). `ncu -u -w` now keeps `pnpm-workspace.yaml` as YAML and updates both the default catalog and named catalogs in place. Inline comments, section comments, and blank lines are preserved.

`pnpm-workspace.yaml`

```yaml
# workspace config
packages:
  - "." # root pkg

# shared versions
catalog:
  eslint: 9.0.0 # keep in sync with CI

catalogs:
  mobile:
    ncu-test-v2: 1.0.0
```

`package.json`

```json
{
  "name": "t",
  "version": "1.0.0",
  "devDependencies": {
    "eslint": "catalog:",
    "ncu-test-v2": "catalog:mobile"
  }
}
```

Diff after `ncu -u -w`:

```diff
-# workspace config
[...]

PR #1589, opened 2026-01-16
fix: fix overwrites pnpm-workspace.yaml and .yarnrc.yml into json

MKruschke:
- [x] add more tests

Refs: #1540

### DIFF
--- a/src/lib/upgradeYamlCatalogDependencies.ts
+++ b/src/lib/upgradeYamlCatalogDependencies.ts
@@ -78,7 +78,7 @@ function changeDependencyIn(
 }
 
 /**
- * Updates a dependency version in a Yarn `catalog` or `catalogs` section.
+ * Updates a dependency version in a PNPM/Yarn `catalog` or `catalogs` section.
  *
  * The function parses the YAML, validates it against `CatalogsConfig`, and
  * applies the change through CST tokens to preserve original formatting (such


src/lib/upgradeCatalogData.ts
MKruschke: renamed to upgradeJsonCatalogDependencies.ts

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-  "editor.defaultFormatter": "esbenp.prettier-vscode"
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true
```
Zamiell: good change imo

src/lib/upgradeYamlCatalogDependencies.ts
```diff
@@ -0,0 +1,113 @@
+import type { Document } from 'yaml'
+import { CST, isCollection, isPair, isScalar, parseDocument } from 'yaml'
+import { CatalogsConfig } from '../types/CatalogConfig'
+
+type UpdateDependencyConfig = {
+  path: string[] // e.g., ['catalogs', 'my-catalog', 'my-dep'] or ['catalog', 'my-dep']
+  newValue: string // e.g., '^2.0.0'
+}
```
raineorshine: I would prefer to inline this type it is only used in one place, then we get the benefits of co-location.

raineorshine (changes requested):
Thank you!

.vscode/settings.json
cctidal: imo this configuration shouldn't be here. Change this in your user configuration.

PR #1590, opened 2026-01-18
fix(tests): resolve EBUSY errors on Windows by retrying directory removal

Zamiell:
Back in August 2025, I added a helper function in 5054297b8f3dae8be29603d52e248a13a273856c to fix EBUSY errors on Windows. However, today I noticed it happening on other files!

Thus, I refactored the tests to all use this same helper function, and changed the logic of it slightly to use exponential backoff, which is more performant.


raineorshine (approved):
Thanks!

PR #1598, opened 2026-02-18
feat(cooldown): support cooldown strings like "7d", "12h", "30m"

aversini:
## Summary

Extends `--cooldown` to accept a duration string in addition to a plain number (days):

- `6` - 6 days
- `"6d"` — 6 days (same as the number `6`)
- `"12h"` — 12 hours
- `"30m"` — 30 minutes

Both CLI and module usage are supported:

```sh
ncu --cooldown 12h
```

```js
ncu({ cooldown: '30m' })
```

Plain numbers still work exactly as before (treated as days). Invalid strings (e.g. `"invalid"`) produce a clear error message.

## Implementation notes

- A small utility `src/lib/parseCooldown.ts` converts the string to a fractional number of days, keeping the existing `satisfiesCooldownPeriod` logic unchanged.
- `initOptions.ts` normalises string values to numbers early in the options processing pipeline, so all downstream code continues to receive a `number | CooldownFunction`.
- Auto-generated files (`RunOptions.ts`, `RunOptions.json`, `README.md`) were regenerated with `npm run build`.

## Test plan

- [x] All 28 cooldown unit tests pass (`npx mocha test/cooldown.test.ts`)
- [x] `npm run build` succeeds
- [x] `npm run lint` is clean
- [x] `npm test` passes (pre-existing pnpm/yarn/bun catalog failures are unrelated)
- [x] Manually verify: `ncu --cooldown 12h` in a real project

@raineorshine I took a quick stab at creating this PR so that I could have the same cooldown capabilities between `ncu` and `pnpm`. I personally use [pnpm minimumReleaseAge](https://pnpm.io/settings#minimumreleaseage) with a cooldown of 12 hours so it's valuable to have the possibility to configure `ncu` with the same value. I'm aware it may not fit the definition of "a balance between minimalism and customizability" but I thought it would not hurt since by default it works as before and as an opt-in, users can tweak a little the cooldown period. Thanks for the consideration and for this great project ✨

### DIFF
--- a/src/lib/parseCooldown.ts
+++ b/src/lib/parseCooldown.ts
@@ -2,7 +2,7 @@
  * Parses a cooldown string (e.g. "6d", "12h", "30m") into a number of days.
  * Returns `null` if the string does not match a valid format.
  */
-export function parseCooldownString(s: string): number | null {
+function parseCooldown(s: string): number | null {
   const match = s.match(/^(\d+(?:\.\d+)?)(d|h|m)$/)
   if (!match) return null
 
@@ -14,3 +14,5 @@ export function parseCooldownString(s: string): number | null {
   // unit === 'm'
   return value / (24 * 60)
 }
+
+export default parseCooldown
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -1,7 +1,7 @@
 import path from 'path'
 import { defaultCacheFile } from './lib/cache'
 import chalk from './lib/chalk'
-import { parseCooldownString } from './lib/parseCooldown'
+import parseCooldown from './lib/parseCooldown'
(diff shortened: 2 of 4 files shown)


src/lib/parseCooldown.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Parses a cooldown string (e.g. "6d", "12h", "30m") into a number of days.
+ * Returns `null` if the string does not match a valid format.
+ */
+export function parseCooldownString(s: string): number | null {
```
raineorshine: For single exports, I prefer a single, default export where the function name and filename exactly match. This convention makes it a bit easier to find when navigating the codebase.

Issue #1595, opened 2026-02-04
Incorrect assumption about `pnpm.cmd` extension on windows systems.


mynk-tmr:
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)

---

## Steps to Reproduce

- Use mise polyglot tool to install pnpm `mise use -g pnpm`
- run ncu on repo
- it fails in installation setup `pnpm.cmd is not a cmdlet .......`

## Logical Bug

If you use mise or scoop, pnpm is installed as binary and available as `pnpm` without extension. The fix is to try with `pnpm` first, if it fails try `pnpm.cmd`, then show error.

<img width="728" height="158" alt="Image" src="https://github.com/user-attachments/assets/063a580d-b411-49a4-b0cb-80f7e634cd3e" />

PR #1606, opened 2026-03-14
fix(pnpm): fallback from pnpm.cmd to pnpm on Windows for non-standard installs

terminalchai:
## Problem

On Windows, pnpm installed via **mise** or **scoop** does not create a .cmd shim in PATH. Spawning pnpm.cmd in these cases throws ENOENT, causing 
cu to fail entirely when using pnpm as the package manager.

Related issue: #1595

## Solution

Add a memoized getPnpmCmd() helper that:
1. On non-Windows: returns 'pnpm' immediately (no change)
2. On Windows: probes pnpm.cmd --version first
   - If successful → returns 'pnpm.cmd' (preserves existing behaviour for standard npm/pnpm global installs)
   - If ENOENT → returns 'pnpm' (fallback for mise, scoop, etc.)
   - Any other error (e.g. non-zero exit) → still returns 'pnpm.cmd' (pnpm exists, something else went wrong)

The result is memoized so the probe runs at most once per process.

Both list() and spawnPnpm() are updated to use wait getPnpmCmd() instead of the hardcoded ternary.

## Changes

- src/package-managers/pnpm.ts: add getPnpmCmd() helper, replace both hardcoded const cmd = process.platform === 'win32' ? 'pnpm.cmd' : 'pnpm' lines

### DIFF
--- a/src/package-managers/pnpm.ts
+++ b/src/package-managers/pnpm.ts
@@ -52,20 +52,28 @@ const npmConfigFromPnpmWorkspace = memoize(async (options: Options): Promise<Npm
 })
 
 /**
- * Resolves the pnpm command to use. On Windows, prefers `pnpm.cmd` (the shim created by npm
- * global installs) but falls back to `pnpm` for installs that do not create .cmd shims
- * (e.g. mise, scoop).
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Thanks!

PR #1654, opened 2026-04-07
fix: vite build warning

Zamiell:
When we run "npm run build", it displays the following warning:

> Entry module "src/index.ts" is using named and default exports together. Consumers of your bundle will have to use `chunk.default` to access the default export, which may not be what you want. Use `output.exports: "named"` to disable this warning.

This pull request fixes it.


raineorshine:
Thank you! 

This may be invalidated by #1649, which will be merged soon.

PR #1657, opened 2026-04-07
fix: e2e.sh on windows

Zamiell:
"lsof" does not exist on Windows, so this uses a Bash native solution - the special `$!` variable.

The CI tests passing should be proof that this is fine.

### DIFF
--- a/test/e2e.sh
+++ b/test/e2e.sh
@@ -12,6 +12,7 @@ registry_port=4873
 registry_local="http://localhost:$registry_port"
 registry_log=$temp_dir/verdaccio.log
 verdaccio_config=$temp_dir/verdaccio-config.yaml
+verdaccio_pid=""
 
 # cleanup on exit
 cleanup() {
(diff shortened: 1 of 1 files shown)


onemen:
I have something on this in #1649

https://github.com/raineorshine/npm-check-updates/pull/1649/changes#diff-f5b4d8aa705a53e6520988584a739326602a8ae71bcc84287402537193091274

Zamiell:
my fix is much better than what you did IMO, can you integrate it into your PR?

onemen:
> my fix is much better than what you did IMO, can you integrate it into your PR?

@raineorshine

What do you think?

raineorshine:
Unless there is another reason, this PRs change is preferable.

onemen:
see this commit in #1649:  https://github.com/raineorshine/npm-check-updates/pull/1649/commits/92f93ab87d0ce45f39b031fcb9209b3579fb882f

PR #1688, opened 2026-04-23
feat(cooldown): fall back to greatest passing version when latest is within cooldown

### LINKED ISSUE #1556: Cooldown does not upgrade to a newer version that is not latest

---

## Steps to Reproduce

Running `npx -y -- npm-check-updates --dep=dev,optional,peer,prod,packageManager --cooldown=2` does not update Svelte Kit to a newer version that is not the latest, as latest is not 2 days old.

The issue is with the --cooldown option: it currently only allows updating to the latest version, but only if the package hasn’t been updated in the last 2 days. It does not update to a version that is newer than the current one but older than the latest.

Dependencies:

```json
{
  "devDependencies": {
    "@sveltejs/kit": "2.22.5"
  },
  "packageManager": "pnpm@10.17.1",
  "volta": {
    "node": "24.8.0",
    "pnpm": "10.17.1"
  }
}
```

Steps:

Run `npx -y -- npm-check-updates --dep=dev,optional,peer,prod,packageManager --cooldown=2 -u`

## Current Behavior

Svelte Kit isn't updating to version older than latest but newer that the one specified in `package.json`.

## Expected Behavior

Svelte Kit should be updated to `2.43.2` if latest is `2.43.4` and not older than 2 days.

### LINKED ISSUE #1687: [duplicate] --cooldown silently drops the entire package when latest tag is within cooldown, instead of falling back to the most recent passing version

---

## Steps to Reproduce

.ncurc:

```js
{
     "cooldown": "7d"
}
```

Dependencies:

```json
{
   "dependencies": {
       "i18next": "^25.10.10"
   }
}
```

Steps:

1. Have a dependency (i18next) where the latest dist-tag points to a version published within the cooldown window, but a previous version exists that is older than the cooldown threshold.
2. Run:
   ```sh
     ncu --cooldown 7d --verbose
   ```

 ## Current Behavior

ncu resolves the latest dist-tag → 26.0.6, detects it is within the cooldown window, logs a skip message, and drops the package entirely from the report. No upgrade is suggested, even  though 26.0.5 is 8 days old and would pass the cooldown threshold.

 ```
   Skipping i18next@26.0.6 due to cooldown (published 2026-04-18T17:10:50.451Z).
 ```

 Result: i18next is not listed as upgradeable. A developer running ncu sees no actionable output for this package, creating a silent false negative.

 ## Expected Behavior

 When the latest dist-tag version is within the cooldown window, ncu should walk back through the version list to find the most recent version that passes the cooldown threshold and
 suggest that instead — the same way npm handles its own min-release-age setting.

 npm itself, when configured with min-release-age=7 in .npmrc, already demonstrates the correct behaviour. Under the same constraints npm outdated correctly reports 26.0.5 as the
 effective latest:

 ```
   Package  Current   Wanted  Latest
   i18next  25.10.10  26.0.5  26.0.5
 ```

Expected ncu output would be equivalent:

 ```
   i18next  ^25.10.10  →  ^26.0.5
 ```

marcosgilf:
## Summary

Closes #1687.

When `--cooldown` is set and the `latest` dist-tag version is within the cooldown window, `--target latest` (the default) silently drops the package entirely — even when older versions exist that pass the threshold. This is the documented behaviour with an existing warning, but it creates a silent false negative for users.

This PR introduces a new opt-in target **`aged`** that resolves this, mirroring how npm itself handles `min-release-age`:

1. Try the `latest` dist-tag version — if it passes cooldown, return it (identical to `--target latest`).
2. If the `latest` version is within the cooldown window, fall back to the **greatest** version across all published versions that passes the cooldown (and all other filters).
3. If no version passes, skip the package entirely.

`--target latest` is **unchanged** — this is purely additive.

## Behaviour comparison

| target | latest within cooldown | result |
|---|---|---|
| `latest` (existing) | yes | package dropped entirely |
| `aged` (new) | yes | falls back to greatest passing version |
| `greatest` (existing) | yes | greatest passing version (ignores dist-tag, includes pre-releases by default) |

`aged` differs from `greatest` in two important ways:
- **Prefers the dist-tag**: when `latest` is outside cooldown, `aged` returns it directly — it does not scan all versions.
- **No auto pre-releases**: `greatest` and `newest` automatically set `pre: true`; `aged` does not — it follows `latest` semantics.

## Changes

- `src/types/Target.ts` — added `'aged'` to `supportedVersionTargets`
- `src/package-managers/npm.ts` — implemented `aged: GetVersion`
- `src/package-managers/yarn.ts`, `pnpm.ts`, `bun.ts` — re-exported `aged`
- `src/types/PackageManager.ts`, `RunOptions.ts` — added `aged` to the interface and union type
- `src/lib/initOptions.ts` — added comment making the `autoPre` exclusion of `aged` explicit
- `src/cli-options.ts` — added `aged` to help table, updated cooldown docs note and warning to point to `aged` as the solution
- `README.md` — added `aged` row to the target table, updated cooldown section with `aged` example and updated the `latest`/`@tag` warning

## Tests

6 new tests in `test/cooldown.test.ts` under `when aged target`:

| Test | What it covers |
|---|---|
| upgrades to latest dist-tag when outside cooldown | Happy path — same as `latest` |
| falls back to most recent passing version when latest is within cooldown | Core new behaviour |
| skips entirely when all versions within cooldown | No false positives |
| does not return a pre-release as the fallback | `filterPredicate → allowPreOrIsNotPre` preserved |
| behaves identically to `latest` when cooldown is not set | Direct equivalence contract |
| respects the dist-tag, not the numerically greatest version | `aged` ≠ `greatest` without cooldown |

### DIFF
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -417,7 +417,7 @@ const extendedHelpTarget: ExtendedHelp = ({ markdown }) => {
       ],
       [
         'latest',
-        `Upgrade to whatever the package's "latest" git tag points to. Excludes prereleases unless --pre is specified.`,
+        `Upgrade to whatever the package's "latest" dist-tag points to. When used with --cooldown, falls back to the greatest version that passes the cooldown threshold if the latest is too recent. Use --target "@latest" for strict behaviour that skips the package instead. Excludes prereleases unless --pre is specified.`,
       ],
       ['minor', 'Upgrade to the highest minor version without bumping the major version.'],
       [
@@ -575,7 +575,7 @@ The value can be a plain number (days) or a string with a unit suffix:
     --cooldown 12h     12 hours
     --cooldown 30m     30 minutes
 
-Note that previous stable versions will ${chalk.bold('not')} be suggested. The package will be completely ignored if its latest published version is within the cooldown period. This is due to a limitation of the npm registry, which does not provide a way to query previous stable versions.
+With the default \`--target latest\`, if the latest dist-tag version is within the cooldown window, ncu falls back to the greatest version that passes the cooldown threshold. To instead skip the package entirely (strict behaviour), use \`--target "@latest"\`.
 
 ${chalk.bold('Example')}:
 
@@ -595,11 +595,20 @@ ${chalk.bold('With default target (latest)')}:
 
 ${codeBlock(`${chalk.cyan('$')} ncu --cooldown 5`, { markdown })}
 
+Falls back to 1.2.0 because:
+
+- Latest version (1.3.0) is only 4 days old (within 5-day cooldown)
+- 1.2.0 is the greatest version that is at least 5 days old
+
+${chalk.bold('With `@latest` strict target')}:
+
+${codeBlock(`${chalk.cyan('$')} ncu --cooldown 5 --target @latest`, { markdown })}
+
 No update will be suggested because:
 
-- Latest version (1.3.0) is only 4 days old.
+- Latest version (1.3.0) is only 4 days old
 - Cooldown requires versions to be at least 5 days old
-- Use \`--cooldown 4\` or lower to allow this update
+- \`@latest\` is strict: no fallback to older versions
 
 ${chalk.bold('With `@beta`/`@tag` target')}:
 
@@ -621,10 +630,6 @@ Each target will select the best version that is at least 5 days old:
     newest   → 2.0.0-beta.1 (most recently published version outside cooldown)
     minor    → 1.2.0        (highest minor version outside cooldown)
     patch    → 1.1.1        (highest patch version outside cooldown)
-
-${chalk.bold('Note for latest/tag targets')}:
-
-> :warning: For packages that update frequently (e.g. daily releases), using a long cooldown period (7+ days) with the default \`--target latest\` or \`--target @tag\` may prevent all updates since new versions will be published before older ones meet the cooldown requirement. Please consider this when setting your cooldown period.
 
 You can also provide a custom function in your .ncurc.js file or when importing npm-check-updates as a module.
 
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -916,7 +916,22 @@ export const latest: GetVersion = async (
   options: Options = {},
   npmConfig?: NpmConfig,
   npmConfigProject?: NpmConfig,
-) => distTag(packageName, currentVersion, { ...options, distTag: 'latest' }, npmConfig, npmConfigProject)
+): Promise<VersionResult> => {
+  const latestResult = await distTag(
+    packageName,
+    currentVersion,
+    { ...options, distTag: 'latest' },
+    npmConfig,
+    npmConfigProject,
+  )
+
+  if (latestResult.cooldown) {
+    const fallback = await greatest(packageName, currentVersion, options, npmConfig, npmConfigProject)
+    if (fallback.version) return fallback
+  }
+
+  return latestResult
+}
 
 /**
  * Fetches the most recently published version, regardless of version number.
--- a/test/cooldown.test.ts
+++ b/test/cooldown.test.ts
@@ -85,7 +85,6 @@ describe('cooldown', () => {
       const result = await ncu({
         packageData,
         cooldown: '6d',
-        target: 'latest',
       })
 
       expect(result).to.have.property('test-package', '1.1.0')
@@ -109,7 +108,6 @@ describe('cooldown', () => {
       const result = await ncu({
         packageData,
         cooldown: '6d',
-        target: 'latest',
       })
 
       expect(result).to.not.have.property('test-package')
@@ -134,7 +132,6 @@ describe('cooldown', () => {
       const result = await ncu({
         packageData,
         cooldown: '12h',
-        target: 'latest',
       })
 
       expect(result).to.have.property('test-package', '1.1.0')
@@ -159,7 +156,6 @@ describe('cooldown', () => {
       const result = await ncu({
         packageData,
         cooldown: '12h',
-        target: 'latest',
       })
 
       expect(result).to.not.have.property('test-package')
@@ -184,7 +180,6 @@ describe('cooldown', () => {
       const result = await ncu({
         packageData,
         cooldown: '30m',
-        target: 'latest',
       })
 
       expect(result).to.have.property('test-package', '1.1.0')
@@ -209,7 +204,6 @@ describe('cooldown', () => {
       const result = await ncu({
         packageData,
         cooldown: '30m',
-        target: 'latest',
       })
 
       expect(result).to.not.have.property('test-package')
@@ -232,7 +226,6 @@ describe('cooldown', () => {
       const resultNumber = await ncu({
         packageData,
         cooldown: 6,
-        target: 'latest',
       })
       stub1.restore()
 
@@ -240,7 +233,6 @@ describe('cooldown', () => {
       const resultString = await ncu({
         packageData,
         cooldown: '6d',
-        target: 'latest',
       })
       stub2.restore()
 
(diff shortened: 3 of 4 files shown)


marcosgilf:
I've created the `aged` tag to avoid changing the default behaviour if the issue is not considered a bug. What do you think about the tag name? could be `mature` or `outdated` to align with npm...

raineorshine (changes requested):
Thanks, this looks great.

I appreciate the conservative, opt-in approach. This is generally how contributions are considered for a mature library like npm-check-updates. However, after reviewing the discussion in #1556, I think the right move is to make this the default behavior of `--target latest`. npm-check-updates is meant to be simple and intuitive by default, and the recent cooldown behavior has been understandably confusing for users.

While the spirit of strict `--target latest` is correct, and we usually need to respect semver semantics, I think the benefits of falling back to filtered greatest versions when under cooldown outweigh the downsides. The downside is that it's possible to upgrade to a version that was never published to the `latest` tag, which can lead to unintended consequences with libraries that have unconventional yet semver-compatible publishing policies, such as even/odd publishing paradigms. But these libraries are not the norm. The vast majority of libraries always publish non-prerelease, non-beta versions to the `latest` tag. It seems better to build around the common, intuitive case, and shift the opt-in burden to the minority cases.

Therefore, I propose shifting the `aged` functionality to the default behavior (implicit or explicit `--target latest`) and use the strict behavior only with tag syntax: `--target "@latest"`. (`--target "@latest"` is nearly identical to `--target latest`, but it can *downgrade* a version if the package version is higher than the published latest. So there is already some precedence for it being more strict.)

Thanks for your assistance and look forward to hearing your thoughts.

marcosgilf:
Happy to help :) 

Makes total sense, I have updated the code to reflect that simpler approach.

src/package-managers/npm.ts
```diff
@@ -916,7 +916,31 @@ export const latest: GetVersion = async (
...
+  const latestResult = await distTag(
+    packageName,
+    currentVersion,
+    { ...options, distTag: 'latest' },
+    npmConfig,
+    npmConfigProject,
+  )
+
+  // latest passed all filters (including cooldown) — return it as-is
+  if (latestResult.version) return latestResult
```
raineorshine: Since `cooldown` and `version` are mutually exclusive, this is unnecessary as far as I can tell. The cooldown block will be skipped and the function will return `latestResult` by default.

src/package-managers/npm.ts
```diff
@@ -916,7 +916,31 @@ export const latest: GetVersion = async (
...
+
+  // latest dist-tag is within the cooldown window — walk back through all versions
+  // to find the greatest one that satisfies the cooldown (and all other filters).
+  // Use --target "@latest" for strict behaviour that drops the package instead.
+  if (latestResult.cooldown) {
+    const fallback = await greatest(packageName, currentVersion, options, npmConfig, npmConfigProject)
+    if (fallback.version) return fallback
+    // no version outside cooldown found — preserve the cooldown marker so callers
+    // (e.g. the "All dependencies not in cooldown" message) still count this package
+    return { cooldown: true }
```
raineorshine: I believe this is unnecessary as well

test/cooldown.test.ts
```diff
@@ -362,12 +390,64 @@ describe('cooldown', () => {
...
+
+      const result = await ncu({ packageData, cooldown, target: 'latest' })
+
+      // Then: pre-release is not suggested as fallback — latest never promotes pre-releases
       expect(result).to.not.have.property('test-package')
 
       stub.restore()
     })
 
+    it('respects the latest dist-tag and does not upgrade to a numerically higher non-latest version', async () => {
```
raineorshine: We already have coverage of default `--target latest behavior.

test/cooldown.test.ts
```diff
@@ -500,6 +580,59 @@ describe('cooldown', () => {
...
+
+      const result = await ncu({ packageData, cooldown, target: '@latest' })
+
+      // Then: returns the latest tag version directly
+      expect(result).to.have.property('test-package', '1.1.0')
+
+      stub.restore()
+    })
+
+    it('drops the package entirely when @latest is within cooldown, with no fallback', async () => {
```
raineorshine: Prefer "skip" or "ignore" rather than "drop".

test/cooldown.test.ts
```diff
@@ -500,6 +580,59 @@ describe('cooldown', () => {
...
+            '1.1.0': new Date(NOW - 5 * DAY).toISOString(),
+            '1.0.1': new Date(NOW - 15 * DAY).toISOString(),
+          },
+          distTags: { latest: '1.1.0' },
+        }),
+      )
+
+      const result = await ncu({ packageData, cooldown, target: '@latest' })
+
+      // Then: package is dropped entirely — no fallback to 1.0.1
```
raineorshine: Prefer "skip" or "ignore" rather than "drop".

test/cooldown.test.ts
raineorshine: Just noticing that this test doesn't make sense. `1.1.0` should be older than `1.2.0`. Would you mind fixing this while you're here?

raineorshine (changes requested):
Thank you!

src/cli-options.ts
raineorshine, line 1: Prefer "skip" or "ignore" rather than "drop".

raineorshine (changes requested):

src/package-managers/npm.ts
marcosgilf: Fixed!

src/package-managers/npm.ts
marcosgilf: You were right, fixed

test/cooldown.test.ts
marcosgilf: Removed

test/cooldown.test.ts
marcosgilf: Changed

test/cooldown.test.ts
marcosgilf: Removed comment, code was selfexplainatory

test/cooldown.test.ts
marcosgilf: Yeah, It is fixed now

src/cli-options.ts
marcosgilf: Changed

test/cooldown.test.ts
```diff
@@ -85,7 +85,6 @@ describe('cooldown', () => {
       const result = await ncu({
         packageData,
         cooldown: '6d',
-        target: 'latest',
```
marcosgilf, line 88: I think this can be removed to focus on the default behavior, more taking into account that later there is a latest describe. Thoughts?

marcosgilf:
I've reviewed the test file with my boy scout hat. I removed when, then comment to leave the code to take by itself, aligned `it` descriptions and removed the `target: "latest"` property in the firsts tests to focus on the default behavior.

onemen:
@raineorshine 

I think it’s best to show, by default, a list of packages filtered by cooldown, highlighted in yellow or at least with a yellow title

For example:

```
Filtered by 7 days cooldown
 eslint    10.1.0  →  10.2.1   (released 6 days ago)

 eslint    10.1.0  →  10.2.0  (released 20 days ago) (available eslint@10.2.1  released 6 days ago)
```

this will give the users all the information they need

I can do it in a separate PR if you prefer

test/cooldown.test.ts
```diff
@@ -299,10 +289,8 @@ describe('cooldown', () => {
       }),
     )
 
-    // When ncu is run with a 0 day cooldown parameter
     const result = await ncu({ packageData, cooldown })
 
-    // Then test-package should be upgraded to version 1.2.0 (latest) - as cooldown of 0 means no cooldown.
```
raineorshine, line 305: While tests are largely self-describing, these comments (and similar comments throughout the test file) add important context that helps explain the *why*. Therefore, I would prefer not to remove them.

raineorshine (changes requested):

raineorshine:
@onemen Thanks for the suggestion. Can you open an issue for this? There are some decisions to make around presentation. Likely we'll put the information under the `--format` option since the default output is intentionally minimal.

test/cooldown.test.ts
marcosgilf: Comments are back!

raineorshine (approved):
Looks good! Ready to merge once tests are passing.

raineorshine (approved):
Thank you!

PR #1693, opened 2026-04-26
Enhance `--format time`: consistent caching, interactive support, and human-readable timestamps

onemen:
### Description

This PR resolves several issues related to the display and persistence of package release timestamps in `npm-check-updates`.

Previously, when using `--format time --cache`, timestamps were only available during the initial network request. Because the existing cache implementation only stored the version string and omitted the release timestamp, subsequent runs using cached data resulted in missing time information. This PR also addresses the omission of timestamps in interactive mode and replaces raw ISO timestamps with a more readable format.

#### 1. Cache Schema Update & Invalidation (`cache.ts`)

Updated the caching logic to store an object `{ version, time }` and introduced a schema versioning system.

- **The Fix:** Added a `schema` version (v1) to the cache file.
- **Invalidation Mechanism:** Updated `checkCacheExpiration` to compare the cache's stored `schema` version with `CURRENT_CACHE_SCHEMA`. If they do not match, the function returns `true` (expired), forcing a clean refresh. This eliminates the need for legacy fallback logic.

#### 2. Consistent Metadata Retrieval (`npm.ts`)

Updated `npm.ts` functions to ensure that package release times are always bundled with version information. This provides a single source of truth, ensuring that downstream functions—including `filterResults`—always have access to timestamp metadata.

#### 3. Human-Readable Formatting

Integrated `timeago.js` to replace raw ISO strings with a human-readable format via `timeAgoFormat(timestamp, 'en_US')`, allowing developers to quickly identify how recently a dependency was published.

#### 4. Interactive Mode Support

Updated `chooseUpgrades` to include the `time` argument, ensuring it is correctly passed to `toDependencyTable` so that timestamps are displayed during the interactive selection process.

### DIFF
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -943,7 +943,7 @@ export const newest: GetVersion = async (
   npmConfig?: NpmConfig,
   npmConfigProject?: NpmConfig,
 ): Promise<VersionResult> => {
-  const result = await npmApi.fetchUpgradedPackumentMemo(
+  const packument = await npmApi.fetchUpgradedPackumentMemo(
     packageName,
     ['time', 'versions'],
     currentVersion,
@@ -957,7 +957,7 @@ export const newest: GetVersion = async (
   // result.versions is an object but is parsed as an array, so manually convert it to an object.
   // Otherwise keyValueBy will pass the predicate arguments in the wrong order.
   const versionsSatisfyingNodeEngine = keyValueBy(
-    result?.versions || {},
+    packument?.versions || {},
     (version: Version, packument: Packument['versions'][string]) =>
       satisfiesNodeEngine(packument, options.nodeEngineVersion) ? { [packument.version]: true } : null,
   )
@@ -965,7 +965,7 @@ export const newest: GetVersion = async (
   // filter out times that do not satisfy the node engine
   // filter out prereleases if pre:false (same as allowPreOrIsNotPre)
   const timesSatisfyingNodeEngine = filterObject(
-    (result?.time || {}) as Index<string>,
+    (packument?.time || {}) as Index<string>,
     version => versionsSatisfyingNodeEngine[version] && (options.pre !== false || !versionUtil.isPre(version)),
   )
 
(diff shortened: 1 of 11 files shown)


onemen:
### Before:

Screenshot 2026-04-26 184838


### After:
Screenshot 2026-04-26 184941

### Interactive mode with time:
Screenshot 2026-04-26 185008



</br></br></br>
@raineorshine , this PR is ready

raineorshine:
Thank you! Will review soon.

src/package-managers/npm.ts
```diff
@@ -711,11 +711,7 @@ export const greatest: GetVersion = async (
   npmConfig?: NpmConfig,
   npmConfigProject?: NpmConfig,
 ): Promise<VersionResult> => {
-  const fields: (keyof Packument)[] = ['versions']
-
-  if (options.cooldown) {
-    fields.push('time')
-  }
+  const fields: (keyof Packument)[] = ['time', 'versions']
```
raineorshine: We only want to request `time` as needed, otherwise it can slow down the request.

raineorshine (changes requested):
Thank you!

src/package-managers/npm.ts
onemen: I have updated the code to implement a 'lazy-load' approach that only fetches package timestamps when specifically required by the user's configuration. This ensures we avoid the performance overhead of full metadata requests during standard runs.

However, there is an issue with the current cache implementation regarding the `--cooldown` feature. The cache does not currently store any state or timestamps relevant to cooldown constraints. If a user runs `ncu --cache` and then follows it with `ncu --cache --cooldown 7`, the result would incorrectly treat the cooldown as 0 because the cached version was saved without that context.

I am planning to refactor the cache to be 'constraint-aware' to address this, but I recommend handling that refactor in a separate PR to keep the current scope manageable. For now, I've updated the logic to **automatically skip the cache whenever `--cooldown` is active** to ensure results remain accurate.

What do you think of this approach?

raineorshine: That sounds like a great approach. Thanks!

raineorshine (approved):

PR #1694, opened 2026-04-27
fix: make --dep and --format robust to PowerShell arg rewriting (comma-or-space-delimited)

onemen:
### **Summary**

PowerShell rewrites comma‑separated arguments when expanding `$args` inside package‑manager–generated `.ps1` shims (such as those created by pnpm).
This causes a user‑supplied value like:

```
--format time,group
```

to be transformed into:

```
--format time group
```

As a result, strict comma‑only parsing breaks for users invoking `ncu` through the shim, even though direct `node build/cli.js` execution works correctly.

```
Invalid option value: --format time group. Valid values are: dep, group, homepage, ownerChanged, repo, diff, time, lines, installedVersion.
```

### **Change**

To make the CLI resilient to this shell behavior, the parsing logic for `--format` and `--dep` has been updated:

```ts
// old
parse: value => (typeof value === 'string' ? value.split(',') : value),

// new
parse: value => (typeof value === 'string' ? value.split(/,|\s/) : value),
```

This allows both comma‑delimited and space‑delimited values to be interpreted consistently.

### **Notes**

This effectively standardizes both options as **comma‑or‑space‑delimited**, which:

- avoids PowerShell shim rewriting issues
- maintains backward compatibility
- improves UX by accepting intuitive input formats
- avoids the need for brittle, environment‑dependent tests

No additional tests are required for this change.

### DIFF
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -721,7 +721,7 @@ const cliOptions: CLIOption[] = [
     description:
       'Check one or more sections of dependencies only: dev, optional, peer, prod, or packageManager (comma-delimited).',
     default: ['prod', 'dev', 'optional', 'packageManager'],
-    parse: value => (value && typeof value === 'string' ? value.split(',') : value),
+    parse: value => (typeof value === 'string' ? value.split(/,|\s/) : value),
     type: 'string | readonly string[]',
   },
   {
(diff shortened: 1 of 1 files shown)


src/cli-options.ts
```diff
@@ -183,7 +183,7 @@ For the SemVer type definition, see: https://git.coolaj86.com/coolaj86/semver-ut
 /** Extended help for the --format option. */
 const extendedHelpFormat: ExtendedHelp = ({ markdown }) => {
   const header =
-    'Modify the output formatting or show additional information. Specify one or more comma-delimited values.'
+    'Modify the output formatting or show additional information. Specify one or more comma-or-space-delimited values.'
```
raineorshine: Let's leave the help documentation just saying comma-delimited as I don't want to deal with support tickets for shell quoting issues.

raineorshine (changes requested):
Hi, thanks for your contribution. Looks good.

PowerShell probably transforms it to something more like `ncu --format "time group"`. `ncu --format time group` is actually a positional argument that is equivalent to `ncu --format time --filter group`.

onemen:
> PowerShell probably transforms it to something more like `ncu --format "time group"`

You are right :)

src/cli-options.ts
onemen: done, all docs changes reverted back

raineorshine (approved):

PR #1711, opened 2026-05-03
feat: add cooldown format and improved visibility for skipped updates

### LINKED ISSUE #1690: Improve visibility of cooldown‑filtered updates and enhance timestamp formatting
@raineorshine, 

This issue is about defining behavior and options so I can open a PR. It focuses on:

- cooldown‑filtered updates (what is shown and how)
- timestamp formatting (`--format time`)
- interaction with cache

---

### 1.1 Current behavior

Given:

- installed: `eslint@10.1.0`
- available:
  - `eslint@10.2.0` (released 20 days ago)
  - `eslint@10.2.1` (released 6 days ago)

With a cooldown that filters out `10.2.1`, the output only shows, after PR #1688:

```
Patch — Backwards‑compatible bug fixes
  eslint    10.1.0  →  10.2.0
```

There is no indication that `10.2.1` exists or that it was filtered by cooldown.

If the user already has `eslint@10.2.0` installed, there is no information at all about the existence of `10.2.1`.

### 1.2 Proposed behavior

- Show a clear indication of:
  - which cooldown is in effect
  - which versions were filtered out because of it
- Optionally group or highlight cooldown‑filtered updates.

Example (illustrative only):

```text
Filtered by 7‑day cooldown
  eslint    10.1.0  →  10.2.1   (released 6 days ago)

Patch — Backwards‑compatible bug fixes
  eslint    10.1.0  →  10.2.0   (released 20 days ago)
  (latest available: eslint@10.2.1 released 6 days ago)
```

---

### **1.3.1 Cooldown visibility line**

Because `--cooldown` may come from CLI, config file, or package manager config, we need a consistent way to show the effective cooldown.

**Decisions:**

- When cooldown is active and **packages were filtered**
  - Show: `2 packages filtered by 7‑day cooldown`
  - Or: `Filtered by 7‑day cooldown`
  - Or: show only a section title without count
  - And/Or:  add the cooldown source -> `2 packages filtered by 7‑day cooldown (from npm config)`
- When cooldown is active and **no packages were filtered**
  - Show: `No packages were filtered by 7‑day cooldown` (possibly green)
  - Or show nothing
- Placement:
  - At the top of the output
  - At the end of the output
  - Only when packages were filtered (proposed default)

**Proposed default:**
Show `x packages filtered by <cooldown> <source>` **only when at least one package was filtered**.

---

### **1.3.2 Section title wording**

Options:

- `Filtered by 7‑day cooldown`
- `2 packages filtered by 7‑day cooldown`
- Compact mode: title only, no list
- Verbosity controlled by `--format cooldown:<mode>`

---

### **1.3.3 Highlighting**

Options:

- Yellow section title
- Yellow filtered package lines
- No color
- Optional: color only when packages were filtered

---

### **1.3.4 Release age formatting**

For `(released 6 days ago)`:

- Parentheses vs no parentheses
- With or without the word `released`
  - `(released 6 days ago)`
  - `(6 days ago)`
- Optional color

---

### **1.3.5 “Latest available” line**

For cases like:

```
eslint    10.1.0 → 10.2.0   (released 20 days ago)
(latest available: eslint@10.2.1 released 6 days ago)
```

Decisions:

- **One line vs two lines**
- **Wording**
  - `latest available:`
  - `latest:`
  - `(10.2.1 is available)` (pnpm‑style)
  - `→ 10.2.1 (6 days ago)
- **Package name vs version only**
  - `eslint@10.2.1`
  - `10.2.1`
- **Coloring**
  - Color the “latest available” part
  - Leave uncolored

---

# **1.3.6 Format control options**

Because `--cooldown` already exists, adding `--format cooldown` would be confusing.
We need a **single, clearly named format option** that controls how cooldown‑related information is displayed.

### **A. Name for the format flag**

Proposed:

- `--format filtered`
  Because it directly describes what the section shows: _filtered versions_.

---

### **B. What the default (no format flag) should show**

When the user runs:

```
ncu --cooldown
```

**Proposed default behavior:**

- Show **only** when packages were filtered
  Example:
  `2 packages filtered by 7‑day cooldown`
  we can add cooldown source
  `2 packages filtered by 7‑day cooldown (from npm config)`
- Show the **list of filtered packages**
- Show **release age** for filtered versions
- Do **not** show “latest available” lines
- Do **not** show color unless the project already uses color in similar contexts

This keeps the default output minimal but informative.

---

### **C. What `--format filtered` should add**

If the user explicitly opts in:

```
ncu --cooldown --format filtered
```

Then show **all** cooldown‑related details:

- Always show the cooldown line
  (`Filtered by 7‑day cooldown` or `No packages were filtered…`)
- Show filtered package list
- Show release age
- Show “latest available” line
- Allow color for:
  - section title
  - filtered items
  - “latest available” line
- Allow more verbose wording

This gives users full visibility when they want it.

onemen:
closes #1690 

## Technical Refactor: `npm.ts` & Cooldown Visibility

I have performed a major refactor of `npm.ts` to support the new `--format cooldown` option while improving overall performance and code maintainability.

### Key Changes:

*   **`findTargetAndFallback`**: Introduced a single-pass reducer that filters and resolves both the target upgrade version and a cooldown fallback simultaneously. By iterating through available versions only once, it avoids multiple loops for filtering and sorting, making the detection of "cooldown-safe" versions significantly more efficient.
*   **`toVersionResult`**: This new utility standardizes how raw resolution data is formatted into a `VersionResult` object. It wraps the target version with its publish time and, if a cooldown is active, attaches the fallback version and relevant metadata.
*   **Uniform Implementation**: With these two functions, the implementation across all `target` functions is now uniform, significantly reducing boilerplate and improving readability.

### PR Highlights:

*   **Enhanced Visibility**: Added `--format cooldown` to display a dedicated "Skipped due to x-day cooldown" section in the terminal, listing version changes and release ages.
*   **Comprehensive Testing**: Added new test cases to `cooldown.test.ts` to verify the logic and the new formatting output.
*   **Documentation**: 
    *   Updated the `--format` options table in the `README.md`.
    *   Added a detailed "Cooldown Formatting" section with usage examples to the `README.md` at the end of `cooldown` section.

@raineorshine, 
Currently, this new format logic does not apply when `options.json` is used. Please let me know if you would like me to extend this support to the JSON output as well.


onemen:
@raineorshine 

this PR is ready for review

src/cli-options.ts
```diff
@@ -796,11 +813,12 @@ const cliOptions: CLIOption[] = [
     long: 'format',
     arg: 'value',
     description:
-      'Modify the output formatting or show additional information. Specify one or more comma-delimited values: dep, group, ownerChanged, repo, time, lines, installedVersion.',
+      'Modify the output formatting or show additional information. Specify one or more comma-delimited values: dep, group, ownerChanged, repo, time, lines, installedVersion, cooldown.',
     parse: value => (typeof value === 'string' ? value.split(/,|\s/) : value),
     default: [],
     type: 'readonly string[]',
-    choices: ['dep', 'group', 'homepage', 'ownerChanged', 'repo', 'diff', 'time', 'lines', 'installedVersion'],
+    // prettier-ignore
```
raineorshine: No need to override prettier

src/lib/initOptions.ts
```diff
@@ -73,14 +76,25 @@ async function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } =
...
+  // save cooldown raw input
+  if (options.cooldown != null) {
+    if (typeof options.cooldown === 'object' && 'raw' in options.cooldown) {
+      const { raw, days } = options.cooldown as unknown as { raw: string; days: number }
+      options._rawcooldown = raw
+      options.cooldown = days
+    } else {
+      options._rawcooldown = options.cooldown
+    }
+  }
```
raineorshine: After reviewing the usage of `raw` and `_rawcooldown`, I would suggest a different approach. Instead of packaging `raw` with the individual parsed value, perhaps you could save all of the raw options on the base options object. Then access it anywhere with `options.raw.cooldown`. The [Options](https://github.com/raineorshine/npm-check-updates/blob/main/src/types/Options.ts) type already extends `RunOptions` with additional metadata, including `args` which is very similar. We could just add a `raw: RunOptions` field.

The CLI gets [programOpts](https://github.com/raineorshine/npm-check-updates/blob/8d98cb4b3045b88d341cc400728fe665d038438e/src/bin/cli.ts#L167) from commander and I think by the time they get to `initOptions` as [runOptions](https://github.com/raineorshine/npm-check-updates/blob/8d98cb4b3045b88d341cc400728fe665d038438e/src/lib/initOptions.ts#L76) you shouldn't have to worry about CLI vs programmatic invocation.

This approach has the benefit of:

- Parsing function doesn't need to worry about raw value.
- Avoids ` & { _rawcooldown?: Options['cooldown'] }` throughout params.
- Avoids ` as unknown as { raw: string; days: number }` cast.
- Future compatible for other parsed string types.

src/lib/logging.ts
```diff
@@ -155,20 +191,25 @@ function getVersion(dep: string): string {
...
   to: toDeps,
+  skippedByCooldown = [],
   format,
   ownersChangedDeps,
   pkgFile,
   time,
 }: {
   from: Index<VersionSpec>
   to: Index<VersionSpec>
+  skippedByCooldown?: CooldownInfo[]
```
raineorshine: For consistency, let's type this and others as `Index<CooldownInfo>`. We use `Index<VersionSpec>` everywhere else in the codebase, so using a new array format is a bit out of place.

src/lib/logging.ts
```diff
@@ -155,20 +191,25 @@ function getVersion(dep: string): string {
...
   format?: readonly string[]
   ownersChangedDeps?: Index<boolean>
   /** See: logging/getPackageRepo pkgFile param. */
   pkgFile?: string
   time?: Index<string>
 }) {
   const pkg = format?.includes('dep') && pkgFile ? JSON.parse(await fs.readFile(pkgFile, 'utf-8')) : null
+  const availableUpdates = format?.includes('cooldown')
+    ? Object.fromEntries(skippedByCooldown.map(({ name, version }) => [name, version]))
+    : null
```
raineorshine: This won't be necessary once we switch to `Index<CooldownInfo>`.

src/lib/logging.ts
```diff
@@ -205,12 +246,27 @@ export async function toDependencyTable({
             : ''
           const timestamp = format?.includes('time') && time?.[dep] ? time[dep] : null
           const publishTime = timestamp ? timeAgoFormat(timestamp, 'en_US') : ''
+
+          const availableVersion = availableUpdates?.[dep]
+          let available = ''
```
raineorshine: Prefer the word `cooldown` somewhere in this variable name for clarity.

src/lib/logging.ts
```diff
@@ -205,12 +246,27 @@ export async function toDependencyTable({
...
+            available = `[cooldown] ${skippedColorized.replace(wildcard, '')}`
+          }
+
           return [
             dep,
             ...(format?.includes('dep') ? [depType ? chalk.gray(depType) : ''] : []),
             from,
             '→',
             toColorized,
+            available,
```
raineorshine: An empty string will inadvertently leave a blank space in the table when `options.cooldown` is false. Instead we want to avoid rendering the column at all.

```js
...[available].filter(Boolean),
```

test/helpers/silenceProgressBar.ts
raineorshine, line 1: Filename should match function name

raineorshine (changes requested):
Thank you!

> Currently, this new format logic does not apply when `options.json` is used. Please let me know if you would like me to extend this support to the JSON output as well.

`--jsonAll`, `--jsonDeps`, and `--jsonUpgraded` strictly output subsets of the package.json file, so there is nowhere to stash the cooldown versions currently without a breaking change. I think you can ignore this for this PR and only add it if there is real demand.

src/cli-options.ts
onemen: done

src/lib/initOptions.ts
onemen: I have found much simpler solution.

since `initOptions.ts` call `parseCooldown` I have removed it from `cli-options.ts`

`cli-options.ts` now just validate the type of cooldown command line argument. 
`initOptions` convert the string to days

src/lib/logging.ts
onemen: done

src/lib/logging.ts
onemen: done

src/lib/logging.ts
onemen: done

src/lib/initOptions.ts
raineorshine: That still requires `& { _rawcooldown?: Options['cooldown'] }` to smuggle in the raw value. Having a uniform, general way to access raw input values is a cleaner solution imo.

onemen: Even if i add 
```ts
  // save raw values for all cli arguments
  const raw: Record<string, any> = {}
  cliOptions.forEach(({ long, short, arg, type }) => {
    const index = programArgs.findIndex(a => a === `--${long}` || (short && a === `-${short}`))

    if (index !== -1) {
      if (arg) {
        // It's an option that takes a value (like --cooldown 7d)
        raw[long] = programArgs[index + 1]
      } else if (type === 'boolean') {
        // It's a boolean flag (like --global)
        raw[long] = !programArgs[index].startsWith('--no-')
      } else {
        // Flag exists but has no defined arg (standalone flag)
        raw[long] = true
      }
    }
  })
```

```ts
const options = {
    ...(rcResult && Object.keys(rcResult.config).length > 0 ? { rcConfigPath: rcResult.filePath } : null),
    ...pickBy(program.opts(), (value: unknown) => value !== undefined),
    args: program.args,
    ...(combinedProgramOpts.filter ? { filter: combinedProgramOpts.filter } : null),
    ...(combinedProgramOpts.reject ? { reject: combinedProgramOpts.reject } : null),
    raw,
  }
```

that code only cover cli usage. if we want to cover `Module/Programmatic Usage` and getting options from .ncurc files we need also to check runOptions in initOptions again.

i have just forgot to add `_rawcooldown` type to Options 

```diff
export type Options = RunOptions & {
+  _rawcooldown?: RunOptions['cooldown']
  args?: any[]
  cacher?: Cacher
  cli?: boolean
  distTag?: string
  json?: boolean
  nodeEngineVersion?: VersionSpec
  packageData?: string
  peerDependencies?: Index<any>
  rcConfigPath?: string
  // A list of local workspace packages by name.
  // This is used to ignore local workspace packages when fetching new versions.
  workspacePackages?: string[]
}
```

I think this is the simplest solution.

raineorshine: I would prefer a cleaner, more general solution than making a special case for cooldown. Please reread my original feedback and let me know if you have any questions.

onemen: I hope now it is what you are looking for

onemen:
@raineorshine 

I think this is ready

raineorshine (approved):
Looks great, thanks!

PR #1738, opened 2026-05-17
fix: normalize blank-line spacing between packages in --deep mode

### LINKED ISSUE #956: Cosmetic suggestions to improve the logs
This is just a cosmetic suggestion when using `ncu` inside monorepo.

The current behavior when running `ncu` with `upgrade` and `deep` options logs something like:
```
Upgrading C:\path\to\project-name\package.json

All dependencies match the latest package versions :)
Upgrading C:\path\to\project-name\packages\package-name-1\package.json

All dependencies match the latest package versions :)
Upgrading C:\path\to\project-name\packages\package-name-2\package.json

No dependencies.
Upgrading C:\path\to\project-name\packages\package-name-3\package.json

No dependencies.
Upgrading C:\path\to\project-name\packages\package-name-4\package.json

All dependencies match the latest package versions :)
Upgrading C:\path\to\project-name\packages\package-name-5\package.json

All dependencies match the latest package versions :)
```

* I suggest if we put the line with the result directly under the line with the path (without an empty line). And separate each package log with an empty line.
Even that I come from a ux design background, it always looks very confusing to me. So the result of this cosmetic suggestion will be easier logs to navigate through.

* I would also suggest using `chalk` not just to color the `:)` green smile at the end, but also to color everything else... especially the result. It will look like a mess if you used more colors (because you already have colors for major, minor, and patch versions) So I suggest coloring the result to be grey (only if dependencies are matched or there are no dependencies at all) Grey will tell the developer to skip those lines because there's no required action to take.

* My last suggestion would be to add a summary at the end, that the developer can skip right to it to take all necessary actions. It would look something like a testing framework.

* I still have many ideas in mind to improve the console like [1] simplifying paths and maybe [2] grouping the packages that required no action together with minimum logging and the ones that required actions with more details. [3] I would also suggest if the developer is using `ncu` with a monorepo, to add an option to their `.ncurc` file that says something like: `monorepo: true` so we can improve and minimize logging if so.

There are a lot of decisions to make here. I would be happy to dig in the package and create PR for you.

terminalchai:
## Problem

When running `ncu --deep` on a workspace with multiple packages, the status output contains inconsistent blank lines between package sections.

Specifically:
- `printUpgrades` always emits a leading blank line before printing package results, even in `--deep` mode where the spacing between packages is managed separately in `runUpgrades`.
- This causes double blank lines between some packages and missing blank lines before others.

Fixes #956.

## Fix

Two changes to normalize the output spacing in `--deep` mode:

1. **`src/lib/logging.ts`**: Skip the leading blank line in `printUpgrades` when `options.deep` is set — deep mode manages its own inter-package spacing.

2. **`src/index.ts`**: Emit a blank separator line before each *subsequent* package's header (`Upgrading <path>` or `Checking <path> catalog dependencies`) so packages are clearly separated without doubling up.

Together these ensure exactly one blank line between packages in deep mode, consistent with the output format for individual packages.

## Tests

Added a regression test in `test/deep.test.ts`:
- Creates a root `package.json` with an outdated dependency and a nested `packages/no-deps/package.json` with no dependencies.
- Asserts the output has no extra blank line between the header and status lines, and that packages are separated by exactly one blank line.
- Uses a `getCliInvocation` helper (ts-node fallback when `build/` is absent) to keep tests runnable in both built and source-only environments.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -232,6 +232,7 @@ async function runUpgrades(options: Options, timeout?: NodeJS.Timeout): Promise<
     analysis = await selectedPackageInfos.reduce(
       async (previousPromise, packageInfo: PackageInfo) => {
         const packages = await previousPromise
+        const isSubsequentPackage = Object.keys(packages).length > 0
         // copy object to prevent share .ncurc options between different packageFile, to prevent unpredictable behavior
         const rcResult = await getNcuRc({ packageFile: packageInfo.filepath, options })
         let rcConfig = rcResult.config
@@ -262,9 +263,15 @@ async function runUpgrades(options: Options, timeout?: NodeJS.Timeout): Promise<
 
           // Print the same message as findPackage for consistency
           const relPathToPackage = path.resolve(indexKey)
+          if (isSubsequentPackage) {
+            print(pkgOptions, '')
+          }
           print(pkgOptions, `${pkgOptions.upgrade ? 'Upgrading' : 'Checking'} ${relPathToPackage} catalog dependencies`)
         } else {
           // Regular file - read from disk
+          if (isSubsequentPackage) {
+            print(pkgOptions, '')
+          }
           const result = await findPackage(pkgOptions)
           pkgData = result.pkgData
           pkgFile = result.pkgFile || packageInfo.filepath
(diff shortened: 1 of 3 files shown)


raineorshine (approved):
Thanks! 

Please resolve merge conflicts and then it's good to go.

raineorshine (approved):

PR #1763, opened 2026-05-27
fix: handle registries missing 'time' metadata during version resolution

### LINKED ISSUE #1746: Packages from registries without "time" metadata are skipped even when cooldown predicate returns 0

---

## Steps to Reproduce

.ncurc:

```js
import { defineConfig } from 'npm-check-updates'

export default defineConfig({
  format: 'cooldown',
  cooldown: packageName => (packageName.startsWith('@my-company') ? 0 : 3),
})
```

Environment:

```
❯ node -v && npm -v
v24.11.1
11.15.0
```

Dependencies:

```json
{
  "dependencies": {
    "@my-company/component-library": "^3.0.0",
    "react": "^19.2.4",
    "vite": "^8.0.9",
    "vitest": "^4.1.5"
  },
  "devDependencies": {
    "npm-check-updates": "^22.2.0"
  }
}
```

`@my-company/*` is published to a private registry (self-hosted GitLab in our case) that does **not** populate the `time` field in the npm metadata response. This is a known GitLab limitation tracked here: https://gitlab.com/gitlab-org/gitlab/-/issues/585659. The same issue affects GitHub Packages and is the cause of [pnpm/pnpm#10610](https://github.com/pnpm/pnpm/issues/10610).

Confirmed empty `time` via:

```sh
curl -s 'https://<gitlab-host>/api/v4/projects/<id>/packages/npm/@my-company/component-library' | jq '.time'
# → null
```

Steps:

1. Configure `.ncurc` with the predicate above.
2. Run `ncu`.

## Current Behavior

`@my-company/component-library` appears under `Skipped due to cooldown`. Notably, no "X ago" timestamp is displayed for this package in the skipped output, unlike the other entries.

```
Skipped due to cooldown
 @my-company/component-library   ^3.0.0  →   ^3.4.0
 @vitest/coverage-v8             ^4.1.6  →   ^4.1.7    1 day ago
 vite                           ^8.0.13  →  ^8.0.14    10 hours ago
 vitest                          ^4.1.6  →   ^4.1.7    1 day ago

Updates
 ...
```

With the predicate instrumented to write to a file:

```js
import { defineConfig } from 'npm-check-updates';
import { appendFileSync } from 'node:fs';

export default defineConfig({
  format: 'cooldown',
  cooldown: packageName => {
    appendFileSync('/tmp/ncu-debug.log', `${packageName}\n`);
    return 3;
  }
});
```

The log contains every other dependency in the project (`react', `vite`, `vitest`, etc.) but `@my-company/component-library` does not appear in it — the function is genuinely never invoked for that package. 

## Expected Behavior

The `cooldown` predicate should be invoked for every package under consideration, regardless of whether the registry response includes a `time` field. Currently the timestamp lookup appears to happen before the predicate is consulted, and a missing timestamp causes ncu to skip the package before the predicate gets a chance to opt it out of the cooldown.

Once the predicate is invoked, a return value of `0` should be honored as "no cooldown" without requiring a timestamp — "no cooldown" can't logically fail an age check.

This would let users on registries that don't emit `time` metadata (GitLab, GitHub Packages, others) opt internal packages out of cooldown via the documented predicate, which is currently the suggested workflow in the README for exactly this use case.

onemen:
close #1746

**Summary**
This PR addresses an issue where packages from registries lacking `time` metadata were incorrectly skipped during version resolution, even when the cooldown period was set to 0.

**Changes**

* **Version Resolution Logic for 'newest':** Updated the resolution strategy to default to the greatest available version when `time` metadata is unavailable for a package, ensuring that 'newest' targets are correctly identified.
* **Cooldown Bypass:** Modified the filtering logic to prevent packages from being skipped due to cooldown predicates when their publish time information is missing.
* **UI Feedback:** Enhanced logging to explicitly indicate when a version is being upgraded despite missing `time` metadata by appending `[missing time]` to the output.

**Motivation**
Previously, the absence of `time` metadata triggered a `skip by cooldown` even when the cooldown was set to zero.

**Testing**

* Verified that packages missing `time` metadata are now correctly included in resolution candidates.
* Confirmed that version selection defaults to the highest semantic version when timestamps are absent for 'newest' targets.
* Verified that log output correctly displays `[missing time]` during the upgrade process.

**Additional Fixes**
* Resolved an edge case where the CLI would incorrectly show a "No package versions were returned" warning when the currently installed version is within the cooldown period and matches the target version.

### Testing Improvements, WIP for future PR

* Currently working on a comprehensive migration from Mocha to Vitest. This transition is already yielding significant performance gains, reducing local test execution times from approximately 2 minutes to under 15 seconds.

### DIFF
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -170,6 +170,7 @@ const findTargetAndFallback = ({
       targetBlockedByCooldown: false,
     }
   }
+  const currentInCooldown = !satisfiesCooldownPeriod(packageName, cur, time?.[cur], options.cooldown)
 
   const result = versions.reduce(
     (acc, versionData) => {
@@ -204,7 +205,7 @@ const findTargetAndFallback = ({
     {
       targetVersion: cur,
       fallbackVersion: cur,
-      targetBlockedByCooldown: false,
+      targetBlockedByCooldown: currentInCooldown,
     } as {
       targetVersion: string
       fallbackVersion: string
@@ -212,7 +213,7 @@ const findTargetAndFallback = ({
     },
   )
 
-  let targetVersion: string | null = result.targetVersion
+  const targetVersion: string | null = result.targetVersion
   let fallbackVersion: string | null = result.fallbackVersion
 
   if (fallbackVersion === result.targetVersion) {
@@ -223,8 +224,9 @@ const findTargetAndFallback = ({
     fallbackVersion = null
   }
 
-  if (!nodeSemver.gt(targetVersion, cur)) {
-    targetVersion = null
+  // don't "Block" current version
+  if (currentInCooldown && targetVersion === cur) {
+    result.targetBlockedByCooldown = false
   }
 
   return {
@@ -562,7 +564,10 @@ npmApi.mockFetchUpgradedPackument =
       )
     }
 
-    const time = (isPackument(partialPackument) && partialPackument.time?.[version]) || new Date().toISOString()
+    const time =
+      isPackument(partialPackument) && partialPackument.time
+        ? partialPackument.time?.[version]
+        : new Date().toISOString()
     const packument: Packument = {
       name,
       'dist-tags': {
@@ -1005,9 +1010,11 @@ export const distTag: GetVersion = async (
 
   const publishTime = packument?.time?.[version!]
   const maybeTime = publishTime ? { time: publishTime } : null
+  const current = nodeSemver.minVersion(currentVersion)?.version ?? '0.0.0'
 
   const isSatisfiesCooldown =
-    tagPackument && satisfiesCooldownPeriod(packageName, tagPackument.version, publishTime, options.cooldown)
+    tagPackument.version === current ||
+    (tagPackument && satisfiesCooldownPeriod(packageName, tagPackument.version, publishTime, options.cooldown))
 
   // latest should not be deprecated
   // if latest exists and latest is not a prerelease version, return it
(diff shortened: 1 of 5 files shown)


onemen:
@raineorshine 

this PR is ready for review

raineorshine (changes requested):
Hi, thanks for the submission.

The diff contains many changes that are unrelated to #1746. Please limit PRs to relevant changes with a reasonable scope and I will be happy to review.

onemen:
@raineorshine 
This PR is ready for review, it also covered the changes from #1766

src/package-managers/npm.ts
```diff
@@ -1119,16 +1122,28 @@ export const newest: GetVersion = async (
...
-      const t2 = packument?.time?.[v2] || ''
-      return t1 > t2 ? 1 : t1 < t2 ? -1 : 0
-    },
-  })
+  const time = packument?.time
+  const isTimeMissing = !time || Object.keys(time).length === 0
+  const packageInfo = { packageName, currentVersion, options, versions, time }
+
+  const versionResult = isTimeMissing
+    ? // we can not search for newest when there are no times, fallback to greatest
```
raineorshine: This is introducing undesirable new semantics. There is no intention for `newest` to fallback to `greatest`. We definitely do not want the added complexity and the need to document it in the README.

Please preserve the existing functionality. If time is missing for a given package, then the user should configure the target differently rather than rely on an opaque fallback mechanism.

raineorshine (changes requested):
Hi, thanks for the submission.

src/package-managers/npm.ts
onemen: Ok

Do you want to print some warning at the bottom of the output for this case?

raineorshine: No, thanks. Let's keep this focused on the issue. I don't want to open this up to formatting and logging considerations.

onemen: done

raineorshine (approved):

test/bin.test.ts
```diff
@@ -287,7 +298,8 @@ describe('bin', async function () {
       const { stdout } = await spawn('node', [bin, '--stdin'], { stdin: JSON.stringify({ dependencies }) })
       stripAnsi(stdout)
         .trim()
-        .should.equal('ncu-test-v2  https://github.com/raineorshine/ncu-test-v2.git#v1.0.0  →  v2.0.0')
+        .replace(/\s+/g, ' ') // Replace all whitespace sequences with a single space
+        .should.equal('ncu-test-v2 https://github.com/raineorshine/ncu-test-v2.git#v1.0.0 → v2.0.0 [missing time]')
```
raineorshine, line 302: I missed this in my review, but this should not show `[missing time]`. The annotation should only be printed when `--format time` or `--format cooldown` is given.

Fixed in b8201313.

PR #1766, opened 2026-05-27
fix: prevent incorrect warnings when no upgrade is found for target

### LINKED ISSUE #1764: Incorrect "No package versions were returned" warning for non-latest --target when no upgrade is available

---

## Steps to Reproduce

Dependencies:

```json
{
  "dependencies": {
    "ncu-test-v2": "2.0.0"
  }
}
```

`ncu-test-v2@2.0.0` is the latest published version, so no minor upgrade is available.

Steps:

1. Run `ncu --target minor` (also happens with other non-`latest` targets).

## Current Behavior

```
No package versions were returned. This may be a problem with your installed npm, the npm registry, or your Internet connection. Make sure npx pacote packument ncu-test-v2 is working before reporting an issue.
```

The message is confusing/incorrect as the registry is available and the package was checked successfully, there just isn't an upgrade available at the `--target` level.

With `--target latest` it correctly prints `All dependencies match the latest package versions :)` for the exact same project.

## Expected Behavior

```
All dependencies match the minor package versions :)
```

### Environment

- node: v24.13.0
- npm-check-updates: 22.2.1

duniul:
Fixes #1764.

AI assisted but tested and reviewed manually.

## Context

When you run `ncu` with `--target` set to anything other than `latest` and there is no upgrade available, ncu prints a "No package versions were returned" warning that sounds like the registry or network is broken. In reality the registry was reached and the packages were checked, there just is no upgrade at the requested level.

Seems to have been introduced in 22.2.0 (#1711).

## Changes

Two small changes to `findTargetAndFallback`:

1. If `versions` comes back empty, return early to ensure "No package versions were returned" is still printed for real registry failures.

2. When ncu finishes checking a package and `targetVersion` is the same as the current, keep that version as the result instead of setting it to `null`. The existing upgrade check still filters it, but ncu can now tell it actually checked the package, so it prints `All dependencies match the <target> package versions :)` as expected.

## Tests

`test/bin.test.ts`: running `ncu --target minor` on a project that is already on the latest minor version now prints "All dependencies match the minor package versions" instead of the warning.

### DIFF
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -171,6 +171,14 @@ const findTargetAndFallback = ({
     }
   }
 
+  if (versions.length === 0) {
+    return {
+      targetVersion: null,
+      fallbackVersion: null,
+      targetBlockedByCooldown: false,
+    }
+  }
+
   const result = versions.reduce(
     (acc, versionData) => {
       const version = versionData.version
@@ -212,7 +220,7 @@ const findTargetAndFallback = ({
     },
   )
 
-  let targetVersion: string | null = result.targetVersion
+  const targetVersion: string = result.targetVersion
   let fallbackVersion: string | null = result.fallbackVersion
 
   if (fallbackVersion === result.targetVersion) {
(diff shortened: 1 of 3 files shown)


onemen:
This Pr already covered by #1763

duniul:
Closing since #1763 fixed this (even included the test 🌝)

PR #1710, opened 2026-05-03
fix: skip intersects() for non-semver specs like catalog: in peer dep checks

terminalchai:
## Problem

When a pnpm workspace lists a dependency using the [catalog protocol](https://pnpm.io/catalogs) (e.g. `"catalog:"` or `"catalog:default"`), that string is not a valid semver range.

The peer dependency constraint checking code called `intersects(upgradedDependencies[peer], peerSpec)` without first verifying that `upgradedDependencies[peer]` is a valid semver range. Since `semver.intersects()` internally calls `semver.parse()` which throws on invalid input, running `ncu --peer -w` on a workspace containing `catalog:` references in `peerDependencies` crashes with:

```
TypeError: Invalid comparator: catalog:
```

The same guard (`!validRange(peerSpec)`) already exists for the _right-hand_ operand — it was just missing for the _left-hand_ one.

Fixes #1604.

## Fix

Add a `!validRange(upgradedDependencies[peer])` guard in both places that call `intersects()` with a value sourced from the current/upgraded dependency map (`getIgnoredUpgradesDueToPeerDeps.ts` and `upgradePackageDefinitions.ts`).

When the installed/upgraded version spec is not a valid semver range (`catalog:`, `file:`, `git+https://…`, etc.) the constraint check is skipped and treated as compatible — matching the existing convention already applied to the peer spec side.

A regression test covering the `catalog:` case is included in `test/peer.test.ts`.

PR #1767, opened 2026-05-27
fix: do not call semver.minVersion on non-semver package protocols

### LINKED ISSUE #1765: `TypeError: Invalid comparator: catalog:` when using `catalog:` protocols and a non-latest --target

---

## Steps to Reproduce

Dependencies:

`pnpm-workspace.yaml`:

```yaml
packages:
  - 'packages/**'

catalog:
  ncu-test-tag: '1.0.0'
```

`packages/a/package.json`:

```json
{
  "dependencies": {
    "ncu-test-tag": "catalog:"
  }
}
```

(an empty `pnpm-lock.yaml` is present so pnpm is detected)

Steps:

1. Run `ncu -w --target minor`

## Current Behavior

Every dependency that uses a `catalog:` reference in a workspace `package.json` is reported as an error in the output table:

```
Checking packages/a/package.json
 ncu-test-tag  TypeError: Invalid comparator: catalog:

Checking pnpm-workspace.yaml catalog dependencies
 ncu-test-tag  1.0.0  →  1.1.0
```

The same seems to happen with other targets than `'latest'` for `catalog:` dependencies.

The default (`--target latest`) does not throw any errors.

image

## Expected Behavior

`ncu -w --target minor` should match the `--target latest` behavior: silently skip the non-registry references in the workspace `package.json` and only upgrade the catalog entry in `pnpm-workspace.yaml`. No per-dependency `TypeError` rows in the table.

### Environment

- node: v24.13.0
- pnpm: 11.3.0
- npm-check-updates: 22.2.1

duniul:
Fixes #1765.

AI assisted but tested and reviewed manually.

## Context

When a workspace uses a non-semver package protocol (like `catalog:`, `workspace:`, `link:` etc.), running `ncu` with a `--target` that isn't `latest` throws a `TypeError: Invalid comparator: catalog:` error for each of those deps.

Related: #1710 fixed the same error on the `--peer` path.

image

## Changes

- `src/lib/queryVersions.ts`: detect the non-registry protocols up front and skip them, the same way the `latest` path already does.
- `src/lib/version-util.ts`: in `filterByLevel`, return a filter that returns false when the input is not a valid version range, instead of trying to parse it.

## Tests

- `test/version-util.test.ts`: `filterByLevel` returns `false` for `catalog:`, `workspace:`, `link:`, and `file:` inputs.
- `test/workspaces.test.ts`: a workspace with all four protocol types in its dependencies runs cleanly under `--target minor`, and the catalog entry still upgrades as expected.

### DIFF
--- a/src/lib/queryVersions.ts
+++ b/src/lib/queryVersions.ts
@@ -14,6 +14,16 @@ import keyValueBy from './keyValueBy'
 import programError from './programError'
 import { createNpmAlias, isGitHubUrl, isPre, parseNpmAlias } from './version-util'
 
+const PKG_MANAGER_PROTOCOLS = ['catalog:', 'workspace:', 'link:', 'file:', 'portal:'] as const
+
+/**
+ * Returns true if the spec uses a package manager protocol that references something
+ * other than a registry version (e.g. pnpm's `catalog:` / `workspace:`, or `link:` / `file:`).
+ */
+function isPackageManagerProtocol(spec: VersionSpec): boolean {
+  return PKG_MANAGER_PROTOCOLS.some(protocol => spec.startsWith(protocol))
+}
+
 /**
  * Get the latest or greatest versions from the NPM repository based on the version target.
  *
@@ -42,6 +52,13 @@ async function queryVersions(packageMap: Index<VersionSpec>, options: Options = 
   async function getPackageVersionProtected(dep: VersionSpec): Promise<VersionResult> {
     const npmAlias = parseNpmAlias(packageMap[dep])
     const [name, version] = npmAlias || [dep, packageMap[dep]]
+
+    // Skip valid specs that are not registry versions, such as different package manager protocols.
+    if (isPackageManagerProtocol(version)) {
+      bar?.tick()
+      return { version: null }
+    }
+
     const targetOption = options.target || 'latest'
     const targetString = typeof targetOption === 'string' ? targetOption : targetOption(name, parseRange(version))
     const [target, distTag] = targetString.startsWith('@')
--- a/src/lib/queryVersions.ts
+++ b/src/lib/queryVersions.ts
@@ -10,19 +10,10 @@ import { type VersionResult } from '../types/VersionResult'
 import { type VersionSpec } from '../types/VersionSpec'
 import { getChalk } from './chalk'
 import getPackageManager from './getPackageManager'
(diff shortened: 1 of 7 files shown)


onemen:
I have change semver.minVersion calls in `npm.ts` in #1763 since the PR change some of these calls and i don't know which PR will be merged first

duniul:
@raineorshine @onemen this PR is ready to go if you have time to take a look! #1763 fixed #1764 but not #1765, this PR should though.

src/lib/queryVersions.ts
```diff
@@ -14,6 +14,16 @@ import keyValueBy from './keyValueBy'
...
 
+const PKG_MANAGER_PROTOCOLS = ['catalog:', 'workspace:', 'link:', 'file:', 'portal:'] as const
+
+/**
+ * Returns true if the spec uses a package manager protocol that references something
+ * other than a registry version (e.g. pnpm's `catalog:` / `workspace:`, or `link:` / `file:`).
+ */
+function isPackageManagerProtocol(spec: VersionSpec): boolean {
+  return PKG_MANAGER_PROTOCOLS.some(protocol => spec.startsWith(protocol))
+}
```
raineorshine: Could you pull [isFetchable](https://github.com/raineorshine/npm-check-updates/blob/042f180f066bfd98d1d988cee8b009bddc2427d3/src/lib/logging.ts#L44-L51) out into a separate file and use that? I believe that provides the protocol check you are looking for.

raineorshine (changes requested):
Thanks!

One small request.

src/lib/queryVersions.ts
duniul: `isFetchable` wouldn't work here without combining it with `!isGithubUrl()`, as short Github urls still need to resolve via `gitTags`. But I pulled out `isFetchable` and `isPackageManagerProtocol` and combined them to avoid the duplicated protocol checks 👍

See [d071786](https://github.com/raineorshine/npm-check-updates/pull/1767/commits/d071786b554d016ff33fe4c2eab18254bee04deb)!

Tried to replicate how other single-function utils were set up, but let me know if you'd prefer it some other way!

raineorshine (approved):
Thanks, looks good!

src/lib/queryVersions.ts
raineorshine: Yes, removing the duplication was the main point! Thanks!

PR #1786, opened 2026-06-09
CI: harden and align GitHub Actions workflows

XhmikosR:
Needs squash when merging.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,23 +16,30 @@ env:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     if: github.event_name != 'pull_request' || github.event.pull_request.changed_files > 0
 
     steps:
       - name: Clone repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
 
       - name: Set up Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: ${{ env.NODE }}
           cache: npm
 
       - name: Cache tsbuildinfo
-        uses: actions/cache@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/lint.yml
```diff
@@ -16,23 +16,30 @@ env:
...
         with:
           node-version: ${{ env.NODE }}
           cache: npm
 
       - name: Cache tsbuildinfo
-        uses: actions/cache@v4
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
           path: node_modules/.cache
           key: tsbuildinfo-${{ runner.os }}-${{ hashFiles('**/tsconfig.json') }}
```
XhmikosR, line 46: is this actually used?

XhmikosR:
@raineorshine: I think this is ready, with one caveat: I cannot really confirm the CodeQL changes... I can adapt the changes later if needed, after merging.

.github/workflows/lint.yml
raineorshine: Yes. It persists the `node_modules/.cache` directory across workflow runs. On a later run, the cache is restored before the build, so TypeScript can reuse prior work instead of re-checking everything from scratch. You can see it configured in the [`tsBuildInfoFile` key in `tsconfig`](https://github.com/raineorshine/npm-check-updates/blob/ec8b9469679ae4ba90b087309af50db6c4a44084/tsconfig.json#L21).

raineorshine (approved):
Looks good, thanks!

XhmikosR:
BTW @raineorshine you may need to go to the repository settings https://github.com/raineorshine/npm-check-updates/security/code-scanning/tools/CodeQL/status and remove the old CodeQL config. This should fix the `CodeQL 1 configuration not found ` issue in PRs. It's not very easy to find what to delete there, but it is what it is :)

raineorshine:
Okay, I've deleted the old CodeQL configuration. Thanks.

XhmikosR:
@raineorshine: I notice that when you merge PRs, you merge too fast (that's OK BTW). Maybe I should tweak the concurrency so that main branch isn't cancelled?

raineorshine:
Yes, we should probably ensure that `main` gets a complete CI run on every merge. Otherwise it's possible, though unlikely, to not see the first merge that broke.

XhmikosR:
I think it should be OK now after #1837. If not, let me know and I'll drop the concurrency altogether. Better have a stable CI than not.

XhmikosR:
Reverted concurrency in #1846.

PR #1797, opened 2026-06-10
feat: respect min-release-age-exclude with min-release-age npm config

### LINKED ISSUE #1796: Feature Request: support min-release-age-exclude
As per https://github.com/npm/cli/pull/9532, npm will soon support excluding packages from the `min-release-age` config which for was added here recently.

It would be great to see this respected here too, so that new versions for packages in the `min-release-age-exclude`  config appear as expected.

Cherry:
Closes #1796

Aided with AI, but validated with tests and manually. This adds support for the upcoming `min-release-age-exclude` config, which when paired with `min-release-age`, can still enforce supply chain attack protection for third-party dependencies, but allow you to keep your own first-party org packages up to date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -511,7 +511,7 @@ cooldown: packageName => (packageName.startsWith('@my-company') ? 0 : 3)
 If `--cooldown` is not set explicitly, `ncu` automatically reads the cooldown configuration from your package manager's config file:
 
 <table>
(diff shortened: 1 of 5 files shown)


raineorshine (approved):
Looks good to me! Thanks!

PR #1805, opened 2026-06-14
package.json: fix npm scripts to be cross-platform

XhmikosR:
The previous way doesn't work on Windows native CMD. It works on GitHub Actions CI because they use bash there which isn't the default Windows setup. The changes make things work everywhere.


package.json
```diff
@@ -39,19 +39,19 @@
...
     "lint:src": "eslint --cache --cache-location node_modules/.cache/.eslintcache --report-unused-disable-directives",
-    "prepare": "src/scripts/install-hooks",
+    "prepare": "git config core.hooksPath .hooks",
     "prepublishOnly": "npm run build",
     "prettier": "prettier . --check",
     "prettier:fix": "prettier . --write",
     "test": "tsc --noEmit && npm run test:unit && npm run test:e2e",
     "test:bun": "mocha test/bun",
-    "test:unit": "mocha test test/package-managers/* --ignore 'test/bun/*'",
+    "test:unit": "mocha test --ignore \"test/bun/*\"",
```
XhmikosR, line 54: Removed the old `test/package-managers/*` since it's already included by the mocha config anyway and this didn't match anything on Windows.

raineorshine (approved):

PR #1807, opened 2026-06-14
build workflow: add npm pack --dry-run

XhmikosR:
We could/should switch later to using something like https://github.com/ai/size-limit so that size bumps are caught.


raineorshine (approved):
Thanks!

PR #1810, opened 2026-06-14
Replace js-yaml with yaml

XhmikosR:
`yaml` is already used in other places in the codebase.

I will make a PR for replacing cosmiconfig with lilconfig, then js-yaml will be dropped from the production bundle.

Other big savings would come from dropping cacache and/or zod, but that's for another time and I'm not sure if it's even possible. Also, update-notifier pulls a significant amount of deps.


raineorshine (approved):
Makes sense since we have to keep `yaml` for full AST capabilities and use of `parseDocument`.

Note that this reverts 5439c87 which removed `yaml` only to have it added back in again in 4ea83fc.

PR #1811, opened 2026-06-14
Replace dequal with node:util isDeepStrictEqual

XhmikosR:
Small savings, but we can do the same with the native isDeepStrictEqual.

### DIFF
--- a/src/lib/initOptions.ts
+++ b/src/lib/initOptions.ts
@@ -1,4 +1,3 @@
-import { dequal } from 'dequal'
 import { propertyOf } from 'lodash-es'
 import picomatch from 'picomatch'
 import cliOptions from '../cli-options'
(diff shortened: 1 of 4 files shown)


src/lib/initOptions.ts
```diff
@@ -148,7 +147,7 @@ async function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } =
 
   // convert to string for comparison purposes
   // otherwise ['a b'] will not match ['a', 'b']
-  if (options.filter && args && !dequal(args.join(' '), Array.isArray(filter) ? filter.join(' ') : filter)) {
+  if (options.filter && args && args.join(' ') !== (Array.isArray(filter) ? filter.join(' ') : filter)) {
```
XhmikosR, line 150: This was/is comparing strings

raineorshine (approved):

Issue #1683, opened 2026-04-21
Huge increase in build size with the use of cosmiconfig


onemen:
**Description:**

When bundling a project with Vite that uses cosmiconfig, the build includes a ~5MB TypeScript chunk even when TypeScript support is explicitly disabled via the `loaders` option.

### Build sizes

With cosmiconfig (`.ts` excluded from `loaders`):
```
build/chunks/typescript-CXzCgl35.js  5,150.16 kB │ gzip: 1,169.39 kB
build/index.js                       1,768.13 kB │ gzip:   533.54 kB
```

Without cosmiconfig (same project):
```
build/index.js                       ~1,700 kB (no TypeScript chunk)
```

### Root cause

`defaultLoaders['.ts']` contains a dynamic `import('typescript')` that Vite's static analysis traces during bundling. Even though the loader is never invoked at runtime, the module graph includes TypeScript.

### Current workaround options

- Use `resolve.alias` in Vite to stub `typescript` to a no-op module
- Mark `typescript` as `external` in Rollup/Vite config
- Switch to `lilconfig` (cosmiconfig's core without bundled heavy loaders)
- Post-process build to delete the TypeScript chunk

Need more research...

onemen:
### Current workaround (stub plugin)

We can implemented a Vite plugin that intercepts `typescript` resolution and returns a stub:

```diff
+/**
+ * cosmiconfig bundles a TypeScript loader that dynamically imports the
+ * `typescript` package (~5MB). Even though we exclude `.ts` from
+ * `searchPlaces`, Vite's static analysis traces the `require('typescript')'
+ * inside `loadTsSync` and includes the entire package in the build.
+ *
+ * This plugin stubs out the `typescript` module to prevent it from being
+ * bundled. The stub is never invoked at runtime because `.ts` configs are
+ * not in our `searchPlaces`.
+ */
+const TYPESCRIPT_STUB = `
+export const version = '0.0.0-stub'
+export function createProgram() {
+  throw new Error('TypeScript loader is not available in this build')
+}
+export default { version, createProgram }
+`
+
 export default defineConfig({
   plugins: [
+    {
+      name: 'stub-typescript',
+      enforce: 'pre',
+      resolveId(id) {
+        if (id === 'typescript') {
+          return '\0stub:typescript'
+        }
+      },
+      load(id) {
+        if (id === '\0stub:typescript') {
+          return TYPESCRIPT_STUB
+        }
+      },
+    },
[...]

PR #1812, opened 2026-06-14
Replace cosmiconfig with lilconfig

XhmikosR:
Along with #1810 this should yield some savings from the bundle size.

The change should be backwards compatible since lilconfig does `searchStrategy: 'global'` by default. The error message is kept the same manually. This also saves us the whole typescript stub thing.


raineorshine:
Hi, thanks for the submission.

We [previously decided](https://github.com/raineorshine/npm-check-updates/issues/1683#issuecomment-4295607440) to go with cosmiconfig + a stub to remove the Typescript loader from the bundle. Can you explain why the proposed solution is better?

XhmikosR:
Yeah, I noticed it. I wouldn't have gone with comisconfig in the first place myself.

lilconfig is a lot smaller. Doesn't need the TypeScript hack in vite. Functionality seems to be identical, minus the custom yaml loader and the error thrown in that case, which I tried to replicate.

https://npmx.dev/compare?packages=cosmiconfig,lilconfig

I know cosmiconfig is more popular, but if lilconfig does the job cleaner, why not use it?

This PR will drop js-yaml altogether from the production bundle and the extra chunk from build.

## Bundle output

| Artifact | main | lilconfig | Delta |
|---|---|---|---|
| index.js raw | 1,626.65 kB | 1,528.74 kB | -97.9 kB |
| index.js gzip | 506.72 kB | 477.46 kB | -29.3 kB |
| index.cjs raw | 1,299.98 kB | 1,223.96 kB | -76.0 kB |
| index.cjs gzip | 468.59 kB | 441.67 kB | -26.9 kB |
| modules transformed | 1403 | 1354 | -49 |
| cli.js / cli.cjs | ~same | ~same | ~same |

## Published tarball

| | main | lilconfig | Delta |
|---|---|---|---|
| package size | 3.5 MB | 3.3 MB | -0.2 MB |
| unpacked size | 13.2 MB | 12.5 MB | -0.7 MB |
| total files | 15 | 13 | -2 |

Not a huge difference, but important IMHO.

raineorshine (approved):
Great, thanks for the explanation.

PR #1813, opened 2026-06-14
Fix CodeQL issues

XhmikosR:
The `1 configuration not found` warning is because the old config needs to be removed from the repo settings.

### DIFF
--- a/src/lib/getNcuRc.ts
+++ b/src/lib/getNcuRc.ts
@@ -102,7 +102,6 @@ async function getNcuRc({
       rawResult = await explorer.load(targetFile)
     } else {
       rawResult = await explorer.search(cwd)
-      targetFile = rawResult?.filepath
     }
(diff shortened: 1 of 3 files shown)


raineorshine (commented):
I need a little more explanation about these changes. On the surface, they all seem to be useful/necessary. They are providing fallbacks for specific cases. If those fallbacks are no longer necessary, I would need an explanation to justify their removal.

XhmikosR:
Oh, sorry, I thought you already saw https://github.com/raineorshine/npm-check-updates/security/code-scanning. It explains each one there.

1. versions is always true there
2. parseError is already handled two lines above so it's always false
3. targetFile is not read after this

All of them are benign, just noticed them because CodeQL flagged them.

raineorshine (approved):
I see now. Thanks!

PR #1819, opened 2026-06-16
Remove lodash-es

XhmikosR:
@raineorshine: I kept the patches as separate commits. I think it'd be better if these landed without a squash for future reference.

I tried to cover all cases regarding undefined/null values which lodash usually covers silently. Let me know if you spot any issues. :)

This change reduces the modules that are transformed from 1403 to 764 and should also speed up the build due to that (most noticeable on Windows). Also, decreases the built files's size and npm pack size, not by a lot, but still.


raineorshine (approved):
Looks fine, thanks!

I personally prefer the `flow` syntax, but I'm okay with the change.

PR #1824, opened 2026-06-17
Switch to async/await in a few more places

XhmikosR:
I didn't change all of the Promise cases, mostly the ones that made sense to me, but open to discussion :)


raineorshine (approved):
Looks good to me! Thanks.

PR #1825, opened 2026-06-17
Replace forEach with for...of

XhmikosR:
I think `for...of` is cleaner and doesn't fire a callback each time. I kept the vendored files as is, and the `keyValueBy` instance.

Let me know if you want me to make any changes or drop this altogether.


raineorshine (approved):
Sure, I'm fine with that.

PR #1826, opened 2026-06-17
Replace nested ternaries with if/else for readability

XhmikosR:
I think the flattened syntax reads better, it's easier to understand than the ladder ternaries.

Open to discussion in case you prefer the old way.

PS. not sure if ESLint has a built-in rule for this.

### DIFF
--- a/src/bin/cli.ts
+++ b/src/bin/cli.ts
@@ -114,19 +114,25 @@ ${chalk.dim.underline(
     .usage('[options] [filter]')
     // See: boolean optional arg below
     .configureHelp({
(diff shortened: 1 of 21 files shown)


raineorshine:
Hi, thanks for the suggestion.

On this matter I have a preference for nested ternaries. I think with a little practice they are very clean and easy to read. I've used them for many years and still prefer the syntax.

It's too bad that prettier doesn't allow a flatter structure. That was a battle I couldn't win.

XhmikosR:
No problem, we just have different opinions on the matter.

PR #1827, opened 2026-06-17
Move src/scripts to the root folder

XhmikosR:
Seems like a better fit since they are not strictly source files.


raineorshine (approved):
Thanks, makes sense.

Just need to resolve one merge conflict.

raineorshine (approved):

PR #1839, opened 2026-06-20
Remove redundant awaits and enable @typescript-eslint/await-thenable

XhmikosR:


raineorshine (approved):
Thanks. This was a legacy `await` from our pre-esm build.

PR #1841, opened 2026-06-20
Move configs out of package.json

XhmikosR:
I'm in the between myself about this. On one hand, the package.json file is growing. On the other hand, we end up with multiple files at root. I mostly prefer this approach in my projects but let me know and I will just rebase to keep the lockfile-lint tightening.


jerone:
One small advantage; it makes the package a few bytes smaller as the separate configs should not be bundled.

raineorshine (approved):
I like it. Dot files are quiet.

PR #1842, opened 2026-06-20
Enforce node: protocol and dedupe imports

XhmikosR:
- add n/prefer-node-protocol and import-x/no-duplicates ESLint rules
- group Node builtins first via <BUILTIN_MODULES> in import sort
- apply node: prefix and merge duplicate imports across the codebase
- use default path import (path.dirname) instead of destructuring dirname
- normalize core module import style to the codebase majority

Let me know if you want me to make any changes, I wasn't sure about the semver change and it's mostly stylistic.


src/package-managers/npm.ts
```diff
@@ -1,10 +1,10 @@
+import fs from 'node:fs'
+import path from 'node:path'
 import { JSONParser } from '@streamparser/json'
 import camelCase from 'camelcase'
 import memoize from 'fast-memoize'
-import fs from 'fs'
 import ini from 'ini'
 import npmRegistryFetch from 'npm-registry-fetch'
-import path from 'path'
 import nodeSemver from 'semver'
```
raineorshine, line 8: Let's use `semver` everywhere, or `nodeSemver` everywhere if they are any name conflicts.

raineorshine (changes requested):
Yes, I like it.

src/package-managers/npm.ts
XhmikosR: The reason I kept `nodeSemver` here is because there's an exported function named `semver` in this file:

https://github.com/raineorshine/npm-check-updates/blob/145b491a8a60923eddf413b4497c8c1e86d357e9/src/package-managers/npm.ts#L1238

src/lib/cache.ts
```diff
@@ -1,6 +1,6 @@
-import fs from 'fs'
-import os from 'os'
-import path from 'path'
+import fs from 'node:fs/promises'
```
XhmikosR, line 1: Also switched to `node:fs/promises` here since we only use that.

src/package-managers/npm.ts
raineorshine: It's a bit of a judgment call. Which do you think is best?

1. Use `nodeSemver` everywhere for consistency.
2. Rename `semver: GetVersion` and use `semver` everywhere for consistency.
3. Allow the exception of `nodeSemver` just in `npm.ts` and live with the inconsistency.

XhmikosR: I clearly went with the third option :P

That being said, I'm not sure I like option 1, so maybe we just rename the exported function if it bothers you.

XhmikosR: Actually, we can't rename it because the value is coming from the CLI. I'd say we live with it...

raineorshine: > I clearly went with the third option :P

Just checking :)

raineorshine (approved):

Issue #1830, opened 2026-06-18
Drop Node.js 20 support?


XhmikosR:
@raineorshine I recently experimented with native TS loading which requires Node.js `22.18.0 || 24.3.0`.

So, without `tsx`, build is faster, build:options is **10x faster** on my VM on cold build, tests are also faster.

On top of that I played with mocha parallel and `test:unit` on CI is **~2x** faster and more on my VM (since I have more CPU cores).

* with native TS loading: https://github.com/raineorshine/npm-check-updates/actions/runs/27712671803/job/81978965610
* with mocha parallel: https://github.com/raineorshine/npm-check-updates/actions/runs/27774531098/job/82183109060

So, my question is, after landing the open PRs (no hurry) and cutting new versions, maybe the next version should be a major version bump so that we can use native TS loading?

raineorshine:
Yes, I'm more than happy with that. Thanks for looking into it.

Fwiw, my policy rather than dropping support for NodeJS versions as soon as they hit EOL is to wait until there is a functional, performance, or security reason to do so. Though with the more disciplined release scheduled, I could be convinced to align more closely with NodeJS. It was more wild west when this library was created a decade ago.

raineorshine:
We can upgrade `markdownlint-cli`, `npm-run-all2`, etc in the same major version bump.

XhmikosR:
I think after Node.js 26, versions will be LTS for 2 years. And given that we already have a clear benefit, I think it's a good time to drop Node.js 20.

I'll prepare a PR for that.

XhmikosR:
Just don't cut the new major version fast enough until we are done with the all changes like dropping tsx etc.

raineorshine:
I downgraded `ini` in #1845, so add that to the list. It slipped through prematurely.

XhmikosR:
Maybe we should enable strict engines on CI? It would have caught this.

onemen:
@raineorshine , @XhmikosR 

There are already security risk just by using any Node version after its end-of-life https://nodejs.org/en/about/eol

We do not need to wait until some vulnerability will hit `npm check updates` users.

XhmikosR:
ncu isn't the only Node.js package out there and it certainly isn't supposed to take users by the hand and teach them that they need to update to supported versions...

That being said, since we have a clear benefit from dropping 20, it'll be done soon after we are done with the less invasive changes.

XhmikosR:
[...]

PR #1844, opened 2026-06-20
Drop Node.js < 22 support and update major deps

XhmikosR:
I'll leave this as draft until we are done with the rest of the minor patches.

The other changes will be made when this is merged so don't cut a new major version immediately after merging this.

TODO:

- [ ] Adapt branch protection rules before/after merging this.

Refs #1830


package.json
```diff
@@ -109,44 +109,44 @@
...
     "mocha": "^11.7.6",
-    "npm-registry-fetch": "^19.1.1",
-    "npm-run-all2": "^8.0.4",
+    "npm-registry-fetch": "^20.0.1",
+    "npm-run-all2": "^9.0.2",
     "p-map": "^7.0.4",
     "parse-github-url": "^1.0.4",
     "picomatch": "^4.0.4",
-    "prettier": "^3.8.5",
+    "prettier": "3.8.5",
```
XhmikosR: I pinned prettier for now because I notice some inconsistent changes. We should unpin it later when the new versions have stabilized.

EDIT: actually the prettier developers suggest to have prettier pinned:

> A reminder, when Prettier is installed or updated, it’s [strongly recommended](https://prettier.io/docs/install#summary) to specify the exact version in package.json: "prettier": "3.9.0", not "prettier": "^3.9.0".

src/bin/cli.ts
```diff
@@ -148,7 +148,7 @@ ${chalk.dim.underline(
     // add --no- prefixed boolean options
     // necessary for overriding booleans set to true in the ncurc
     if (type === 'boolean') {
-      program.addOption(new Option(`--no-${long}`).default(false).hideHelp())
```
XhmikosR, line 151: This is the breaking change from commander.js v15:

>Breaking: only lone --no-* option sets default option value to true, default not implicitly set when define both positive and negative option in either order (https://github.com/tj/commander.js/pull/2405)

package.json
```diff
@@ -17,7 +17,7 @@
     "versions"
   ],
   "engines": {
-    "node": "^20.19.0 || ^22.12.0 || >=24.0.0",
+    "node": "^22.22.2 || ^24.15.0 || >=26.0.0",
```
raineorshine, line 20: Are these minor upgrades the minimum required?

- `^22.12.0` → `^22.22.2`
- `>=24.0.0` → `^24.15.0`

XhmikosR: Yeah, it's what npm-registry-fetch requires and its subpackages.

raineorshine (approved):

XhmikosR:
Just remember to adapt the branch protection rules because otherwise only you can merge :)

raineorshine:
Just did :)

XhmikosR:
I still see `Node 20 on ubuntu-latestExpected — Waiting for status to be reported`...

raineorshine:
Requires hitting the Save button apparently...

PR #1851, opened 2026-06-21
Move generic utils in lib/utils

XhmikosR:
This might conflict with #1842, ping me to rebase.

I went with the generic utils, we might want to re-organize the lib files later further in subfolders.


raineorshine:
Hi, thanks for the suggestion.

I prefer a flatter folder structure. Subfolders create ambiguities as not everything may clearly fall into one category. If there are unambiguous categories following certain architectural conventions (components, routes, middleware, etc) then I think subfolders are justified. Otherwise a flat organization structure means that contributors do not need to learn a particular folksonomy to know where to find things. Ubiquitous editor support for fuzzy matched Go to File and agentic development lessen the need for carefully curated directory structures anyway.

PR #1853, opened 2026-06-21
More shell tweaks

XhmikosR:
It needs a test on macOS since I don't use such platform myself.

### DIFF
--- a/.hooks/post-commit
+++ b/.hooks/post-commit
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+# Enable strict mode: 
+# -E: inherit ERR traps in functions/subshells
+# -e: exit on any command failure
+# -u: error on unset variables
+# -o pipefail: fail pipelines if any command fails
 set -Eeuo pipefail
 
 SHORT_SHA=$(git rev-parse HEAD)
--- a/.hooks/post-commit
+++ b/.hooks/post-commit
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
+
+set -Eeuo pipefail
+
 SHORT_SHA=$(git rev-parse HEAD)
-LINT_LOG="$TMPDIR"/lint."$SHORT_SHA".log
+LINT_LOG="${TMPDIR:-/tmp}/lint.${SHORT_SHA}.log"
 
 # strip color
 strip() {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -52,7 +52,7 @@
     "test": "tsc --noEmit && npm run test:unit && npm run test:e2e",
     "test:bun": "mocha test/bun",
     "test:unit": "mocha test --ignore \"test/bun/*\"",
-    "test:e2e": "./test/e2e.sh",
+    "test:e2e": "bash ./test/e2e.sh",
```
XhmikosR, line 55: This is so that it works on Windows too. I cannot actually test it since my dev VM is using WSL1 which is not supported, but it should work with WSL2.

.hooks/post-commit
```diff
@@ -10,28 +13,28 @@ strip() {
...
+    local ERRORS
+    ERRORS=$(sed 1,4d "${LINT_LOG}")
 
     # Trigger apple- or OSA-script on supported platforms
-    /usr/bin/osascript -e "display notification \"$ERRORS\" with title \"$*\""
+    /usr/bin/osascript -e "display notification \"${ERRORS}\" with title \"$*\""
   fi
-
-  # clean up
-  rm "$LINT_LOG"
```
XhmikosR, line 22: Moved this later on since it left the log file for non-macOS platforms.

.hooks/post-commit
```diff
@@ -10,28 +13,28 @@ strip() {
...
-set -o pipefail
-
 # Do NOT run this when rebasing or we can't get the branch
 branch=$(git branch --show-current)
-if [ -z "$branch" ]; then
+if [[ -z "${branch}" ]]; then
+  exit 0
+fi
+
+if [[ ! -d node_modules ]]; then
```
XhmikosR, line 37: Not perfect, but the script won't hang if one hasn't run `npm install`.

.hooks/post-commit
```diff
@@ -10,28 +13,28 @@ strip() {
...
-    /usr/bin/osascript -e "display notification \"$ERRORS\" with title \"$*\""
+    /usr/bin/osascript -e "display notification \"${ERRORS}\" with title \"$*\""
   fi
-
-  # clean up
-  rm "$LINT_LOG"
 }
 
-# ensure failed lint exit code passes through sed
-set -o pipefail
```
raineorshine, line 26: This line was moved but the comment got dropped. I would prefer to keep the comment since it is minimal and helps explain the behavior to those who may not be as familiar with shell scripting.

raineorshine (changes requested):
Looks good, thanks

.hooks/post-commit
XhmikosR: But if we go by that, I'd need to explain every set option...

XhmikosR: BTW https://explainshell.com/explain?cmd=set+-Eeuo+pipefail

test/e2e.sh
```diff
@@ -21,23 +20,23 @@ cleanup() {
   exit_status=$?
 
   # shut down verdaccio
-  if [ -n "${verdaccio_pid}" ]; then
+  if [[ -n "${verdaccio_pid}" ]]; then
     echo Shutting down verdaccio
-    kill -9 ${verdaccio_pid} 2>/dev/null
-    wait ${verdaccio_pid} 2>/dev/null
+    kill -9 ${verdaccio_pid} 2>/dev/null || true
```
XhmikosR, line 25: These `|| true` don't have any effect right now, but if/when `set -e` is enabled, we don't want these lines to fail.

.hooks/post-commit
raineorshine: A nitpick I know, but it's generally good form to preserve comments that are already in the codebase. 

I won't make you explain every set option :).

raineorshine (approved):

raineorshine:
> It needs a test on macOS since I don't use such platform myself.

Tested on macOS and works ✓

.hooks/post-commit
XhmikosR: I actually did explain all of them just to be safe :)

PR #1856, opened 2026-06-21
package.json: remove duplicate typecheck script

XhmikosR:
`npm run lint` already runs the `lint:types` script so it was running twice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
     "test:bun": "mocha test/bun",
     "test:unit": "mocha test --ignore \"test/bun/*\"",
     "test:e2e": "./test/e2e.sh",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -53,7 +53,6 @@
     "test:bun": "mocha test/bun",
     "test:unit": "mocha test --ignore \"test/bun/*\"",
     "test:e2e": "./test/e2e.sh",
-    "typecheck": "tsc --noEmit",
```
XhmikosR, line 56: @raineorshine: if you are used to using this script, I could add it back as an alias

raineorshine (approved):
I think `lint:types` was missed before. Thanks.

PR #1857, opened 2026-06-21
Remove the unused libnpmconfig.d.ts

XhmikosR:
I wonder what's the replacement for libnpmconfig nowadays? Maybe it won't be too hard to migrate and drop the custom vendored files from the repo.

### DIFF
--- a/src/types/libnpmconfig.d.ts
+++ b/src/types/libnpmconfig.d.ts
@@ -1,2 +1,0 @@
-// add to tsconfig compilerOptions.paths
-declare module 'libnpmconfig'
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -20,7 +20,6 @@
     "incremental": true,
     "tsBuildInfoFile": "node_modules/.cache/tsbuildinfo",
     "paths": {
-      "libnpmconfig": ["./src/types/libnpmconfig"],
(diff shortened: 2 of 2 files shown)


src/types/libnpmconfig.d.ts
```diff
@@ -1,2 +0,0 @@
-// add to tsconfig compilerOptions.paths
-declare module 'libnpmconfig'
```
raineorshine, line 2: This can be removed along with:

https://github.com/raineorshine/npm-check-updates/blob/509ee72373ea9b91f9ad9d6c63386f1f35bead95/tsconfig.json#L23

raineorshine (changes requested):
> I wonder what's the replacement for libnpmconfig nowadays? Maybe it won't be too hard to migrate and drop the custom vendored files from the repo.

Possible, but high risk since we don't have test coverage for all the various npm config values and how they affect the invocation of `npm-registry-fetch`. One misread config key (e.g. `strict-ssl` vs `strictSSL`) and it can break the npm fetch.

src/types/libnpmconfig.d.ts
XhmikosR: I removed those in #1858, it'll need a rebase after this PR is merged.

raineorshine (approved):
Thanks

PR #1858, opened 2026-06-21
Simplify tsconfig.json

XhmikosR:
* simplified includes/excludes
* remove the redundant noImplicitAny which is true with strict
* remove unneeded libs
* remove unneeded paths


tsconfig.json
```diff
@@ -15,13 +15,8 @@
     "resolveJsonModule": true,
```
XhmikosR, line 16: This is currently not needed, but I left it since it doesn't hurt.

raineorshine (approved):

PR #1861, opened 2026-06-21
Vite config cleanup

XhmikosR:


vite.config.ts
```diff
@@ -18,31 +18,35 @@ function buildOptionsPlugin(): Plugin {
...
   return {
     name: 'chmod-bin',
-    closeBundle() {
+    writeBundle(options) {
+      // only the ES round; the cjs round emits cli.cjs.
+      // chmodSync will throw if cli.js is missing so that we notice.
+      if (options.format !== 'es') return
       chmodSync('build/cli.js', 0o755)
+      // drop the empty `export {}` cli.d.ts dts emits for the export-less cli entry
+      rmSync('build/cli.d.ts', { force: true })
```
XhmikosR: Not 100% sure about this. I couldn't find a way to exclude this from dts and it's just a file with an empty export. I could drop the removal if you prefer that.

raineorshine: Seems fine

raineorshine (approved):
Thanks

PR #1865, opened 2026-06-23
cache test: remove any previous stale cache file to prevent failures

XhmikosR:
I only hit this a couple of times on my dev VM, I probably killed the process and the cache file wasn't cleaned up. Or I did run `ncu` on my dev VM between test runs.

Let me know if it doesn't make sense to delete the cache file before the cache tests, maybe I'm missing something.


raineorshine (approved):
Seems reasonable

PR #1869, opened 2026-06-24
Doctor tests cleanup

XhmikosR:
I think I got all the temp files now to be removed, I'll monitor later in case I missed something.


test/doctor.test.ts
```diff
@@ -363,6 +362,23 @@ else {
   })
 
   describe('yarn', () => {
+    // yarn classic drops yarn--<timestamp> proxy-script dirs in the temp dir and does not reliably clean them up
```
XhmikosR, line 365: I noticed this on my Windows VM. On Linux the behavior was different.

raineorshine (approved):

PR #1870, opened 2026-06-24
CI updates

XhmikosR:
Please check the individual commits

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -35,13 +35,6 @@ jobs:
           node-version: ${{ env.NODE }}
           cache: npm
 
-      - name: Cache tsbuildinfo
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: node_modules/.cache
-          key: tsbuildinfo-${{ runner.os }}-${{ hashFiles('**/tsconfig.json') }}
-          restore-keys: tsbuildinfo-${{ runner.os }}-
-
       - name: Install npm dependencies
         run: npm ci
 
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -1,0 +1,42 @@
+name: E2E Tests
+
+on:
+  push:
+    branches:
+      - main
+      - '!dependabot/**'
+  pull_request:
+    branches:
+      - '**'
+  workflow_dispatch:
+
+env:
+  FORCE_COLOR: 2
+  NODE: 24
+
+permissions:
+  contents: read
+
+jobs:
+  e2e:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    if: github.event_name != 'pull_request' || github.event.pull_request.changed_files > 0
+
+    steps:
+      - name: Clone repository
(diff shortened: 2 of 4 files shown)


.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: ${{ env.NODE }}
+          cache: npm
+
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: E2E Tests
+        run: npm run test:e2e
```
XhmikosR, line 42: Note that I haven't added the separate build step because prepublish does run build. Let me know if you want this changed, but previously we were running build twice.

.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
+
+      - name: Set up Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: ${{ env.NODE }}
```
XhmikosR, line 35: I only added Node.js 24 on Ubuntu, let me know if you want a different combination. The run does not take long, so we could probably add more cases in the matrix if you prefer. Just keep in mind, that Windows is always slower than Ubuntu.

.github/workflows/lint.yml
XhmikosR: I'm still not sure this is worth it, i.e. if it has any meaningful impact. It probably takes more time to cache and restore than the run itself...

.github/workflows/build.yml
```diff
@@ -21,7 +21,6 @@ jobs:
   build:
     runs-on: ubuntu-latest
     timeout-minutes: 15
-    if: github.event_name != 'pull_request' || github.event.pull_request.changed_files > 0
```
raineorshine, line 24: This is used to prevent unnecessary builds when Copilot kicks offs its session by opening an empty draft PR.

.github/workflows/lint.yml
raineorshine: I don't have any particular experience with this, so I'll follow your recommendation.

.github/workflows/test-e2e.yml
raineorshine: I'm fine with that for the e2e tests.

raineorshine (changes requested):

.github/workflows/test-e2e.yml
raineorshine: That should be fine, thanks.

.github/workflows/build.yml
XhmikosR: OK, I restored this and added it to the new workflows too.

.github/workflows/lint.yml
XhmikosR: Doesn't look like it hurts the lint step and the total workflow time seems to be reduced as expected.

raineorshine (approved):

PR #1871, opened 2026-06-24
fix: skip cache write when cooldown is active

XhmikosR:
The read was already skipped, so a cooldown fallback poisoned later non-cooldown runs.


raineorshine (approved):
Good catch, thanks.

PR #1891, opened 2026-06-28
Use async glob instead of glob.sync in getAllPackages

XhmikosR:
Not a huge difference, I suppose, but the functions are already async.


raineorshine (approved):
Yes, might as well. Thanks.

PR #1901, opened 2026-06-29
Abort packument stream once required fields are parsed

### LINKED ISSUE #1351: Very high bandwidth usage in monorepo

---

## Steps to Reproduce

.ncurc:

```js
{
    "format": "group",
    "root": ".",
    "upgrade": true,
    "workspaces": true
}

```

Steps:

I've have a monorepo that has a main react package and some utility packages with the sames dependencies: @types/node, eslint, vitest and typescript.

I used [AppNetworkCounter](https://www.nirsoft.net/utils/app_network_counter.html) to monitor the bandwidth usage.

## Current Behavior

My guess is, that this script re-checks every dependency even if it was already checked before. That is, if I have 4 packages in my monorepo each with `@types/node`, it checks it four times instead of caching it. As I wanted to update my monorepo, the bandwidth shot in the air with like 3MB/s and 1 or 2MB of requests and file reading for each dependency.

## Expected Behavior

- Find a way to reduce the needed bandwidth for each dependency
- Cache duplicate packages

XhmikosR:
Closes #1351

I had to use AI to create a script to measure the bytes transferred, but it went down from **1.6MB** to **1.3KB** for typescript (it probably is gzipped). So the logic was correct, the abort controller was missing, it seems.

I could definitely use another set of eyes to confirm my findings just so that we are 100% sure it indeed fixes the issue.

### DIFF
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -91,7 +91,8 @@ const fetchPartialPackument = async (
       return npmRegistryFetch.json(url.href, fetchOptions)
     } else {
       tag = tag || 'latest'
-      const response = await npmRegistryFetch(url.href, fetchOptions)
+      const controller = new AbortController()
+      const response = await npmRegistryFetch(url.href, { ...fetchOptions, signal: controller.signal })
       const parser = new JSONParser({ paths: ['$.*'], keepStack: false })
       const partialPackument: Partial<Packument> = { name }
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
I'm going to merge now as this is an easy, clean patch. We can reopen #1351 if this doesn't happen to fix that issue completely.

PR #1902, opened 2026-06-29
Fix cli options being overridden by .ncurc in --deep mode

### LINKED ISSUE #1355: cli options don't override the default options in .ncurc.js

---
when you set default options in .ncurc.js, these options don't be overridden by those that provided via cli

for example if you have a .ncurc.js like this

```js
module.exports = {
  target: "minor"
  ...
}
```

then you triggered the following command 

```
ncu --target=latest
```

then, the minor target is used

XhmikosR:
Fixes #1355

Requires #1895

### DIFF
--- a/src/bin/cli.ts
+++ b/src/bin/cli.ts
@@ -219,6 +219,12 @@ ${chalk.dim.underline(
     }
   }
 
+  // track which options were set on the command line (not the rc config) so they keep
+  // precedence over per-package configs reloaded in --deep mode. getOptionValueSource reflects
+  // the earlier program.parse(process.argv) and lets commander handle combined short options
+  // (e.g. -gf) instead of parsing them here.
+  const cliKeys = Object.keys(programOpts).filter(key => program.getOptionValueSource(key) === 'cli')
+
   // See defaultOptionValues comment above
   ;(program as any)._optionValues = defaultOptionValues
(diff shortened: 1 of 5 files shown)


src/bin/cli.ts
```diff
@@ -221,6 +221,20 @@ ${chalk.dim.underline(
     }
   }
 
+  // track which options were set on the command line (not the rc config) so they keep
+  // precedence over per-package configs reloaded in --deep mode.
+  const cliKeys: string[] = []
+  for (let i = 0; i < programArgs.length; i++) {
+    const arg = programArgs[i]
+    if (typeof arg !== 'string' || !arg.startsWith('-')) continue
+    const option = cliOptionsMap[arg.replace(/^--?(no-)?/, '')]
```
raineorshine: This will break on combined short options:

e.g.
`ncu -gf npm` == `ncu -g -f npm`

It would be best to leave the parsing to commander anyway. Partially recreating the options parsing logic means that different parsing methodologies live side-by-side and can easily lead to errors like the one above. 

Can you use `program.opts()`?

raineorshine (changes requested):

XhmikosR:
@raineorshine: Can you check again please?

raineorshine (approved):

PR #1906, opened 2026-07-01
Drop jsonlines dependency in favor of inline NDJSON parsing

XhmikosR:
Not a big dependency, but this seems simpler.

Non-whitespace diff: https://github.com/raineorshine/npm-check-updates/pull/1906/changes?w=1


raineorshine (approved):
👍

PR #1907, opened 2026-07-01
Fix major/minor/patch grouping when a range operator changes

chatman-media:
`partChanged` and `colorizeDiff` strip the leading `^`/`~` before comparing versions, but only when `from` and `to` start with the same character. That assumption breaks for `<` and `<=` declarations, since `upgradeDependencyDeclaration` normalizes those into a `^` range. For example `"<4.17.20"` upgrading to `"^4.18.1"` is really a minor bump, but the leftover `<` on `from` blocks the wildcard stripping, so the whole thing gets diffed as `major` and shown in red under `--format group`.

I switched both functions to strip the operator off each side independently, reusing the `stripRange` helper that already existed for this exact regex. Added a test for each function with the `<`/`<=` → `^` case, and confirmed it fails on the current code and passes after the fix. Also checked against a real `lodash` upgrade with the CLI to see the group heading flip from "Major" to "Minor".

### DIFF
--- a/src/lib/version-util.ts
+++ b/src/lib/version-util.ts
@@ -158,6 +158,9 @@ export function isWildPart(versionPartValue: Maybe<string>) {
   return versionPartValue === '*' || versionPartValue === 'x'
 }
 
+/** Strips semver range prefixes (^, ~, >=, <=, >, <) from a version string. */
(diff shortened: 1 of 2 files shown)


raineorshine (approved):
Thanks!

PR #1918, opened 2026-07-06
Increase coverage

XhmikosR:
These are mostly AI-generated,.

I tried to focus on the more meaningful tests and not just cover everything.


raineorshine (approved):
Seems reasonable, thanks

PR #1919, opened 2026-07-06
Inline packageJson.version in toDependencyTable

XhmikosR:


raineorshine (changes requested):
If I'm seeing this correctly, `getPackageVersion` is only used by `toDependencyTable`, so I suggest removing it completely and just using `packageJson.version`.

XhmikosR:
> If I'm seeing this correctly, `getPackageVersion` is only used by `toDependencyTable`, so I suggest removing it completely and just using `packageJson.version`.

Good catch, BTW, I didn't even see the redundancy :)

raineorshine (approved):

PR #1922, opened 2026-07-07
Preserve YAML content outside the top-level map when upgrading catalogs

XhmikosR:
@raineorshine: not sure if this should `close` #1540, if it's not already fixed.


raineorshine (approved):

raineorshine:
I'm not able to go into the details on this one, but feel free to post to #1540 and see if this issue fixes the reported problem. Thanks.

PR #1928, opened 2026-07-09
Update dependencies

XhmikosR:
* @types/node                        ^26.1.0  →   ^26.1.1
* @typescript-eslint/eslint-plugin   ^8.62.1  →   ^8.63.0
* @typescript-eslint/parser          ^8.62.1  →   ^8.63.0
* eslint                             ^10.6.0  →   ^10.7.0
* eslint-plugin-jsdoc               ^63.0.11  →  ^63.0.13
* eslint-plugin-n                    ^18.2.1  →   ^18.2.2
* prettier                             3.9.4  →     3.9.5
* vite                                ^8.1.3  →    ^8.1.4
* vitest                              ^4.1.9  →   ^4.1.10

Note that typescript 7.x is out but that should be handled separately, see https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/#running-side-by-side-with-typescript-6.0

Maybe we'd need to wait until v7.1 since right now it doesn't ship with an API and we'd need to install v6.x simultaneously. I'd personally just ignore typescript 7 in dependabot for now.


raineorshine (approved):
> Maybe we'd need to wait until v7.1 since right now it doesn't ship with an API and we'd need to install v6.x simultaneously. I'd personally just ignore typescript 7 in dependabot for now.

I'm fine ignoring it until it supports a clean switch.

PR #1929, opened 2026-07-09
Fall back to package.json with -p deno when no deno.json is found

### LINKED ISSUE #1462: Deno 2.0 support
Deno 2.0 support managing dependencies in `package.json`, but using `ncu` with `-p deno` failes when it doesn't find `deno.json` with confusing error
```
No package.json

Please add a package.json to the current directory, specify the --packageFile or --packageData options, or pipe a package.json to stdin and specify --stdin.
```
It should probably fallback to using `package.json`.

XhmikosR:
Fixes #1462

### DIFF
--- a/src/lib/findPackage.ts
+++ b/src/lib/findPackage.ts
@@ -76,14 +76,16 @@ async function findPackage(options: Options): Promise<{
     pkgData = data || getPackageDataFromFile(pkgFile, pkgPath)
   } else {
     // find the closest package starting from the current working directory and going up to the root
-    pkgFile = pkgPath
-      ? await findUp(
-          !options.packageFile && options.packageManager === 'deno' ? ['deno.json', 'deno.jsonc'] : pkgPath,
-          {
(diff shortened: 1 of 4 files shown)


src/lib/findPackage.ts
```diff
@@ -76,14 +76,16 @@ async function findPackage(options: Options): Promise<{
...
-          },
-        )
-      : null
+    const cwd = options.cwd || process.cwd()
+    if (!pkgPath) {
+      pkgFile = null
+    } else if (!options.packageFile && options.packageManager === 'deno') {
+      // Deno 2.0 also supports managing dependencies in package.json,
+      // so fall back to it when no deno.json/deno.jsonc is found.
+      pkgFile = (await findUp(['deno.json', 'deno.jsonc'], { cwd })) || (await findUp(pkgPath, { cwd }))
```
raineorshine, line 85: Good fallback semantics. Matches deno exactly.

raineorshine (approved):

PR #1931, opened 2026-07-14
CI: further tighten workflows

XhmikosR:


.github/workflows/release.yml
```diff
@@ -21,6 +24,6 @@ jobs:
           persist-credentials: false
 
       - name: Release
-        uses: softprops/action-gh-release@718ea10b132b3b2eba29c1007bb80653f286566b # v3.0.1
-        with:
-          generate_release_notes: true
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh release create "$GITHUB_REF_NAME" --generate-notes
```
XhmikosR, line 29: This should have the same effect but I cannot test it. If it breaks something, we can fix it after a new tag is created.

raineorshine (approved):

PR #1932, opened 2026-07-15
Clean up and enforce type exports

XhmikosR:
Not sure which one you prefer or if we should even go with this @raineorshine.

That being said, might be a good chance to revisit/review that the exports are OK before we proceed with the next version bump.


raineorshine (approved):
I don't have a preference, so I'll defer to you on this.

XhmikosR:
OK, let's give it a go and we can revisit it later, if needed.

PR #1934, opened 2026-07-15
Replace libnpmconfig and figgy-pudding with an in-house npm config loader

XhmikosR:
figgy-pudding code was already stripped down in 036be7f03ecaca8f8f9e007c9c740e2051dbd597. So, this is a natural continuation to my eyes, assuming we test everything properly.

I tried to cover all cases with new tests. Note that I tried to switch to `@npmcli/config`, but that resulted in a huge bundle size bump. I'm not aware of another package that we could use instead of the manual way.

Notes:

1. I'm not sure if we should change the internal functions to use async fs methods; it probably does not matter a lot
2. Not sure if we should adapt the return types of the touched functions (`null` vs `{}`)
3. The new tests might not be perfect, let me know if you have any better ideas


raineorshine:
It would be nice to be free of the legacy vendored dependency, but note that this is a high risk, low reward change. npm config reading is sensitive and it's hard to have good test coverage since we pass through most config variables as-is. i.e. it depends on the user's machine. 

For that reason, I would be most comfortable with the following release strategy:

1) first release on the `next` tag and solicit testers.
2) release to `latest` as its own patch release (after the next major version bump) so that we are prepared to quickly patch or revert if the community discovers any regressions.

I don't have strong opinions about the test coverage or syntax since it's not an active part of the codebase.

raineorshine (approved):

XhmikosR:
Sounds good! Let's wait after v23.0.0 is out then.

XhmikosR:
@raineorshine: when do you plan to cut a next tag? I don't have any other breaking changes, at least, I think.

PR #1935, opened 2026-07-15
Fix dependency section regexes and fix packages not upgrading in all sections when versions differ

### LINKED ISSUE #1594: Not all versions of a package are updated if the package is listed multiple times
- [X] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)

---

## Steps to Reproduce

.ncurc:

No `.ncurc` file is used.

Dependencies:

```json
{
    "dependencies": {
        "fudgel": "2.6.0"
    },
    "peerDependencies": {
        "fudgel": "2.6.1"
    }
}
```

Steps:

1. Start with two different versions of the same package.
2. Run `npm-check-updates -u --dep dev,optional,peer,prod,packageManager`

## Current Behavior

Only one of the packages is updated to the latest instead of both.

## Expected Behavior

Both packages are updated

## Sample Output

```
$ cat package.json
{
    "dependencies": {
        "fudgel": "2.6.0"
    },
    "peerDependencies": {
        "fudgel": "2.6.1"
    }
}
```

```
$ npm-check-updates -u --dep dev,optional,peer,prod,packageManager
Upgrading /[removed]/package.json
[====================] 1/1 100%

 fudgel  2.6.0  →  3.1.0

Run npm install to install new versions.
```

```
$ cat package.json
{
    "dependencies": {
        "fudgel": "3.1.0"
    },
    "peerDependencies": {
        "fudgel": "2.6.1"
    }
}
```

```
$ npm-check-updates -u --dep dev,optional,peer,prod,packageManager
Upgrading /[removed]/package.json
[====================] 1/1 100%

 fudgel  2.6.1  →  3.1.0

Run npm install to install new versions.
```

```
$ cat package.json
{
    "dependencies": {
        "fudgel": "3.1.0"
    },
    "peerDependencies": {
        "fudgel": "3.1.0"
    }
}
```

XhmikosR:
I started looking at the regexes to figure out how to fix #1594 here: https://github.com/raineorshine/npm-check-updates/compare/main...xmr/regex. Then I realized we could/should use jsonc-parser instead of the fragile regexes so I made this branch.

Fixes #1594.

This probably shouldn't be squashed when merging.


test/dep.test.ts
```diff
@@ -197,6 +197,108 @@ describe('--dep', () => {
...
+            'ncu-test-v2': '^99.9.9',
+          },
+        })
+      } finally {
+        await removeDir(tempDir)
+        stub.restore()
+      }
+    })
+
+    it('do not downgrade a section that is already newer than the target', async () => {
```
raineorshine: I don't think this test belongs here, as it is not a general rule and depends on the `--target`. e.g. Targeting specific tags will downgrade versions by design.

raineorshine (commented):
Yes, this is the right way to do this. There was no other way to enable section-specific upgrades without a better foundation. Thanks!

test/dep.test.ts
XhmikosR: Good point, removed it.

raineorshine (approved):

PR #1941, opened 2026-07-17
Fix and clean up skipped tests

XhmikosR:


test/getCurrentDependencies.test.ts
```diff
@@ -128,12 +128,6 @@ describe('getCurrentDependencies', () => {
       })
     })
 
-    it.skip('should filter org dependencies by regex', () => {
-      expect(getCurrentDependencies(deps, { filter: /store/ })).toStrictEqual({
-        '@ngrx/store': '4.0.0',
-      })
-    })
```
raineorshine, line 135: I think this functionality is still desired and should be either implemented or left in as a placeholder. 

What do you think?

raineorshine (commented):
Thanks

test/getCurrentDependencies.test.ts
XhmikosR: The functionality is there, the test was broken. I'll fix it.

raineorshine (approved):

PR #1942, opened 2026-07-19
Skip peers cache when cooldown is active

### LINKED ISSUE #1722: `peers` data missing timestamps when restored from cache with --cooldown
### Description

The recent enhancements to `--format time` (introduced in PR #1693) implemented consistent caching and human-readable timestamps. However, the `peers` data type was omitted during the update. Currently, when peer dependency information is restored from the cache, it lacks the associated timestamp, leading to inconsistent output compared to standard dependencies.

### Steps to Reproduce

1. Run `ncu --cache --cooldown 5 --peer --format time` on a project with peer dependencies to prime the cache.
2. Run the command again to trigger a cache restore.
3. Observe that while standard dependencies display human-readable timestamps, the peer dependencies do not.

### Expected Behavior

Peer dependency data should include a `time` metadata field when cached and restored, matching the implementation for regular dependencies and devDependencies.

@raineorshine

When we are using `cooldown` we skip cache for `cacher.get` but not for `cacher.getPeers`

XhmikosR:
Fixes #1722

### DIFF
--- a/src/lib/getPeerDependenciesFromRegistry.ts
+++ b/src/lib/getPeerDependenciesFromRegistry.ts
@@ -58,7 +58,7 @@ async function getPeerDependenciesFromRegistry(packageMap: Index<Version>, optio
   if (!packageManager.getPeerDependencies) return {}
 
   const numItems = Object.keys(packageMap).length
-  let bar: ProgressBar
+  let bar: ProgressBar | undefined
   if (!options.json && options.loglevel !== 'silent' && options.loglevel !== 'verbose' && numItems > 0) {
     bar = new ProgressBar('[:bar] :current/:total :percent', { total: numItems, width: 20 })
     bar.render()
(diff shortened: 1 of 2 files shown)


src/lib/getPeerDependenciesFromRegistry.ts
```diff
@@ -58,7 +58,7 @@ async function getPeerDependenciesFromRegistry(packageMap: Index<Version>, optio
   if (!packageManager.getPeerDependencies) return {}
 
   const numItems = Object.keys(packageMap).length
-  let bar: ProgressBar
+  let bar: ProgressBar | undefined
```
XhmikosR, line 61: This is unrelated but it was inconsistent with the other two places.

XhmikosR:
Actually, I'm not sure if we need to do this at all.

The peers cache is keyed by name + exact version, and a published version's peerDependencies are immutable. So, it'll be worse than before. Correct me if I'm wrong @raineorshine.

Another solution would be to revamp caching and cache the raw registry data and resolve at read time. The cache file increase is something I cannot estimate though.

EDIT: the cache size increase is insane with the second approach. For the ncu repo, it goes from ~4KB to ~116MB since the raw packument data is huge.
EDIT2: I ended up going with this: https://github.com/raineorshine/npm-check-updates/compare/main...xmr/cache. I'm still not sure if we really want to go that way since cache still grows a lot, but on demand, let me know if you prefer that, otherwise, I think we can drop this PR and the linked issue.

raineorshine:
That seems more reasonable, though `versions` can get a bit large.

@rbnayax or @CreativeTechGuy might have an opinion. I generally stay out of `--peer` related issues.

CreativeTechGuy:
I'm not sure I have enough context to add much here. But in theory the idea seems sensible.

XhmikosR:
I'm going to close this PR and the linked issue. I'll open the cache PR later.

PR #1947, opened 2026-07-24
Pin Docker base image by digest

XhmikosR:
Unsure if we want this... Sure, it's better practice, and we can get dependabot to update the hash on a weekly basis (we can move to monthly if you prefer). But it's extra churn.


raineorshine (approved):
As long as it's synced up with other dependency updates it won't add additional burden.

Thanks!

PR #1957, opened 2026-07-27
Shorten long build metadata in the upgrades table

### LINKED ISSUE #1539: The display effect of the version number containing build metadata needs to be optimized when the screen width is small, and it does not look clear enough now.
Image

Image

XhmikosR:
Fixes #1539.

Should we also shorten the `to` column too?

### DIFF
--- a/test/version-util.test.ts
+++ b/test/version-util.test.ts
@@ -426,6 +426,27 @@ describe('version-util', () => {
     })
   })
 
+  describe('shortenBuildMetadata', () => {
+    it('truncate long build metadata', () => {
+      expect(
+        versionUtil.shortenBuildMetadata(
+          '10.14.0+sha512.ad27a79641b49c3e481a16a805baa71817a04bbe06a38d17e60e2eaee83f6a146c6a688125f5792e48dd5ba30e7da52a5cda4c3992b9ccf333f9ce223af84748',
+        ),
+      ).toBe('10.14.0+sha512.ad27a79641b49...')
+    })
+    it('keep short build metadata', () => {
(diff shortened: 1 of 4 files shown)


test/version-util.test.ts
```diff
@@ -426,6 +426,24 @@ describe('version-util', () => {
     })
   })
 
+  describe('shortenBuildMetadata', () => {
+    it('cut off long build metadata', () => {
+      expect(
+        versionUtil.shortenBuildMetadata(
+          '10.14.0+sha512.ad27a79641b49c3e481a16a805baa71817a04bbe06a38d17e60e2eaee83f6a146c6a688125f5792e48dd5ba30e7da52a5cda4c3992b9ccf333f9ce223af84748',
+        ),
+      ).toBe('10.14.0+...')
```
raineorshine: I think it should keep all the characters up to the truncation limit. e.g. `10.14.0+sha512.ad2...`

raineorshine (changes requested):
> Should we also shorten the `to` column too?

Yes, thanks

test/version-util.test.ts
XhmikosR: OK, fixed both.

XhmikosR:
Actually, I'll revisit this later, do not merge yet please.

EDIT: I pushed an alternative approach. Both should be fine, the latter is more verbose. Let me know which way you prefer.

raineorshine (changes requested):
This approach looks good to me. Feel free to merge.

raineorshine (approved):

PR #1936, opened 2026-07-15
Replace timeago.js with the native Intl.RelativeTimeFormat

XhmikosR:
Not sure about this one, TBH. But since I had created the branch,I decided to open a PR. :)

PR #1960, opened 2026-07-28
Lazily construct the RelativeTimeFormat to avoid eager ICU init on startup

XhmikosR:
Follow up to #1936 

Missed this there.

| case | 23.0.0 | main | this branch |
|---|---|---|---|
| API ESM | 61.9 (min 57.6) | 74.1 (min 69.9) | 62.1 (min 58.5) |
| CLI ESM | 102.0 (min 94.4) | 104.9 (min 98.6) | 101.3 (min 95.4) |
| node baseline | 27.9 (min 24.3) | 27.9 (min 23.5) | 27.9 (min 24.5) |

Your call if you think it's worth it, I just noticed load time had regressed compared to v23.0.0.


raineorshine (approved):
Yes, thanks

PR #1961, opened 2026-07-29
Further e2e.sh improvements

XhmikosR:
See the individual commits.

Would be nice if you could test it on macOS since the script was broken there after f4869532. I plan to change the E2E CI workflow to run on macOS later too.

I manually tested this on Fedora, and CI is also green.

### DIFF
--- a/test/e2e.sh
+++ b/test/e2e.sh
@@ -118,7 +118,7 @@ fi
 
 # Test: ncu -v
 echo ncu -v
-npx --registry "${registry_local}" npm-check-updates -v
+npx --yes --registry "${registry_local}" npm-check-updates -v
 
 # Test: cli
 # Create a package.json file with a dependency on npm-check-updates since it is already published to the local registry
@@ -134,7 +134,7 @@ EOF
 # --configFilePath to avoid reading the repo .ncurc
 # --cwd to point to the temp package file
 # --pre 1 to ensure that an upgrade is always suggested even if npm-check-updates is on a prerelease version
-npx --registry "${registry_local}" npm-check-updates \
+npx --yes --registry "${registry_local}" npm-check-updates \
   --configFilePath "${temp_dir}" \
   --cwd "${temp_dir}" \
   --pre 1 \
--- a/test/e2e.sh
+++ b/test/e2e.sh
@@ -18,25 +18,25 @@ verdaccio_config="${temp_dir}/verdaccio-config.yaml"
 
 verdaccio_pid=""
 
-# cleanup on exit
+# Cleanup on exit
 cleanup() {
   local exit_status=$?
 
-  # shut down verdaccio
+  # Shut down Verdaccio
   if [[ -n "${verdaccio_pid}" ]]; then
-    echo Shutting down verdaccio
+    echo Shutting down Verdaccio
     kill -9 "${verdaccio_pid}" 2>/dev/null || true
     wait "${verdaccio_pid}" 2>/dev/null || true
   fi
 
-  # delete authToken
+  # Delete authToken
   # WARNING: The original authToken cannot be restored because it is protected and cannot be read with 'npm config get'.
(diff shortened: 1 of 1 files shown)


test/e2e.sh
```diff
@@ -67,47 +88,39 @@ uplinks:
...
-  if npm view "npm-check-updates@${package_version}" version --registry "${registry_local}" >/dev/null 2>&1; then
-    publish_visible=true
-    break
-  fi
-  sleep 1
-done
-
-if [[ "${publish_visible}" == false ]]; then
-  echo "Warning: npm-check-updates@${package_version} not visible in local registry after retry window"
+if ! retry npm view "npm-check-updates@${package_version}" version --registry "${registry_local}" >/dev/null 2>&1; then
```
XhmikosR, line 116: Not sure if the retry is even needed nowadays. From this PR run I see this:

```
2026-07-29T09:05:21.4439478Z Waiting for npm-check-updates@23.0.0 in local registry
2026-07-29T09:05:21.6358462Z ncu -v
```

Maybe this is an old issue that no longer applies?

We can try dropping it later.

test/e2e.sh
```diff
@@ -16,27 +17,28 @@ registry_local="http://${registry_addr}"
...
   if [[ -n "${verdaccio_pid}" ]]; then
-    echo Shutting down verdaccio
+    echo Shutting down Verdaccio
     kill -9 "${verdaccio_pid}" 2>/dev/null || true
     wait "${verdaccio_pid}" 2>/dev/null || true
   fi
 
-  # delete authToken
-  # WARNING: The original authToken cannot be restored because it is protected and cannot be read with 'npm config get'.
-  npm config delete "//${registry_addr}/:_authToken" || true
```
raineorshine, line 34: Is this safe to remove? Is there any record from `git blame` that shows why it was added to begin with?

test/e2e.sh
```diff
@@ -48,7 +50,26 @@ cleanup() {
 
 trap 'cleanup' EXIT
 
-# create verdaccio config
+# Used instead of `timeout`, which is not available on macOS
```
raineorshine, line 53: Interesting. `timeout` came preinstalled on my macOS in `/usr/local/bin/timeout`. Apparently that's not universal.

test/e2e.sh
raineorshine: I would recommend doing a bit of investigation with git blame to try to better understand its origins before concluding that it's no longer needed.

test/e2e.sh
XhmikosR: It's safe because the config used now is in the temp folder the script creates; we don't touch the system .npmrc at all, see https://github.com/raineorshine/npm-check-updates/pull/1961/changes/0637ab89e933a2dc4ba24dd97940e82f9242f9e9

But please confirm because that's the whole idea of that patch :)

test/e2e.sh
XhmikosR: Yeah, I did that hence my suggestion... And since tests have been reworked, we could give it a go later. I didn't remove it in this PR.

Let me know if you want me to remove it here and see how it goes. But it'd be easier to revert if it's a separate PR.

test/e2e.sh
XhmikosR: It seems it's not standard for macOS judging from the CI environment. That being said, do give this PR a go locally on macOS if you can. This solution is posix compliant but it's more code that's all.

test/e2e.sh
raineorshine: Ah, I see. Yes, if it seems safe we can remove it in a separate PR.

test/e2e.sh
raineorshine: That makes sense, thanks

test/e2e.sh
raineorshine: Confirmed

raineorshine (approved):

PR #1963, opened 2026-07-29
Fix Wildcard case

XhmikosR:


raineorshine (approved):

XhmikosR:
Missed two local variables, I'll change them later

PR #1966, opened 2026-08-02
Cache slim packuments keyed by name and resolve at read time

XhmikosR:
I'm still not sure if the cache size increase is acceptable or not. Based on the package cache can increase a lot.

### DIFF
--- a/src/lib/cache.ts
+++ b/src/lib/cache.ts
@@ -4,7 +4,9 @@ import path from 'node:path'
 import { CURRENT_CACHE_SCHEMA, type CacheData, type Cacher } from '../types/Cacher.ts'
 import { type Index } from '../types/IndexType.ts'
 import { type Options } from '../types/Options.ts'
+import { type Packument } from '../types/Packument.ts'
 import { type Version } from '../types/Version.ts'
+import { keyValueBy } from './keyValueBy.ts'
 import { print } from './logging.ts'
 
 export const CACHE_DELIMITER = '___'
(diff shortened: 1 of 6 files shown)


raineorshine:
What problem are you trying to solve? What are the tradeoffs?

XhmikosR:
This branch is the alternative one for #1942.

#1722 is really about how the cache stores data. It caches the resolved version per `name+target`, which can't represent a cooldown window, so cooldown runs bypass the version cache entirely and always refetch.

This PR caches the raw packument (versions + time) keyed by name and resolves at read time, so every target and cooldown come from one entry. The peers asymmetry the issue points at is harmless on its own (peers are keyed by exact version and immutable, so they're already safe to cache under cooldown), but caching the version data is what actually lets cooldown runs stop refetching.

The cost is a bigger cache than a single resolved version per target. I slimmed it to only the fields the resolvers read, and only fetch time/version lists when the target needs them. On this repo's 67 deps: 13 KB for a default run, 1.2 MB for `--target greatest`, 2.1 MB for `--cooldown`. There's also a schema bump (old caches get discarded) and entries refetch if they're missing a field the target needs.

So, we can also just drop this PR too, and leave things as is. I just looked at alternative solutions. Personally I don't even use cache myself.

raineorshine:
Thanks. While it is a big % increase in cache size compared to current `main`, it is in absolute terms quite small. 

Since this only impacts users who use the cache, and I can't imagine 2.1MB breaking anything too badly, this seems relatively safe to me. That said, it adds a bit of complexity so I would be perfectly fine dropping and avoiding the maintenance burden unless someone else has a strong opinion. Thanks.

PR #1968, opened 2026-08-02
Update overrides to fix the remaining npm vulnerabilities

XhmikosR:
Naming both parents is required; npm 10 ignores nested rules that skip a level

Should be dropped if/when verdaccio updates those packages.

After #1967 and this PR we should be down to zero vulnerabilities. Not that it's so important since those were mostly devDependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2657,22 +2657,6 @@
       "dev": true,
       "license": "MIT"
     },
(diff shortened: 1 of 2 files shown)


XhmikosR:
It's no longer needed after the recent updates in #1967.

Issue #2, opened 2013-09-09
Option to check for updates to global modules


pesho:
Add an option (-g or --global, like in npm) to check for updates to the globally installed modules.

PR #1973, opened 2026-08-03
feat(pnpm): read registries from pnpm-workspace.yaml

### LINKED ISSUE #1964: Feature request: copy the registry from the package manager's settings (`registries` in `pnpm-workspace.yaml`)
- [x] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues): #1605
- [x] I am using the [latest version](https://www.npmjs.com/package/npm-check-updates?activeTab=versions) of `npm-check-updates`
- [x] I am using a [compatible version of node](https://github.com/raineorshine/npm-check-updates/blob/main/package.json#L20)

---

Hey 👋🏻 thanks for your work on ncu !

Same idea as #1605 / #1639 for the cooldown, but for the registry.

pnpm 11 added [`registries`](https://pnpm.io/settings#registries) to `pnpm-workspace.yaml` ([pnpm/pnpm#11189](https://github.com/pnpm/pnpm/pull/11189)) — `default` is the workspace-file equivalent of the `registry` `.npmrc` setting, and wins over `.npmrc` when defined ([pnpm/pnpm#11754](https://github.com/pnpm/pnpm/pull/11754)):

```yaml
registries:
  default: https://my-private-registry.example/
  '@my-org': https://my-org-registry.example/
```

Would be nice if ncu picked that up when the package manager is pnpm, the way `minimumReleaseAge` already is.

## Steps to Reproduce

.ncurc: N/A

Dependencies:

```json
{
  "dependencies": {
    "husky": "9.1.0",
    "@types/semver": "7.5.0"
  }
}
```

Steps:

1. `pnpm-workspace.yaml` next to it:

   ```yaml
   registries:
     default: https://bogus-registry.invalid/
     '@types': https://bogus-registry.invalid/
   minimumReleaseAge: 4320
   ```

2. Run `npx npm-check-updates@23 -p pnpm --target minor` (with `NPM_CONFIG_REGISTRY` unset, so no registry leaks in from the environment)

## Current Behavior

`minimumReleaseAge` is read from `pnpm-workspace.yaml`, `registries` is not — neither `default` nor the scoped entry — and the fetches go to `https://registry.npmjs.org`:

```console
$ npx npm-check-updates@23 -p pnpm --target minor
Using minimumReleaseAge from pnpm-workspace.yaml: 3 days
Checking /tmp/repro/package.json


 husky  9.1.0  →  9.1.7

Minor   Backwards-compatible features
 @types/semver  7.5.0  →  7.7.1

Run npx npm-check-updates -p pnpm --target minor -u to upgrade package.json
```

`pnpm` does resolve against those settings — `pnpm add lodash` → `GET https://bogus-registry.invalid/lodash` — and the same URLs in `.npmrc` (`registry=` / `@types:registry=`) are honored by ncu (run from an already-installed ncu, since `npx` itself can no longer reach a registry in that case):

```console
 husky          FetchError: request to https://bogus-registry.invalid/husky failed, reason: getaddrinfo ENOTFOUND bogus-registry.invalid
 @types/semver  FetchError: request to https://bogus-registry.invalid/@types%2Fsemver failed, reason: getaddrinfo ENOTFOUND bogus-registry.invalid
```

So only the workspace file is left out. Against a private or proxied registry ncu keeps working, but reports versions, deprecations and `--cooldown` dates from the public one.

## Expected Behavior

`registries.default` (and scoped entries) used as registry when the package manager is pnpm, with pnpm's precedence — workspace file over `.npmrc`, `--registry` and `.ncurc` still winning — and ideally a log line like the cooldown one: `Using registry from pnpm-workspace.yaml: …`.

Workaround meanwhile is duplicating the URL in `.ncurc` next to `registries.default` and keeping both in sync by hand:

```js
// .ncurc.mjs
export default {
  registry: 'https://my-private-registry.example/',
};
```

Environment: `npm-check-updates@23.0.0`, `pnpm@11.9.0` (`registries` needs >= 11.0.0), node 22.22.2.

Happy to open a PR next to the `minimumReleaseAge` lookup if you want it.

smarlhens:
Fixes #1964.

pnpm 11 added [`registries`](https://pnpm.io/settings#registries) to `pnpm-workspace.yaml` ([pnpm/pnpm#11189](https://github.com/pnpm/pnpm/pull/11189)). ncu already read `minimumReleaseAge` from that file but ignored `registries`, so a project whose registry lives only there was silently checked against the public registry.

`registries` is read in `npmConfigFromPnpmWorkspace` and converted to npm config variables — `default` to `registry`, scoped entries to `@scope:registry` — the same way `npmConfigFromYarn` converts `npmScopes` from `.yarnrc.yml`. `getEngines` and `packageAuthorChanged` are wired through it too, as yarn already does, so they hit the same registry as the version lookup.

Two smaller things came out of it:

- Retries dropped the workspace config: `fetchUpgradedPackument` did not forward `npmConfigWorkspaceProject` when retrying, so a failed request to a private registry silently retried against the public one. It affected the workspace `.npmrc` before this PR too.
- The registries reader and the `minimumReleaseAge` reader had the same read/parse/layer logic, so they now share a `readPnpmConfig` helper.

`registries` requires pnpm >= 11, so the pnpm <= 10 `rc` file is not consulted; pnpm's global `config.yaml` is, since it supports the setting from 11.11.

Before, with `registries.default` and `registries['@types']` both pointing at `https://bogus-registry.invalid/`:

```console
Using minimumReleaseAge from pnpm-workspace.yaml: 3 days
Checking /tmp/repro/package.json

 husky  9.1.0  →  9.1.7

Minor   Backwards-compatible features
 @types/semver  7.5.0  →  7.7.1
```

After:

```console
Using minimumReleaseAge from pnpm-workspace.yaml: 3 days
Checking /tmp/repro/package.json

 husky          FetchError: request to https://bogus-registry.invalid/husky failed, reason: getaddrinfo ENOTFOUND bogus-registry.invalid
 @types/semver  FetchError: request to https://bogus-registry.invalid/@types%2Fsemver failed, reason: getaddrinfo ENOTFOUND bogus-registry.invalid
```

Tests were written first and were red before the implementation. `test/package-managers/pnpm.test.ts` covers parsing and the workspace/global layering; `test/registryPnpm.test.ts` covers the behaviour through `pnpm.latest` against `.invalid` hosts, so the host in the fetch error identifies the registry that was used: default, scoped, global `config.yaml`, precedence over `.npmrc`, precedence of an explicit `--registry`, and the retry case. `npx vitest run` is green (798 passed, 2 skipped), as are `tsc --noEmit`, `eslint` and `prettier --check`.

## Merge order

Reverted per review — `mergeNpmConfigs` is untouched, so the npmrc layers keep precedence over `npmConfigWorkspaceProject` and therefore over `registries` in `pnpm-workspace.yaml`. pnpm resolves the opposite way ([pnpm/pnpm#11754](https://github.com/pnpm/pnpm/pull/11754)); the test `is overridden by the registry in .npmrc` documents ncu's order so the divergence is pinned rather than implicit. Happy to follow up with a pnpm-scoped version in a separate PR if you want it.

### DIFF
--- a/src/package-managers/pnpm.ts
+++ b/src/package-managers/pnpm.ts
@@ -32,7 +32,7 @@ type PnpmList = {
 }[]
 
 /** Shape of the pnpm-workspace.yaml minimumReleaseAge settings. */
-export interface PnpmWorkspaceMinimumReleaseAge {
+interface PnpmWorkspaceMinimumReleaseAge {
   /** Minimum release age in minutes (pnpm's native unit). */
   minimumReleaseAge: number
   /** List of package name glob patterns excluded from the minimum release age constraint. */
@@ -43,6 +43,14 @@ export interface PnpmWorkspaceMinimumReleaseAge {
 interface MinimumReleaseAgeLayer {
   minimumReleaseAge?: number
   minimumReleaseAgeExclude: string[]
+}
+
+/** The registries resolved from pnpm's `registries` and `registry` settings. */
+interface PnpmWorkspaceRegistries {
+  /** Registry used for packages that do not match a scoped entry. */
+  default?: string
+  /** Registries keyed by package scope, e.g. `{ '@myorg': 'https://registry.example.com/' }`. */
+  scoped: Index<string>
 }
 
 /** Coerces an arbitrary config value into a non-negative minimumReleaseAge number (in minutes), or undefined if invalid. */
@@ -199,14 +207,6 @@ const getPnpmWorkspaceMinimumReleaseAge = async (
   return { minimumReleaseAge, minimumReleaseAgeExclude }
 }
 
-/** The registries resolved from pnpm's `registries` and `registry` settings. */
-export interface PnpmWorkspaceRegistries {
-  /** Registry used for packages that do not match a scoped entry. */
-  default?: string
-  /** Registries keyed by package scope, e.g. `{ '@myorg': 'https://registry.example.com/' }`. */
-  scoped: Index<string>
-}
-
 /** Matches an environment variable placeholder that interpolate did not resolve, e.g. `${MY_REGISTRY}`. */
 const envPlaceholder = /\$\{[^}]*\}/
 
@@ -361,14 +361,6 @@ const withNpmWorkspaceConfig =
     return getVersion(packageName, currentVersion, options, registries, workspaceNpmrc)
   }
 
-export const distTag = withNpmWorkspaceConfig(npm.distTag)
-export const greatest = withNpmWorkspaceConfig(npm.greatest)
-export const latest = withNpmWorkspaceConfig(npm.latest)
-export const minor = withNpmWorkspaceConfig(npm.minor)
-export const newest = withNpmWorkspaceConfig(npm.newest)
-export const patch = withNpmWorkspaceConfig(npm.patch)
-export const semver = withNpmWorkspaceConfig(npm.semver)
-
 /** Builds the pnpm argv from the given args and npm options. */
 const buildArgs = (args: string | string[], npmOptions: NpmOptions): string[] => [
   ...(npmOptions.global ? ['global'] : []),
@@ -393,8 +385,6 @@ async function spawnPnpm(
   return spawnCommand('pnpm', buildArgs(args, npmOptions), spawnPleaseOptions, spawnOptions)
 }
 
-export { defaultPrefix, getPeerDependencies } from './npm.ts'
-
 /**
  * Wraps an npm function whose last three parameters are options plus the two npm config layers, supplying them
  * from pnpm's config. This splits the pnpm config across the same two layers withNpmWorkspaceConfig uses, so
@@ -414,6 +404,14 @@ const withPnpmConfig =
     return fn(...leading, options, registries, workspaceNpmrc)
   }
 
+export { defaultPrefix, getPeerDependencies } from './npm.ts'
+export const distTag = withNpmWorkspaceConfig(npm.distTag)
+export const greatest = withNpmWorkspaceConfig(npm.greatest)
+export const latest = withNpmWorkspaceConfig(npm.latest)
+export const minor = withNpmWorkspaceConfig(npm.minor)
+export const newest = withNpmWorkspaceConfig(npm.newest)
+export const patch = withNpmWorkspaceConfig(npm.patch)
+export const semver = withNpmWorkspaceConfig(npm.semver)
 export const getDistTags = withPnpmConfig(npm.getDistTags)
 export const getEngines = withPnpmConfig(npm.getEngines)
 export const packageAuthorChanged = withPnpmConfig(npm.packageAuthorChanged)
--- a/src/package-managers/pnpm.ts
+++ b/src/package-managers/pnpm.ts
@@ -18,7 +18,6 @@ import { type SpawnOptions } from '../types/SpawnOptions.ts'
 import { type SpawnPleaseOptions } from '../types/SpawnPleaseOptions.ts'
 import { type SpawnResult } from '../types/SpawnResult.ts'
 import { type Version } from '../types/Version.ts'
-import { type VersionSpec } from '../types/VersionSpec.ts'
 import * as npm from './npm.ts'
 
 // return type of pnpm ls --json
@@ -158,15 +157,15 @@ const getPnpmMajorVersion = async (): Promise<number | null> => {
  * minimumReleaseAgeExclude patterns are merged across all considered layers. Returns null if no
  * layer defines a minimumReleaseAge.
  *
- * @param pnpmMajorVersion Optional override, used by tests to avoid spawning pnpm.
+ * @param options.pnpmMajorVersion Optional override, used by tests to avoid spawning pnpm.
  * undefined resolves the major version from the installed pnpm. null reads both globals.
  * A number selects config.yaml for >= 11 and rc for <= 10.
- * @param cwd Directory to search upwards from for pnpm-workspace.yaml. Defaults to process.cwd().
+ * @param options.cwd Directory to search upwards from for pnpm-workspace.yaml. Defaults to process.cwd().
  */
 const getPnpmWorkspaceMinimumReleaseAge = async (
-  pnpmMajorVersion?: number | null,
-  cwd?: string,
+  options: { pnpmMajorVersion?: number | null; cwd?: string } = {},
 ): Promise<PnpmWorkspaceMinimumReleaseAge | null> => {
+  const { pnpmMajorVersion, cwd } = options
   const globalConfigDir = getPnpmGlobalConfigDir()
   const pnpmWorkspacePath = await findUp('pnpm-workspace.yaml', { cwd })
   const globalConfigYamlPath = path.join(globalConfigDir, 'config.yaml')
@@ -259,9 +258,12 @@ const resolvePnpmRegistries = async (pnpmWorkspacePath?: string): Promise<PnpmWo
   return { default: defaultRegistry, scoped }
 }
 
-/** Reads registries settings from pnpm's config, searching upwards from cwd for pnpm-workspace.yaml. */
-const getPnpmWorkspaceRegistries = async (cwd?: string): Promise<PnpmWorkspaceRegistries> =>
-  resolvePnpmRegistries(await findUp('pnpm-workspace.yaml', { cwd }))
+/** Reads registries settings from pnpm's config, searching upwards from cwd for pnpm-workspace.yaml.
+ *
+ * @param options.cwd Directory to search upwards from for pnpm-workspace.yaml. Defaults to process.cwd().
+ */
+const getPnpmWorkspaceRegistries = async (options: { cwd?: string } = {}): Promise<PnpmWorkspaceRegistries> =>
+  resolvePnpmRegistries(await findUp('pnpm-workspace.yaml', { cwd: options.cwd }))
 
 /** Parses the output of `pnpm ls -g --json` into a { name: version } index. */
 const parseList = (stdout: string, command: string, stderr?: string): Index<string | undefined> => {
@@ -291,15 +293,24 @@ export const list = async (options: Options = {}): Promise<Index<string | undefi
   return parseList(stdout, command, stderr)
 }
 
-/** Reads the npmrc that sits next to pnpm-workspace.yaml, or an empty config if it does not exist. */
-const npmConfigFromWorkspaceNpmrc = async (options: Options, pnpmWorkspaceDir: string): Promise<NpmConfig> => {
+/**
+ * Reads and parses the npmrc that sits next to pnpm-workspace.yaml, or null if it does not exist.
+ * Memoized on the directory, like readPnpmConfig, so a workspace with many packages reads and parses it once.
+ */
+const readWorkspaceNpmrc = memoize(async (pnpmWorkspaceDir: string): Promise<NpmConfig | null> => {
   const pnpmWorkspaceConfigPath = path.join(pnpmWorkspaceDir, '.npmrc')
   const contents = await fs.readFile(pnpmWorkspaceConfigPath, 'utf-8').catch(() => null)
-  if (contents == null) return {}
+  return contents == null ? null : npm.normalizeNpmConfig(ini.parse(contents), pnpmWorkspaceDir)
+})
 
-  print(options, `\nUsing pnpm workspace config at ${pnpmWorkspaceConfigPath}:`, 'verbose')
+/** Reads the npmrc that sits next to pnpm-workspace.yaml, or an empty config if it does not exist. */
+const npmConfigFromWorkspaceNpmrc = async (options: Options, pnpmWorkspaceDir: string): Promise<NpmConfig> => {
+  const config = await readWorkspaceNpmrc(pnpmWorkspaceDir)
+  if (config == null) return {}
 
-  return npm.normalizeNpmConfig(ini.parse(contents), pnpmWorkspaceDir)
+  print(options, `\nUsing pnpm workspace config at ${path.join(pnpmWorkspaceDir, '.npmrc')}:`, 'verbose')
+
+  return config
 }
 
 /** pnpm's config split by the npm config layer each part belongs to, since they rank differently. */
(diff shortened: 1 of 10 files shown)


src/cli-options.ts
```diff
@@ -929,7 +929,8 @@ const cliOptions: CLIOption[] = [
     long: 'registry',
     short: 'r',
     arg: 'uri',
-    description: 'Specify the registry to use when looking up package versions.',
+    description:
+      'Specify the registry to use when looking up package versions. Overrides the registries read from the package manager config, such as `registries` in `pnpm-workspace.yaml`.',
```
raineorshine: I would rather than extend the description here. It's common knowledge that CLI options take precedence.

raineorshine (changes requested):
Thank you!

Let's leave the merge order alone. I would want that in a separate PR and scoped to pnpm, if at all.

Otherwise looks good.

smarlhens:
Both points addressed in c728c899.

**Merge order** — reverted. `mergeNpmConfigs` is back to its original order, and the diff no longer touches it beyond forwarding `npmConfigWorkspaceProject` through the retry. Consequence worth naming: `.npmrc` now wins over `registries` in `pnpm-workspace.yaml`, which is the opposite of pnpm's own resolution, so I flipped the precedence test to `is overridden by the registry in .npmrc` to pin the divergence rather than leave it implicit.

**`--registry` description** — removed, agreed. Reverted it in `cli-options.ts` and re-ran `npm run build:options`, so `README.md`, `RunOptions.ts` and `RunOptions.json` are back to their original text and are no longer part of the diff.

That leaves 4 files changed: `src/package-managers/pnpm.ts`, the retry fix in `src/package-managers/npm.ts`, and the two test files.

raineorshine (approved):
Thank you!

Please resolve conflicts, then ready to merge.

XhmikosR:
These AI-generated fixes are half correct most of the time... @raineorshine did you use Claude Opus for review?

smarlhens:
Hey @XhmikosR, hope you’re well!

Yes, I used Claude Opus 5. I initially thought it would save some time, but I can see the code still needs work.

If you’re willing to, I’d really appreciate a review and any guidance on how we could improve the PR.

XhmikosR:
Claude Opus review:

## Review: PR #1973 - feat(pnpm): read registries from pnpm-workspace.yaml

**Verified locally on the PR branch:** `tsc --noEmit` clean, prettier clean, eslint clean, full unit suite green (876 passed, 3 skipped, 70 files).

The feature shape is right. I checked it against pnpm's docs: `registries` is a `Record<string,string>` with a `default` key plus `@scope` keys, added in pnpm 11.0, and readable from the global `config.yaml` since 11.11.0. So the comment about skipping the pnpm ≤10 `rc` file is accurate, and the global-config fallback is justified.

The npm.ts retry fix is a real pre-existing bug (the workspace config was dropped on every retry, so retries silently fell back to the wrong registry). I confirmed the new `keeps using the registry when a fetch is retried` test actually guards it - reverting the npm.ts hunk makes exactly that test fail.

### 1. Registry precedence is now inconsistent between code paths

`src/package-managers/pnpm.ts` passes the pnpm config two different ways:

- `withNpmWorkspaceConfig` → 5th arg → `npmConfigWorkspaceProject`, which `mergeNpmConfigs` spreads **first** (lowest precedence, below `npmConfigUser`).
- the new `getEngines` / `packageAuthorChanged` → 4th arg → `npmConfigLocal`, which is spread **after** `npmConfigUser` (higher precedence).

Result: the same project resolves versions and engines from different registries. Reproduced with `npm_config_registry=https://env-registry.invalid/` plus a workspace `registries.default: https://pnpm-registry.invalid/`:

```
pnpm.latest      -> request to https://env-registry.invalid/ncu-test-v2
pnpm.getEngines  -> request to https://pnpm-registry.invalid/ncu-test-v2/1.0.0
```

`--enginesNode` and `--format ownerChanged` will hit a different registry than the version lookup. yarn avoids this by passing its config as `npmConfigLocal` in *every* path including `withNpmConfigFromYarn` (yarn.ts:345-348).

Two ways out, and this is the maintainer's call:
- minimal: keep the pnpm config at `npmConfigWorkspaceProject` precedence and drop the new `getEngines`/`packageAuthorChanged` overrides (but then those paths ignore pnpm registries entirely);
- better: switch `withNpmWorkspaceConfig` to pass it as `npmConfigLocal` like yarn. That makes all paths agree *and* moves pnpm-workspace.yaml above the ambient/user `.npmrc`, which is closer to pnpm's actual precedence. It does change existing behavior for the workspace `.npmrc`, so it needs a deliberate decision.

Related: the test `is overridden by the registry in .npmrc` bakes in the divergence from pnpm, with a comment admitting it. That's fine as a documented choice, but it's the same underlying question - worth resolving deliberately rather than by accident.

### 2. The new `getPnpmWorkspaceRegistries` tests aren't isolated from the machine's pnpm config

The new outer `beforeEach` in `test/package-managers/pnpm.test.ts` creates a temp dir and chdirs, but never sets `XDG_CONFIG_HOME` - unlike the existing `getPnpmWorkspaceMinimumReleaseAge` describe, which does exactly that "to isolate the global config layers from the machine running the tests".

Since `getPnpmWorkspaceRegistries` now reads the global `config.yaml`, those tests read the developer's real pnpm config. I confirmed it: with a `$XDG_CONFIG_HOME/pnpm/config.yaml` containing `registries.default`, 3 of the 5 new tests fail:

```
× returns no registries when pnpm-workspace.yaml does not define any
× ignores non-string and empty registry values
× ignores a registries value that is not a map
- "default": undefined
+ "default": "https://machine-global.example.com/"
```

Green on CI, red for any pnpm 11 user with a global registry configured. The describe needs its own isolated `XDG_CONFIG_HOME`.

### 3. `getDistTags` was left out

pnpm still re-exports npm's `getDistTags` unwrapped, while yarn overrides it (yarn.ts:415). It's used by `src/lib/resolveDistTagsInPeerDependencies.ts:56`, so dist-tagged peer deps still resolve against the public registry for pnpm users with a private one. If `getEngines` and `packageAuthorChanged` are worth wrapping, this is too.

### 4. The new outer hooks apply to every test in the file

`tempDir`, `originalCwd`, and `writeWorkspace` were added at the `describe('pnpm')` level, where they shadow the identically-named ones inside `getPnpmWorkspaceMinimumReleaseAge`. Every test in the file - including the pure `parseList` ones - now does an `mkdtemp` + `rm -rf`. They should live in the `getPnpmWorkspaceRegistries` describe (which also fixes #2 naturally).

### 5. `pnpm-workspace.yaml` is now read and parsed twice per run

`npmConfigFromPnpmWorkspace` does a `findUp`, then `getPnpmWorkspaceRegistries` does another, and `getPnpmWorkspaceMinimumReleaseAge` does a third - each re-reading and re-parsing the same YAML. `readPnpmConfig` was introduced as the shared helper, so memoizing it on `(filePath, format)` is the natural fix; there's precedent in `parseNpmrc` at npm.ts:466.

### 6. `options.cwd` is honored for registries but not minimumReleaseAge

`getPnpmWorkspaceRegistries` takes a `cwd` and threads it into `findUp`; `getPnpmWorkspaceMinimumReleaseAge` still calls bare `findUp('pnpm-workspace.yaml')`. Pre-existing, but the PR makes `--cwd` work for one setting and not its neighbor. Worth fixing in the same pass.

### 7. Verbose output regression

The old function returned `{}` early when there was no `pnpm-workspace.yaml`. It now always falls through to `print(options, config, 'verbose')`, so any pnpm project without a workspace file prints a bare `{}` line under `--verbose`. Guard the print on a non-empty config.

### Smaller stuff

- `test/registryPnpm.test.ts` duplicates the tempdir + XDG setup of the new nested describe almost verbatim. It also sits at the test root while every other pnpm test is in `test/package-managers/`.
- `catch (e) {}` → `catch {}` (the codebase uses that form, e.g. yarn.ts:403).
- `PnpmWorkspaceMinimumReleaseAge` is exported but `PnpmWorkspaceRegistries` isn't, though both back a `pnpmApi` function.

### Verdict

Solid, well-tested contribution that solves the reported problem, and the retry fix is a genuine catch. #1 and #2 should block: the first ships two different registries for the same project, the second is a test that only passes on machines without the very config the feature targets. The rest is cleanup.

raineorshine:
> These AI-generated fixes are half correct most of the time... @raineorshine did you use Claude Opus for review?

No, manual. Sorry if I missed anything.

smarlhens:
Hey @XhmikosR @raineorshine, thanks for the review, really useful.
I followed all the findings. Hope it's better now, thanks again for your time 🙏🏻

XhmikosR:
@smarlhens: please do not force push the branch any more.

Claude Opus review:

## Re-review after the force-push

Everything from the last round is fixed:

- all four paths now pass the pnpm config as `npmConfigWorkspaceProject`, so versions, dist-tags, engines and authors resolve from the same registry, and there's a test asserting it
- `getDistTags` is wrapped now
- `readPnpmConfig` is memoized, so the yaml is read and parsed once
- `usePnpmConfigDirs` isolates cwd and `XDG_CONFIG_HOME` per test. I re-ran the pnpm tests with a global `config.yaml` containing `registries.default` and `minimumReleaseAge`, plus `npm_config_registry` set. All 39 pass. That combination broke 3 tests in the previous revision.
- `--cwd` is honored for minimumReleaseAge too
- no more bare `{}` under `--verbose`, `catch {}`, test moved next to the other pnpm tests

Also checked the shape against pnpm's docs: `registries` is `Record<string,string>` with `default` plus `@scope` keys, added v11.0.0, readable from the global `config.yaml` since v11.11.0. Matches what the code does, and the comment about skipping the pnpm <= 10 `rc` file is right.

Not interpolating env vars in the URLs is also correct for pnpm-workspace.yaml. pnpm stopped expanding them there in v11.5.3 on purpose, since it's an untrusted location. Worth noting yarn's `npmRegistryKeyValue` does interpolate, so the difference is deliberate, not an oversight.

Remaining:

### `registry` (singular) in pnpm-workspace.yaml is still ignored

pnpm 11 accepts the plain `.npmrc` settings in pnpm-workspace.yaml, and the v11.5.3 release note names `registry` explicitly as one of the "settings of pnpm-workspace.yaml that define registry URLs". So this:

```yaml
registry: https://private.example.com/
```

still gets checked against the public registry. Same bug as #1964, just a different key. `parseRegistries` only looks at `parsed.registries`. Should be a one-liner falling back to `parsed.registry` for the default, plus a test.

### Precedence is still below the ambient npm config

Now that it's consistent everywhere and there's a comment and a test spelling it out, this is your call rather than a bug. But the practical effect is that anyone who has ever run `npm config set registry` has a `registry=` in `~/.npmrc` that silently beats `registries.default` in pnpm-workspace.yaml, which is the exact failure the PR is fixing. pnpm itself ranks pnpm-workspace.yaml above `.npmrc`.

Moving `withNpmWorkspaceConfig` to the `npmConfigLocal` layer like yarn does would fix it and keep all paths in agreement, at the cost of changing how the workspace-adjacent `.npmrc` ranks today. Just flagging that the current choice leaves a silent-wrong-registry case in place.

### Env vars in the global config.yaml

pnpm does expand env vars in registry URLs in the global `config.yaml` (that's the "trusted location" the v11.5.3 note points you at). ncu reads that file now but passes the value through literally, so `default: ${MY_REGISTRY}` there won't work. Minor, but it's the one place interpolation is expected.

### `--peer` still bypasses it

`getPeerDependencies` is re-exported from npm unwrapped and shells out to `npm view`, so peer lookups on a pnpm project with a private registry keep hitting the public one. Pre-existing and a different mechanism from the rest of this PR, so fine to leave, but it's the last hole.

### Small stuff

- the doc comment on `withNpmWorkspaceConfig` still says only "passes the npmrc located next to the pnpm-workspace.yaml"
- when there's no `.npmrc` next to pnpm-workspace.yaml, the verbose output prints the merged config with no header line above it. yarn prints "Merged yarn config in npm format:" for the same case
- README has a table for which cooldown setting each package manager reads. Registries deserve a line there too, especially the `.npmrc`-wins bit
- the tests assert on `.invalid` hostnames appearing in fetch error messages. First use of that pattern in this repo. Fast and offline-safe on a normal resolver, but a resolver that hijacks NXDOMAIN would change the error. Low priority

---

Please make sure sure everything is fixed before pushing.

smarlhens:
@XhmikosR 👋🏻 

Pushed, all three addressed.

Precedence: left `mergeNpmConfigs` alone per https://github.com/raineorshine/npm-check-updates/pull/1973#pullrequestreview-4859858730, moved only pnpm's `registries` to `npmConfigLocal`, the workspace `.npmrc` keeps its layer. Project/cwd `.npmrc` still wins, that's the separate PR.

I added a new commit with the changes. The forced update was the commit message only, `git diff 004c688a bdf5fe7e` is empty.

src/package-managers/pnpm.ts
```diff
@@ -175,12 +161,14 @@ const getPnpmMajorVersion = async (): Promise<number | null> => {
  * @param pnpmMajorVersion Optional override, used by tests to avoid spawning pnpm.
  * undefined resolves the major version from the installed pnpm. null reads both globals.
  * A number selects config.yaml for >= 11 and rc for <= 10.
+ * @param cwd Directory to search upwards from for pnpm-workspace.yaml. Defaults to process.cwd().
  */
 const getPnpmWorkspaceMinimumReleaseAge = async (
   pnpmMajorVersion?: number | null,
+  cwd?: string,
```
raineorshine: Using multiple positional arguments makes for an awkward situation when trying to pass only the second (i.e. the explicit `undefined`). A better design would be an optional options object.

src/package-managers/pnpm.ts
```diff
@@ -277,11 +381,60 @@ async function spawnPnpm(
...
+// The wrappers below split the pnpm config across the same two layers withNpmWorkspaceConfig uses, so that every
+// code path resolves the same registry. Otherwise a project resolves versions and engines from two different ones.
+
+/**
+ * Fetches all dist-tags published for a package.
+ *
+ * @param packageName
+ * @returns Promised {tag: version} collection
+ */
+export const getDistTags = async (packageName: string, options: Options = {}): Promise<Index<Version>> => {
```
raineorshine: Try to follow the pattern used by `distTag` and `greatest`. Notice how cleanly and functionally they use higher-order functions to transform the npm functions.

```
export const distTag = withNpmWorkspaceConfig(npm.distTag)
export const greatest = withNpmWorkspaceConfig(npm.greatest)
```

Your new versions of `getDistTags`, `getEngines`, and `packageAuthorChanged` repeat a lot of boilerplate. Instead make a higher order function that abstracts away the mapping and allows point-free style at the call site.

e.g.

```
/** Wraps an npm function whose last two parameters are the npm config layers, supplying them from pnpm's config. */
const withPnpmConfig =
  <A extends unknown[], R>(fn: (...args: [...A, Options, NpmConfig, NpmConfig]) => Promise<R>) =>
  async (...args: [...A, Options?]): Promise<R> => {
    const options = (args.at(-1) ?? {}) as Options
    const leading = args.slice(0, -1) as A
    const { registries, workspaceNpmrc } = await npmConfigFromPnpmWorkspace(options)
    return fn(...leading, options, registries, workspaceNpmrc)
  }

export const distTag = withPnpmConfig(npm.distTag)
export const latest = withPnpmConfig(npm.latest)
export const getDistTags = withPnpmConfig(npm.getDistTags)
export const getEngines = withPnpmConfig(npm.getEngines)
export const packageAuthorChanged = withPnpmConfig(npm.packageAuthorChanged)
```

raineorshine (changes requested):
Thanks for the update and sorry for the long delay. I'm on this project only on weekends mainly.

raineorshine (changes requested):
Thanks, looks good!

I can't push directly to your fork, so please group the types at the top and the exports at the bottom for better organization:

```diff
diff --git a/src/package-managers/pnpm.ts b/src/package-managers/pnpm.ts
index 9a145177..c78a32e9 100644
--- a/src/package-managers/pnpm.ts
+++ b/src/package-managers/pnpm.ts
@@ -32,7 +32,7 @@ type PnpmList = {
 }[]
 
 /** Shape of the pnpm-workspace.yaml minimumReleaseAge settings. */
-export interface PnpmWorkspaceMinimumReleaseAge {
+interface PnpmWorkspaceMinimumReleaseAge {
   /** Minimum release age in minutes (pnpm's native unit). */
   minimumReleaseAge: number
   /** List of package name glob patterns excluded from the minimum release age constraint. */
@@ -45,6 +45,14 @@ interface MinimumReleaseAgeLayer {
   minimumReleaseAgeExclude: string[]
 }
 
+/** The registries resolved from pnpm's `registries` and `registry` settings. */
+interface PnpmWorkspaceRegistries {
+  /** Registry used for packages that do not match a scoped entry. */
+  default?: string
+  /** Registries keyed by package scope, e.g. `{ '@myorg': 'https://registry.example.com/' }`. */
+  scoped: Index<string>
+}
+
 /** Coerces an arbitrary config value into a non-negative minimumReleaseAge number (in minutes), or undefined if invalid. */
 const coerceMinimumReleaseAge = (raw: unknown): number | undefined => {
   const value = typeof raw === 'number' ? raw : typeof raw === 'string' && raw.trim() !== '' ? Number(raw) : NaN
@@ -199,14 +207,6 @@ const getPnpmWorkspaceMinimumReleaseAge = async (
   return { minimumReleaseAge, minimumReleaseAgeExclude }
 }
 
-/** The registries resolved from pnpm's `registries` and `registry` settings. */
-export interface PnpmWorkspaceRegistries {
-  /** Registry used for packages that do not match a scoped entry. */
-  default?: string
-  /** Registries keyed by package scope, e.g. `{ '@myorg': 'https://registry.example.com/' }`. */
-  scoped: Index<string>
-}
-
 /** Matches an environment variable placeholder that interpolate did not resolve, e.g. `${MY_REGISTRY}`. */
 const envPlaceholder = /\$\{[^}]*\}/
 
@@ -361,14 +361,6 @@ const withNpmWorkspaceConfig =
     return getVersion(packageName, currentVersion, options, registries, workspaceNpmrc)
   }
 
-export const distTag = withNpmWorkspaceConfig(npm.distTag)
-export const greatest = withNpmWorkspaceConfig(npm.greatest)
-export const latest = withNpmWorkspaceConfig(npm.latest)
-export const minor = withNpmWorkspaceConfig(npm.minor)
-export const newest = withNpmWorkspaceConfig(npm.newest)
-export const patch = withNpmWorkspaceConfig(npm.patch)
-export const semver = withNpmWorkspaceConfig(npm.semver)
-
 /** Builds the pnpm argv from the given args and npm options. */
 const buildArgs = (args: string | string[], npmOptions: NpmOptions): string[] => [
   ...(npmOptions.global ? ['global'] : []),
@@ -393,8 +385,6 @@ async function spawnPnpm(
   return spawnCommand('pnpm', buildArgs(args, npmOptions), spawnPleaseOptions, spawnOptions)
 }
 
-export { defaultPrefix, getPeerDependencies } from './npm.ts'
-
 /**
  * Wraps an npm function whose last three parameters are options plus the two npm config layers, supplying them
  * from pnpm's config. This splits the pnpm config across the same two layers withNpmWorkspaceConfig uses, so
@@ -414,6 +404,14 @@ const withPnpmConfig =
     return fn(...leading, options, registries, workspaceNpmrc)
   }
 
+export { defaultPrefix, getPeerDependencies } from './npm.ts'
+export const distTag = withNpmWorkspaceConfig(npm.distTag)
+export const greatest = withNpmWorkspaceConfig(npm.greatest)
+export const latest = withNpmWorkspaceConfig(npm.latest)
+export const minor = withNpmWorkspaceConfig(npm.minor)
+export const newest = withNpmWorkspaceConfig(npm.newest)
+export const patch = withNpmWorkspaceConfig(npm.patch)
+export const semver = withNpmWorkspaceConfig(npm.semver)
 export const getDistTags = withPnpmConfig(npm.getDistTags)
 export const getEngines = withPnpmConfig(npm.getEngines)
 export const packageAuthorChanged = withPnpmConfig(npm.packageAuthorChanged)
```

raineorshine (approved):
Thank you!

PR #1962, opened 2026-07-29
CI cleanup and tweaks

XhmikosR:
E2E failing on macOS will be fixed by #1961. Unsure if we want E2E tests to run on Windows too, but macOS is worth it IMHO. Let me know what you think.

TODO:

- rebase after #1961 is merged and make sure everything is green here before merging
- adapt branch protection rules

PR #1977, opened 2026-08-05
e2e.sh: drop  the npm view retry

XhmikosR:
It doesn't look like it's really needed

Follow up to #1961.

Let's keep this as draft for a while to test things out, especially after #1962.


XhmikosR:
@raineorshine: CI seems to pass all times I rebased this branch, so it looks like this is indeed unneeded.

PR #1978, opened 2026-08-05
lint: remove FORCE_COLOR=1

XhmikosR:
Unsure why this was there in the first place. I see it was introduced in https://github.com/raineorshine/npm-check-updates/commit/3833db54beabf2d064e10b2a745b0b26db756785 but not sure what it fixes?

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "build": "vite build",
     "build:options": "node scripts/build-options.ts",
     "build:analyze": "cross-env ANALYZER=true vite build",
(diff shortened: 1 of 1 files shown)


raineorshine:
`FORCE_COLOR=1` is needed to keep color in the terminal when the linters run via npm-run-all. With `--parallel --aggregate-output`, each child's stdout is a pipe rather than a TTY, so tools like ESLint and tsc disable color on their own. 

With `FORCE_COLOR=1`:

image


Without `FORCE_COLOR=1`:

image

XhmikosR:
Alright, I guess it needs to stay...

raineorshine:
100% needs to stay

PR #1980, opened 2026-08-06
Deprecate CHANGELOG.md

### LINKED ISSUE #1167: Please update CHANGELOG or write release notes.
- [x] I am using `node >= 14`

---

I am developing a package called [npm-update-package](https://github.com/npm-update-package/npm-update-package) that relies heavily on [npm-check-updates](https://github.com/raineorshine/npm-check-updates).
It would be helpful if you could describe the changes in each version of the CHANGELOG, including minor updates and patch updates.
Alternatively, please write release notes using GitHub's Releases.

Thank you.

XhmikosR:
Closes #1167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
-This file only documents **major version** releases. For smaller releases, you're stuck reading the [commit history](https://github.com/raineorshine/npm-check-updates/commits/main).
+**This file is no longer maintained.** Release notes are published on the [GitHub releases page](https://github.com/raineorshine/npm-check-updates/releases).
+
+---
 
 ## [22.0.0] - 2026-04-25
 


raineorshine (approved):

dzek69:
Another project removes the good way of having changelogs 😢 

Changelogs in a files are actually useful. GitHub can die. Browsing changelog now requires an API/a browser.

Changelog as a repo file:
- can be moved to any git host with just a `git push`
- can be easily accessed by AI tools (something I see you're heavily using),
- can be accessed offline after you clone a repo
- can be accessed and therefore searched in as a whole (ctrl+f, grep), instead of clicking through paginated views on github UI

Locking into a propertiary solution instead does not sound like something an open-source software believer should do.

Can we revert this change? Having changelog on the release page is very useful, but changelog there can be just a mirror from the file changelog.

@raineorshine @XhmikosR

XhmikosR:
Personally, I don't like having to keep CHANGELOG.md in sync. The only source of truth is the GitHub releases now. If @raineorshine wants to change this, it's their project.

raineorshine:
@dzek69 Those are fair points. If the arguments for a CHANGELOG.md are minor UX improvements and more of an Open Source ethos, and the arguments for deprecating it are merely for convenience or tidiness, I would lean towards the former.

We can still have GitHub releases be the source of truth. Keeping CHANGELOG.md in sync is no maintenance burden if we create a new workflow to automate it.

dzek69:
Yes, automation is what I wanted to propose. AFAIK there are projects that either helps with syncing CHANGELOG.md file into Releases and the other way around. I did not really verify what is used for releases in this project.

I see the PR, thank you for keeping the CHANGELOG.md file alive!

PR #1987, opened 2026-08-07
Assorted fixes

XhmikosR:
Please check the individual commits.

Better not squash when merging. I'll keep this as draft until the other PRs are merged in case it needs a rebase.


raineorshine:
Thank you! Will review in another session.

raineorshine:
I'm looking over this PR, and unfortunately there are too many unrelated changes. I can review individual commits, as you suggest, but then as soon as I have a comment or need to discuss anything, it all ends up in the same place. It is also harder to pause and resume my review when all of the changes are in one PR.

The previous workflow we had worked well for me. Please submit one PR per functional change which will facilitate easy review and well-organized conversations. Thanks.

PR #1989, opened 2026-08-07
fix: strip control characters from homepage/repo format output

### LINKED ISSUE #1988: --format homepage/repo output isn't stripped of terminal escape sequences
## Summary

`ncu`'s `--format homepage` and `--format repo` options add a column to its dependency-upgrade table showing each dependency's homepage URL or source repository URL, read directly from that dependency's own `package.json` (`src/lib/logging.ts`). Unlike a package name or version, these fields have no character restrictions. Since running `ncu` is exactly what a developer does to check for updates, and any dependency (including transitive ones never directly chosen) can carry an arbitrary value in these fields, any escape sequence embedded there is written to the terminal and interpreted live during a routine check.

## POC

(available upon request)

## Impact

I validated this fully offline using `ncu`'s documented `--registry` static-file option: created a local dependency whose `package.json` had a `homepage` field containing a raw OSC title-set escape sequence, ran `ncu --format homepage`, and inspected the raw stdout bytes directly (a plain redirect to a file is sufficient here, since this is a non-interactive CLI). The captured output contained the exact injected sequence, unmodified, directly adjacent to the homepage URL text, confirming it reaches the terminal without any stripping. I used a benign title-set marker specifically to confirm the sequence reaches output unstripped, rather than a more aggressive payload; the specific range of what an injected OSC/CSI sequence can do depends on the victim's terminal emulator and was not enumerated further.

## Suggested Fix

Strip control characters, particularly the ESC byte, from `homepageUrl` and `repoUrl` before they're added to the output table. I'd suggest fixing this at both values narrowly, rather than at the shared table-rendering function, since other cells in the same table carry intentional ANSI color codes that need to be preserved.

carfeii:
Fixes #1988

`--format homepage` and `--format repo` read a dependency's own `package.json` `homepage`/`repository` fields with no validation and wrote them straight into the upgrade table. Unlike a package name or version, these fields have no character restrictions, so a crafted dependency could inject terminal escape sequences into ncu's own output during a routine dependency check.

## Change

Adds `sanitizeForDisplay` (`src/lib/logging.ts`), which strips control characters (code points below `0x20` or `0x7f`). Applied narrowly to `homepageUrl` and `repoUrl` where they're computed, rather than at the shared table-rendering function, since other cells in the same table (dependency type label, colorized version diff) carry intentional `chalk` ANSI color codes that must not be stripped.

## Testing

- Reproduced the issue fully offline via the documented `--registry` static-file option: a local dependency with a `homepage` field containing a raw OSC title-set escape sequence, `ncu --format homepage`, raw stdout bytes inspected directly (hex-dumped, not just visually inspected) confirmed the raw escape bytes reached output adjacent to the homepage URL text.
- Re-ran the same reproduction against this fix; the injected control bytes are gone, only harmless printable text remains.
- `tsc --noEmit` and the relevant unit tests (`table.test.ts`, `logging.test.ts`, `getRepoUrl.test.ts`, `format.test.ts`, 29 tests total) pass unchanged.
- Note on environment: the available Node here was v18.19.1, below the `engines.node` requirement, which caused a hard install failure; used a Node v26.0.0 binary for installing dependencies, building, and testing, no project files depend on this.

### DIFF
--- a/src/lib/logging.ts
+++ b/src/lib/logging.ts
@@ -96,6 +96,16 @@ export function printSorted<T extends { [key: string]: any }>(options: Options, 
 }
 
 /** Create a table with the appropriate columns and alignment to render dependency upgrades. */
+/**
+ * Strips control characters (e.g. terminal escape sequences) from text that
+ * originates from an unvalidated package.json field (homepage, repository),
+ * which has no character restrictions, before it is written to the terminal.
+ */
+function sanitizeForDisplay(s: string): string {
+  // eslint-disable-next-line no-control-regex
+  return s.replace(/[\u0000-\u001f\u007f]/g, '')
+}
+
 function renderDependencyTable(rows: string[][]) {
   const table = new Table({
     colAligns: ['left', 'right', 'right', 'right', 'left', 'left'],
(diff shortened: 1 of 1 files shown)


XhmikosR:
Why use this instead of https://nodejs.org/api/util.html#utilstripvtcontrolcharactersstr? If it's even needed, that is.

XhmikosR:
Closing as this is incomplete. Also, the linked issue is overstated.

PR #1994, opened 2026-08-08
Strip terminal escape sequences from package and registry text

### LINKED ISSUE #1988: --format homepage/repo output isn't stripped of terminal escape sequences

XhmikosR:
Fixes #1988.

### DIFF
--- a/src/lib/logging.ts
+++ b/src/lib/logging.ts
@@ -2,6 +2,7 @@
  * Logging functions.
  */
 import fs from 'node:fs/promises'
+import { stripVTControlCharacters } from 'node:util'
 import Table from 'cli-table3'
 import semver from 'semver'
 import { type CooldownFunction } from '../types/CooldownFunction.ts'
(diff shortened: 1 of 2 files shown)


raineorshine (approved):
Thank you

PR #1996, opened 2026-08-08
Add support for excluding deprecated versions

### LINKED ISSUE #1550: Do not suggest deprecated versions by default
I noticed that in version 17 the default was changed to include deprecated versions, but it doesn't seem to be mentioned in the changelog. The reason seems to be higher bandwidth, but IMO saving bandwidth at the cost of safety and correctness is not recommended.

XhmikosR:
Please check the individual patches. The second patch can be considered a breaking change, though.

**If merged this should not be squashed.**

Fixes #1550.

### DIFF
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -809,8 +809,8 @@ async function fetchUpgradedPackument(
         new Set([
           'dist-tags',
           ...fields,
-          ...(!options.deprecated ? (['deprecated', 'versions'] as const) : []),
-          ...(options.enginesNode ? (['engines', 'versions'] as const) : []),
+          // deprecated and engines are only set on entries of versions, not on the packument itself
+          ...(!options.deprecated || options.enginesNode ? (['versions'] as const) : []),
         ]),
       ),
       fullMetadata ? null : tag,
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -141,6 +141,34 @@ const fetchPartialPackument = async (
   }
 }
 
+/**
+ * Fetches the manifest published to a dist-tag in packument shape.
+ * Returns null if the tag does not exist or the registry does not serve manifests.
+ */
+const fetchTagManifest = async (
+  name: string,
+  tag: string,
+  opts: npmRegistryFetch.FetchOptions,
+): Promise<Partial<Packument> | null> => {
+  let manifest: Partial<Packument>
+  try {
+    // a manifest is a few KB, so read it whole instead of aborting the stream partway
+    manifest = await fetchPartialPackument(name, [], null, { ...opts, fullMetadata: true }, tag)
+  } catch (err: any) {
+    // a missing tag and an unsupported endpoint are both 4xx
+    if (err.statusCode >= 400 && err.statusCode < 500) return null
+    throw err
+  }
+
+  if (!manifest?.version) return null
+
+  return {
+    name,
+    'dist-tags': { [tag]: manifest.version },
+    versions: { [manifest.version]: manifest as Packument['versions'][string] },
+  }
+}
+
 interface GreatestWithFallbackResult {
   targetVersion: string | null
   fallbackVersion: string | null
@@ -763,6 +791,18 @@ async function fetchUpgradedPackument(
   let result: Partial<Packument> | undefined
   try {
     const tag = options.distTag || 'latest'
+
+    // when a single tag is requested, its manifest exposes deprecated without the whole versions map
+    if (options.distTag && !options.deprecated && !fullMetadata && fields.length === 1 && fields[0] === 'dist-tags') {
+      const manifest = await fetchTagManifest(packageName, tag, npmConfigMerged)
+      if (manifest) return manifest
+
+      // the manifest is missing both when the tag does not exist and when the registry does not
+      // serve manifests, so resolve the tag before paying for the versions map
+      const distTags = await fetchPartialPackument(packageName, ['dist-tags'], tag, npmConfigMerged)
+      if (!distTags['dist-tags']?.[tag]) return distTags
+    }
+
     result = await fetchPartialPackument(
       packageName,
       Array.from(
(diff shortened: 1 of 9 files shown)


src/package-managers/npm.ts
```diff
@@ -142,6 +142,34 @@ const fetchPartialPackument = async (
...
+ */
+const fetchTagManifest = async (
+  name: string,
+  tag: string,
+  opts: npmRegistryFetch.FetchOptions,
+): Promise<Partial<Packument> | null> => {
+  let manifest: Partial<Packument>
+  try {
+    // a manifest is a few KB, so read it whole instead of aborting the stream partway
+    manifest = await fetchPartialPackument(name, [], null, { ...opts, fullMetadata: true }, tag)
```
raineorshine, line 156: Could we find a way to do this without `fullMetadata: true`? npm-check-update's README alone is 36KB.

The streaming and short circuiting of `fetchPartialPackument` was carefully designed to only fetch the minimum data needed based on the version target and fields that are needed. I would love to preserve that if possible.

raineorshine (changes requested):

src/package-managers/npm.ts
XhmikosR: The README isn't in the version manifest, only in the packument:

| | bytes |
|---|---|
| `GET /npm-check-updates` | 2,530,383 (`readme` is 37,213 of that) |
| `GET /npm-check-updates/latest` | 4,940, no `readme` field |

npmjs also ignores the accept header on version endpoints. `application/json`, the corgi mix and bare `application/vnd.npm.install-v1+json` all give back the same bytes (webpack: 11,624 for all three).

I changed it to:

```ts
manifest = await fetchPartialPackument(name, ['version', 'deprecated', 'engines'], null, opts, tag, false)
```

Numbers through a logging proxy, 4 deps (express, eslint, typescript, ncu-test-deprecated), counting only what's read before the abort:

| | requests | bytes |
|---|---|---|
| `--deprecated` (packument, early abort) | 4 | 44,931 |
| default (tag manifest) | 5 | 20,915 |

The manifest path moves less because the abort still pulls a full 16KB chunk per package, and a manifest is 2-7KB. The extra request is the packument fallback for ncu-test-deprecated, whose latest really is deprecated.

`foundAll` never fires here. It needs every requested field to be present, and `deprecated` is missing on versions that aren't deprecated, so the manifest gets read to the end. Fine at this size, but the abort doesn't buy anything on this path.

The last arg is a new `retryFullDoc` param. The corgi retry is for registries that don't serve abbreviated packuments, but `fetchTagManifest` already falls back to the packument on 4xx, so requesting the same manifest url again is wasted. Without it, `--target @sometag` does two 404s per package that doesn't have the tag instead of one.

That being said, I'm not sure the last patch really saves us anything...

raineorshine: > The README isn't in the version manifest, only in the packument

Ah, thanks for clarifying.

> `foundAll` never fires here. It needs every requested field to be present, and `deprecated` is missing on versions that aren't deprecated, so the manifest gets read to the end. Fine at this size, but the abort doesn't buy anything on this path.

Yes, this was my concern. But if the version manifest is small, then the abort is hardly worth it.

---

Given the above, what's your opinion about the current approach? Are there any code paths where the performance suffers? Any tradeoffs we need to take into account if we allow `deprecated` to be checked and excluded by default again?

Thanks!

XhmikosR: Two corrections to my last comment after measuring properly.

The last patch doesn't buy anything, so I dropped it. I compared the branch with and without it: same requests, same bytes, every path. `retryFullDoc` was only cancelling out a retry that `fullMetadata: true` never triggered anyway, since the catch already skips the corgi retry when `opts.fullMetadata` is set. So it was an extra param for nothing. `fullMetadata: true` on a version url is really "this response is one small json doc, parse it whole", not "give me the fat packument".

Also my 44,931 number was a single run and it's not reproducible. The aborted stream isn't deterministic, how much lands before the abort takes effect varies a lot. Six runs of `--deprecated` on the same 4 deps:

| run | bytes |
|---|---|
| 1 | 83,895 |
| 2 | 342,408 |
| 3 | 342,408 |
| 4 | 723,928 |
| 5 | 878,046 |
| 6 | 973,200 |

The manifest path gives 20,915 on all six runs. So the gap is bigger than I said, not smaller, but the honest metric here is requests, not bytes.

On your question about paths that suffer, I found one and fixed it. `--target @sometag` where a package doesn't have that tag was fetching the whole versions map for nothing:

| | requests | bytes |
|---|---|---|
| `--target @nonexistent` before | 8 | 10,153,582 |
| `--target @nonexistent` after | 8 | 73,896 - 372,046 |

The manifest 404s both when the tag is missing and when a registry doesn't serve manifests, and the fallback couldn't tell them apart so it always asked for versions. Now it resolves dist-tags first and returns early when the tag isn't there. Squashed into the first commit.

The rest, measured 3 runs each, all stable:

| | requests | bytes |
|---|---|---|
| latest, deprecated excluded | 5 | 20,915 |
| `-t minor`, deprecated excluded | 4 | 10,153,454 |
| `-t minor --deprecated` | 4 | 10,153,454 |
| `--cooldown 7` excluded vs `--deprecated` | 6 vs 5 | ~20.4M both |

minor/patch/greatest/newest/semver are identical either way, they already pull versions so the check is free. cooldown and `--format time` skip the manifest entirely since they need `time`.

The one tradeoff left for the default flip: a registry that doesn't serve `/<pkg>/<tag>` falls back to the packument for every package, so the default gets as slow as `--no-deprecated` is today. I faked one with a proxy that 404s every version endpoint, 13 requests and ~10.5MB vs 4 requests when deprecated is allowed. Nothing to do about it, you can't check deprecation without versions there, but worth knowing before flipping.

My opinion: the approach is fine for npm. It leans on `/<pkg>/<tag>` being cheap and carrying `deprecated`, which holds on npmjs and degrades to current behavior elsewhere. The 5th request in the default case is the packument fallback for the one package whose latest really is deprecated, which is the case we actually want to catch.

raineorshine: Thanks for the additional work.

> The one tradeoff left for the default flip: a registry that doesn't serve `/<pkg>/<tag>` falls back to the packument for every package, so the default gets as slow as `--no-deprecated` is today. I faked one with a proxy that 404s every version endpoint, 13 requests and ~10.5MB vs 4 requests when deprecated is allowed. Nothing to do about it, you can't check deprecation without versions there, but worth knowing before flipping.

I'm fine optimizing for registry.npmjs.org and similar.

raineorshine (approved):
Breaking change, but ready any time.

XhmikosR:
I'm not sure the next version should be again a major bump just for this, though. Alternatively, I could drop 9231fafc2d95279324093acae956b96ee0a13166 for now and make it a minor version bump.

raineorshine:
Yes, it would be good to get the other changes in so that it's an easier merge when we do decide to make it the default.

XhmikosR:
I think it's fair to make this the default in a minor version. Deprecated versions shouldn't have been showing up by default. But I dropped the patch for now. Feel free to enable it later by default.

raineorshine:
npm-check-updates follows semver strictly, since it's a core infra tool that many teams use in automated workflows. Any breaking change that is not a bug fix will need to be released as a major version.

XhmikosR:
Then I suggest that we enable this by default and not make it opt-in.

PR #1997, opened 2026-08-08
Do not let rejectOnError:false swallow spawn errors

XhmikosR:


raineorshine (approved):
Thanks. If this is a weakness in `spawn-please`, we could also a change there to improve the API.

XhmikosR:
Yeah, it's mostly the API.

rejectOnError: false skips the error listener, so a spawn that fails (ENOENT, command not installed) emits an unhandled error event and takes the process down instead of rejecting. Those are two different things, "the command exited non-zero" and "the command never ran", and the flag shouldn't control the second one.

The other half is the reject value. On a non-zero exit it rejects with stderr as a bare string, so there's no err.message, no stack, no exit code, and the stdout collected so far is dropped. That one hurts for npm ls / pnpm ls, which exit non-zero on unmet peers but still print usable JSON on stdout.

So for v4 I'd do:

* always attach the error listener, reject with the real Error no matter what rejectOnError is
* reject non-zero exits with an Error carrying stdout, stderr and code
* add code to the resolved object
* stop mutating the caller's options object

Breaking, since the reject value goes from string to Error. The problem is that spawn-please has outdated deps and CI, if you could tackle that happy to provide a PR.

Unrelated, the README is out of date: it says it rejects on non-empty stderr, but the code rejects on a non-zero exit code.

raineorshine:
I like the plan. I modernized https://github.com/raineorshine/spawn-please so it's ready for contributions. Breaking upgrades are ready to be released together.

PR #1999, opened 2026-08-08
Speed up `--format ownerChanged`

XhmikosR:
- read _npmUser from the two version manifests instead of the full packument
- run getOwnerPerDependency at --concurrency instead of one dependency at a time

---

Noticed this while looking at #961.

`--format ownerChanged` is the most expensive path in ncu, for two unrelated reasons:

- `packageAuthorChanged` requested the full packument just to read `_npmUser` for two versions. The abbreviated packument has the same version list, and `_npmUser` is in the version manifest, so it now resolves the two versions from the abbreviated packument and fetches those two manifests.
- `getOwnerPerDependency` awaited one dependency at a time. It now uses `pMap` at `--concurrency`, like `queryVersions` and `getPeerDependenciesFromRegistry`.

30 outdated dependencies, gzipped wire bytes:

| | requests | bytes |
| --- | --- | --- |
| before | 30 sequential | 6.69 MB |
| after | 90 at concurrency 8 | 4.84 MB |

The extra requests are 1-3 KB each and now overlap, so effective round trips drop from 30 to ~11. Per package the spread is wide: mocha 871 KB -> 97 KB, htmlparser2 38 KB -> 30 KB, and a package with very few versions costs ~2 KB more.

---

**I'm not sure we want this to land at all, just opening for discussion and testing.**

### DIFF
--- a/src/package-managers/npm.ts
+++ b/src/package-managers/npm.ts
@@ -697,23 +697,28 @@ export async function packageAuthorChanged(
 ): Promise<boolean> {
   const npmConfig = npmApi.findNpmConfig()
   // merge the project/cwd .npmrc so a scoped private registry is respected, like the main fetch path
-  const npmConfigMerged = mergeNpmConfigs(
-    { npmConfigUser: { ...npmConfig, fullMetadata: true }, npmConfigLocal },
-    options,
-  )
+  const npmConfigMerged = mergeNpmConfigs({ npmConfigUser: { ...npmConfig }, npmConfigLocal }, options)
   const result = await fetchPartialPackument(packageName, ['versions'], null, npmConfigMerged)
-  if (result.versions) {
-    const pkgVersions = Object.keys(result.versions)
-    const current = nodeSemver.minSatisfying(pkgVersions, currentVersion)
-    const upgraded = nodeSemver.maxSatisfying(pkgVersions, upgradedVersion)
-    if (current && upgraded && result.versions[current]._npmUser && result.versions[upgraded]._npmUser) {
-      const currentAuthor = result.versions[current]._npmUser?.name
-      const latestAuthor = result.versions[upgraded]._npmUser?.name
-      return currentAuthor !== latestAuthor
-    }
-  }
+  if (!result.versions) return false
 
-  return false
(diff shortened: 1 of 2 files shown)


raineorshine:
> I'm not sure we want this to land at all, just opening for discussion and testing.

Why not? Looks like a good optimization.

XhmikosR:
I don't have anything against this, just the fact that it's more requests.

raineorshine (commented):
Right, it could potentially decrease performance when there is high network latency.

Other suggestions:

- The two manifest fetches could be skipped entirely when current === upgraded (no-op comparison), avoiding 2 requests.
- Since concurrency and the extra hop are independent improvements, it may be worth landing the pMap change on its own.

XhmikosR:
I split the changes in two branches:

* `xmr/owner-changed-concurrency` has only the pMap change
* `xmr/owner-changed` has only the manifest fetches

They touch different files so they can go in either order.

I added the no-op short circuit, but it does not seem to fire in practice. I checked 150 upgrades on two projects with `--target latest`, `semver`, `minor` and `patch`, and `current === upgraded` never came up. `getOwnerPerDependency` only gets the deps that were actually upgraded, so `maxSatisfying(to)` is always above `minSatisfying(from)`. It's one line and `packageAuthorChanged` is exported, so I kept it, but don't expect it to save any requests from the CLI.

About the latency: it's 28% fewer bytes for 3x the requests, so it wins when bandwidth is the bottleneck and loses when latency is. The extra requests are 1-3 KB each and they overlap now, so on a 30 dep project the effective round trips go from 30 to about 11. The spread per package is big: mocha goes from 871 KB to 97 KB, while a package with very few versions costs ~2 KB more.

raineorshine:
Thanks. I merged #2017.

I'm going to hold off on this PR. I don't think it's a good idea to optimize bandwidth at the expense of users with high latency. They tend to be the ones most sensitive to network performance.

XhmikosR:
It still is probably worth landing this. We can't cover all use cases and if it improves things for most users, then it's worth it IMHO.

raineorshine:
I'm not convinced that it is an improvement for most users.

PR #2011, opened 2026-08-10
Fix crash when a dist-tag points to a version missing from the packument

XhmikosR:


src/package-managers/npm.ts
```diff
@@ -1061,21 +1061,20 @@ export const distTag: GetVersion = async (
...
+  }
 
   const publishTime = packument?.time?.[version!]
   const maybeTime = publishTime ? { time: publishTime } : null
   const current = (nodeSemver.validRange(currentVersion) && nodeSemver.minVersion(currentVersion)?.version) || '0.0.0'
 
   const isSatisfiesCooldown =
     tagPackument.version === current ||
-    (tagPackument && satisfiesCooldownPeriod(packageName, tagPackument.version, publishTime, options.cooldown))
+    satisfiesCooldownPeriod(packageName, tagPackument.version, publishTime, options.cooldown)
```
raineorshine, line 1082: I think this causes the fallback packument to silently bypass the deprecated/engines/pre/peer filters in `filterPredicate` below. So with `--engines-node` or `--no-deprecated`, a dist-tag version missing from `versions` gets accepted unchecked instead of falling through to `greatest`.

The old code had the same hole, but only on the `!packument.versions` path, which in practice was rare, because `fetchUpgradedPackument` force-adds `versions` to the fetched fields whenever `!options.deprecated` or `options.enginesNode`:

```ts name=src/package-managers/npm.ts url=https://github.com/raineorshine/npm-check-updates/blob/3eb7a1e918898cc5ed5e8d2235b340355e8c5e9a/src/package-managers/npm.ts#L769-L774
          'dist-tags',
          ...fields,
          ...(!options.deprecated ? (['deprecated', 'versions'] as const) : []),
          ...(options.enginesNode ? (['engines', 'versions'] as const) : []),
```

raineorshine (changes requested):

src/package-managers/npm.ts
XhmikosR: You're right about deprecated and engines. The fallback has no `deprecated` and no `engines`, so `allowDeprecatedOrIsNotDeprecated` and `satisfiesNodeEngine` both pass. `pre` and `peer` are fine though: `allowPreOrIsNotPre` only needs `version`, and `satisfiesPeerDependencies` only needs `name` + `version`, both of which the fallback has.

One correction on the old code. On the `packument.versions` path it didn't silently accept, it threw. `tagPackument.version === current` runs before the `tagPackument &&` guard, so that guard was already dead. That's the crash I was fixing, so the change turned a crash into an unchecked accept.

Fixed by gating the accept on the same condition that force-adds `versions`:

```ts
const isUnfilterable = !versionPackument && (!options.deprecated || options.enginesNode)
```

If `versions` was fetched for filtering and the tagged version isn't in it, it falls through to `greatest`.

I had to split the test too. The original passed `{ distTag: 'latest' }` with no `deprecated`, which is falsy, so it was asserting the hole. Now it's three cases: fallback works with `deprecated: true` (the runtime default), and it's rejected with `--no-deprecated` and `--engines-node`. The two negative ones use a custom dist-tag so they short-circuit before `greatest` and don't hit the network.

raineorshine (approved):

PR #2012, opened 2026-08-10
CI: switch to ubuntu-slim image for low priority workflows

XhmikosR:
Testing this to see if it's worth it.

EDIT: lint seems to be 3x slower, build is OK.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ permissions:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     timeout-minutes: 15
     if: github.event_name != 'pull_request' || github.event.pull_request.changed_files > 0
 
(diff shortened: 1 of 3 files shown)


raineorshine (commented):
Probably stick with `ubuntu-latest` so we get the benefits of caching.

XhmikosR:
Caching should be enabled regardless of the runner. Unless you know something I don't know :)

raineorshine:
You're right. I was thinking they might use different caches, but I guess the cache keys are the same either way.

3x lint slowdown is undesirable though.

XhmikosR:
Yeah, I'm not going to land something that slows us down on every push :)

That being said, I still want to experiment with this more. Build looks like is worth it.

XhmikosR:
So from all this I only see build worth switching and that by not so much. The release workflow, I'm not sure `gh` is installed in the slim image at all. The lint workflow is slower. It's acceptable but it's slower.

PR #2018, opened 2026-08-11
Replace chalk with styleText

XhmikosR:
This is something I've been meaning to do for some time now since our supported Node.js versions support this natively.

We still have chalk in our deps tree as a dependency of others. update-notifier -> boxen pulls chalk 5.6.2, so this only drops the direct dep, about 10 kB off build/index.js. It's still one less dep to keep current, we just had a chalk 5 -> 6 major bump in #1975. 

Only `src/lib/chalk.ts` imported chalk, so this swaps that wrapper's internals and renames it to `src/lib/style.ts` (`style`, `styleInit`, `getStyle`).

styleText differs from chalk in three ways the wrapper now handles:

* it throws on non-strings
* `styleText('red', '')` returns escape codes instead of `''` (colorizeDiff hits this on unchanged versions)
* it leaves the style open across newlines instead of closing per line, which matters for the multi-line update-notifier message boxen puts a border around

The tradeoff is we now own that handling instead of chalk, which is why there are tests for each.

Color is resolved once in `getStyle` and everything uses `validateStream: false`, since styleText otherwise re-runs color detection on every call (roughly 10x slower per call, which works out to well under a millisecond in practice).

Diffed old vs new CLI output for `--help <option>` and the upgrade path including `--format group`: byte-identical piped, under `FORCE_COLOR` and under `NO_COLOR`.

Note that current main is already approximately -90 production deps compared to v16.x.

That being said, chalk is not the biggest of our deps, so I understand if we don't land this. One of our biggest deps chains is update-notifier now. I tried to replace it in another branch.


raineorshine (approved):
Thanks. I'm fine with it either way.

XhmikosR:
I want to have another look at it later in case I missed something before we land this, then.

XhmikosR:
I tested this further and I don't see anything broken. Please review the changes yourself too. The performance regression is real, but when making thousands of calls and not in our real use cases AFAICT.

raineorshine (approved):
Thanks, looks good.

PR #2019, opened 2026-08-11
Replace update-notifier with latest-version

XhmikosR:
Testing this to see what we might be missing with a hand-rolled solution, which admittedly, I don't prefer us having to maintain.

This is one of our longest deps chain. I tried to find a replacement package without success so far, but I'll look into it more later.

If we go with this solution, I need to add tests.

This saves ~ 30 production deps.


raineorshine:
sindresorhus is a good dev and keeps his packages clean and up-to-date. I'm okay keeping the dependency and not having the maintenance burden unless there is a big advantage otherwise.

XhmikosR:
I hear you, and like I said above I don't like the idea of us having to maintain this. I'm still experimenting with it because it's worth it as things are now: https://npmgraph.js.org/?q=update-notifier

It's a ~~huge~~ big package, but unfortunately I haven't found a clean replacement package for it. If it's updated to be more streamlined this PR will be moot.

Let's leave it open so that we don't forget about it and in case I find another solution.

PR #2022, opened 2026-08-13
Sanitize registry and package manager text in error output

XhmikosR:
Follow up to #1994.

All these, including #1994 seem a little far-fetched, but since we landed that, we might as well fix the remaining sites...

### DIFF
--- a/src/lib/logging.ts
+++ b/src/lib/logging.ts
@@ -211,7 +211,12 @@ export async function toDependencyTable({
             format?.includes('installedVersion') || format?.includes('homepage') || format?.includes('repo')
               ? await getPackageJson(dep, { pkgFile })
               : null
-          const from = (format?.includes('installedVersion') ? packageJson?.version : fromDeps[dep]) || ''
+          // node_modules can hold a bogus version, so fall back to the declared spec.
+          // semver.valid tolerates padding like a trailing CR, so that is checked separately.
+          const installedVersion = format?.includes('installedVersion') ? packageJson?.version : undefined
+          const installedIsClean =
+            !!installedVersion && installedVersion.trim() === installedVersion && !!semver.valid(installedVersion)
+          const from = (installedIsClean ? installedVersion : fromDeps[dep]) || ''
           const depType =
             dep in (pkg?.devDependencies ?? {})
               ? 'dev'
--- a/src/lib/logging.ts
+++ b/src/lib/logging.ts
@@ -72,6 +72,18 @@ export function print(
   }
 }
 
+/** Strips escape sequences and control characters from package- and registry-supplied text. Wrap the untrusted substring, not a finished message, since chalk styling uses escape sequences too. */
+export function sanitizeForDisplay(str: string): string {
+  // stripVTControlCharacters leaves bare control characters like CR, which can overwrite a rendered line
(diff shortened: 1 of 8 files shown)


raineorshine:
The good thing about #1994 is that it limited `sanitizeForDisplay` usage to [logging.ts](https://github.com/raineorshine/npm-check-updates/blob/main/src/lib/logging.ts). I don't know if I am a fan of spreading it throughout the codebase. It shouldn't really be the concern of other modules. This PR violates that encapsulation to me.

XhmikosR:
You're right, moved it back into logging.ts.

sanitizeForDisplay is only imported by programError.ts and npm.ts now. programError is part of logging anyway, and doing it there covers every caller at once instead of each one sanitizing itself. The npm.ts one is a single call for the cooldown publishTime, which is a raw registry timestamp.

The rest goes through two new logging exports, printError and printErrorVerbose. index.ts was calling console.error directly and skipping logging entirely, so that's gone too.

I dropped the calls in yarn.ts, pnpm.ts and npm.ts parseJson. Those are thrown Errors and they already get sanitized where they're printed, either printErrors from #1994 or programError now, so they were redundant.

One thing I want your opinion on: index.ts and doctor.ts pipe the package manager stderr straight to the console. Sanitizing those would strip npm's own colors from `ncu --install` and `--doctor` output, so I left them alone.

Also, I split the semver commit. The part that changes upgrade selection is its own commit now, so it's easy to drop or move to another PR.

src/lib/logging.ts
```diff
@@ -72,6 +72,29 @@ export function print(
...
+/** Prints an error to stderr. Strips escape sequences, since error text usually comes from a registry or package manager. */
+export function printError(err: unknown, { color = false }: { color?: boolean } = {}) {
+  const message = sanitizeForDisplay(errorText(err))
+  console.error(color ? chalk.red(message) : message)
+}
+
+/** Prints a caught error under a description of what failed, at the verbose loglevel. */
+export function printErrorVerbose(options: Options, context: string, err: unknown) {
+  print(options, `\n${context}:\n${sanitizeForDisplay(errorText(err))}`, 'verbose')
+}
```
raineorshine: I would prefer not to create separate functions for different combinations of print parameters. We already have a simple, centralized `print` function that can handle different settings. If you can extend that, even by detecting the message type at runtime, that would be ideal.

In this case polymorphism results in a simpler, more readable API. `print` could also take an options object if the positional args are getting too long.

raineorshine (changes requested):

src/lib/logging.ts
XhmikosR: I can fold them in.

printError skips print's loglevel/json gating on purpose, it mirrors what the old console.error calls did in the unhandledRejection handler and the install failure catch, both meant to always show even under --json or --loglevel silent. Merge into print() as-is and those two go quiet under --json, which nobody asked for.

Keep the bypass for those two, or is going quiet under --json/--silent fine for a crash-path error?

raineorshine: They should probably go quiet under --json/--silent. I don't think we should be writing to the console when `--json` is specified, and the exit code should be the only feedback with `--silent`.

raineorshine (approved):

PR #2026, opened 2026-08-17
Add support for devEngines

### LINKED ISSUE #1504: Feature request: support update `devEngines.packageManager.version`
`devEngines` is supported by npm (since v10.9.0) and corepack (sice v0.32.0). It looks like a successor of `packageManager`. Supporting this would be greate.

XhmikosR:
Fixes #1504.

Not sure if we should make it opt-in or not. If not, I can drop the last patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ Options that take no arguments can be negated by prefixing them with `--no-`, e.
   </tr>
   <tr>
     <td>--dep &lt;value&gt;</td>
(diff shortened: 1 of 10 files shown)


raineorshine (approved):
I would do opt-in for now. It's easier to start there and go to default later if it's really needed.

XhmikosR:
You can squash when merging then @raineorshine :)

PR #2030, opened 2026-08-23
feat: add CLIOption.requires to document option dependencies

raineorshine:
## Why

Some options only do anything when another option is also specified, but the generated usage examples didn't say so:

- `--interactiveSelect` only applies with `-i`, yet the help and README showed `ncu --interactiveSelect [value]`.
- `--doctorInstall` / `--doctorTest` only apply in doctor mode.
- `--doctor`'s `-u` requirement was hardcoded as an `option.long === 'doctor'` special case in `renderExtendedHelp`.

## What changed

**New `requires` field on `CLIOption`** — the long names of other options required for the option to work:

```ts
requires?: string[]
```

`renderExtendedHelp` prepends them to the usage examples, so `ncu --help <option>` and the auto-generated README Advanced Options section stay in sync with the option definitions:

- Prefers the short flag (`-i`, `-u`, `-d`) and includes the required option's arg (`-r [uri]`).
- Resolves **transitively**, ordered least → most specific, with cycle/duplicate protection — `--doctorInstall` requires `--doctor`, which requires `-u`, giving `ncu -u -d --doctorInstall [command]`.
- Not prepended to the `--no-` negation line, since the requirements don't apply when disabling the option.
- Throws on an unknown option name, so a typo fails the build.

This replaces the hardcoded `--doctor` special cases. The three rendering functions moved below `cliOptionsMap` since they need it for lookups (`no-use-before-define` is enforced).

**Options configured:** `--doctor` → `upgrade`; `--doctorInstall` / `--doctorTest` → `doctor`; `--interactiveSelect` → `interactive`; `--cacheExpiration` → `cache`; `--registryType` → `registry`.

**README** regenerated with `npm run build:options`:

```diff
-ncu --interactiveSelect [value]
+ncu -i --interactiveSelect [value]
-ncu --registryType [type]
+ncu -r [uri] --registryType [type]
-ncu --doctor -u        →  ncu -u --doctor
-ncu -du                →  ncu -u -d
```

**Tests** — a `requires` block in `test/cli-options.test.ts` validating that every reference resolves to a real option, plus the prefix, transitivity, and unprefixed negation behavior.

## Reviewer notes

- The `--doctor` usage lines changed form: `ncu -u -d` instead of `ncu -du`, since the generic prefix keeps flags separate. Equivalent, but let me know if you'd rather preserve the combined `-du`.
- Three options were left unconfigured because they depend on *one of* several options, which a flat list would misdocument: `--cacheFile` (`--cache` **or** `--cacheClear`), `--root` (`--workspace` **or** `--workspaces`, and it defaults to `true` so the meaningful usage is `--no-root`), and `--mergeConfig` (`--deep` **or** `--packageFile`). The field could be extended to accept nested arrays for "one of" (`requires: [['cache', 'cacheClear']]`) if that's wanted.
- `--registryType` is a judgment call: `--registry` is strictly required only for `registryType json`, though `--registryType` is a no-op without it either way.
- `npm run test:unit` (986 passed, 1 skipped), `tsc --noEmit`, eslint, and prettier are all clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -565,9 +565,9 @@ Skipped due to 7-day cooldown
 Usage:
 
 ```sh
-ncu --doctor -u
+ncu -u --doctor
 ncu --no-doctor
-ncu -du
+ncu -u -d
 ```
 
 Iteratively installs upgrades and runs your project's tests to identify breaking upgrades. Reverts broken upgrades and updates package.json with working upgrades.
@@ -773,7 +773,7 @@ Control the auto-install behavior.
 Usage:
 
 ```sh
-ncu --interactiveSelect [value]
+ncu -i --interactiveSelect [value]
 ```
 
 Default: auto
@@ -867,7 +867,7 @@ ncu-test-return-version  1.0.0  →  2.0.0
 Usage:
 
 ```sh
-ncu --registryType [type]
+ncu -r [uri] --registryType [type]
 ```
 
 Specify whether `--registry` refers to a full npm registry or a simple JSON file.
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -43,45 +43,6 @@ const parseNumberOption =
     throw new Error(`${optionName} must be a number`)
   }
 
-/** Renders the extended help for an option with usage information. */
-export const renderExtendedHelp = (option: CLIOption, { markdown }: { markdown?: boolean } = {}) => {
-  let output = ''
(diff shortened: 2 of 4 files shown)


XhmikosR:
I'll have a look after we are done with my open PRs because I spent a lot of time splitting the patches :)

PR #2032, opened 2026-08-24
Use lib/exists in pnpm adapter instead of a local pathExists

XhmikosR:
BTW unrelated @raineorshine, you should protect the main branch to require reviews to merge. Right now everyone with push rights (unsure if we are the only ones, but still) can merge even without a review.


raineorshine (approved):

raineorshine:
> BTW unrelated @raineorshine, you should protect the main branch to require reviews to merge. Right now everyone with push rights (unsure if we are the only ones, but still) can merge even without a review.

Thanks, added.

PR #2060, opened 2026-08-30
Replace the #catalog path sentinel with a catalog flag on PackageInfo

XhmikosR:
@raineorshine: I have rebased my other branches but all of them belong to the previous one. So it doesn't make a lot of sense to open them now. Unsure how to proceed because it needs to be sequential... One reviewed, then rebase and the next one, etc.


raineorshine (approved):

raineorshine:
Sorry, I don't understand what you're asking.

PR #2061, opened 2026-08-30
Replace fp-and-or with native code

XhmikosR:
Doesn't save us a lot, a few milliseconds in some cases, but it's one less dependency.

I know it's your package, but we only use it once. It's your call, @raineorshine :)


raineorshine (approved):
I'm fine removing it.

PR #2065, opened 2026-08-30
Expose spawn and usesAddCommand on the package manager adapters

XhmikosR:


src/package-managers/pnpm.ts
```diff
@@ -402,3 +407,19 @@ export const pnpmApi = {
   getPnpmWorkspaceRegistries,
   parseList,
 }
+
+/**
+ * Reads pnpm's minimumReleaseAge as a cooldown.
+ * pnpm does not read .npmrc min-release-age; only its own native config is consulted.
+ */
+export const getCooldown = async (options: Options): Promise<NativeCooldown | null> => {
+  const config = await pnpmApi.getPnpmWorkspaceMinimumReleaseAge({ cwd: options.cwd })
```
raineorshine: `pnpmApi` is just for export. Use `getPnpmWorkspaceMinimumReleaseAge` un-namespaced within the module.

raineorshine (changes requested):

src/package-managers/pnpm.ts
XhmikosR: Fixed

src/package-managers/npm.ts
```diff
@@ -1354,4 +1357,44 @@ export const semver: GetVersion = async (
   return toVersionResult({ ...packageInfo, ...versionResult })
 }
 
+/** Reads npm's min-release-age config as a cooldown. Also used for package managers with no native setting. */
+export const getCooldown = async (): Promise<NativeCooldown | null> => {
+  const config = npmApi.findNpmConfig()
```
raineorshine, line 1366: 👀

src/package-managers/yarn.ts
```diff
@@ -449,10 +450,30 @@ export const packageAuthorChanged = async (
...
   parseJsonLines,
   extractFirstJsonLine,
 }
+
+/**
+ * Reads yarn's npmMinimalAgeGate as a cooldown.
+ * yarn does not read .npmrc min-release-age; only its own native config is consulted.
+ */
+export const getCooldown = async (options: Options): Promise<NativeCooldown | null> => {
+  const config = await yarnApi.getYarnMinimalAgeGate(options)
```
raineorshine: 👀

raineorshine (changes requested):
Thanks.

Sorry I didn't mark the others individually.

src/package-managers/npm.ts
XhmikosR: They are needed for tests. And the other change broke them...

raineorshine: I would prefer if the function was memoized and given its own module-level identifier for use within the module. That would more conventional then strapping it onto the export bundle and then accessing that within the module.

XhmikosR: Not sure about making this change here. Please check if the current solution is acceptable.

raineorshine (approved):

PR #2066, opened 2026-08-30
Apply option parse to ncurc and module values, not just the cli path

XhmikosR:


raineorshine (approved):

raineorshine:
Found a few gaps actually. I'll add test coverage to `main` and post back soon.

src/lib/getNcuRc.ts
```diff
@@ -122,10 +122,9 @@ async function getNcuRc({
 
   const filePath = rawResult?.filepath
 
-  // convert the config to valid options by removing $schema and parsing format
+  // convert the config to valid options by removing $schema. Values are coerced in initOptions.
   const { $schema: _, ...rawConfig } = rawResult?.config || {}
-  const config: Options = rawConfig || {}
-  if (typeof config.format === 'string') config.format = cliOptionsMap.format.parse!(config.format)
+  const config: Options = rawConfig
```
raineorshine, line 127: This is the regression. The per-package `.ncurc` reloaded at `src/index.ts:258` is merged into `pkgOptions` at line 266, long after `initOptions` has run, so nothing coerces it any more.

An unparsed string `format` is not merely inert there — it inverts: `'no-group'.includes('group')` is `true`, so `{ "format": "no-group" }` in a nested `.ncurc` enables the grouping it is meant to disable.

Suggest extracting the loop below as e.g. `parseOptions(options)` and calling it on `pkgOptions` in `runUpgrades`. That closes this and the pre-existing gap for every other option in a nested config — `cooldown: "7d"` and `dep: "prod,dev"` were never coerced there either.

src/lib/initOptions.ts
```diff
@@ -115,6 +115,23 @@ async function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } =
...
+  // The cli path already went through commander, which applies parse. Values coming from the ncurc or
+  // the module API have not, so coerce them here from the same option definitions.
+  const rawCooldown = options.cooldown
+  for (const option of cliOptions) {
+    const { parse, accumulate } = option
+    if (!parse || accumulate) continue
+    const key = option.long as keyof Options
+    const value = options[key]
+    // defaults are already in their coerced form, and commander does not parse them either
+    if (value === undefined || value === option.default) continue
```
raineorshine: This line is load-bearing and silent when it breaks — please pin it. `cacheFile`'s default is `~/.ncu-cache.json`, and `parse` on it returns `<cwd>/~/.ncu-cache.json`, which `resolveCacheFile` no longer recognizes as the default, so the cache quietly moves to a junk path.

```ts
expect((await initOptions({})).cacheFile).toBe(defaultCacheFile)
```

src/lib/initOptions.ts
```diff
@@ -115,6 +115,23 @@ async function initOptions(runOptions: RunOptions, { cli }: { cli?: boolean } =
...
+    try {
+      options[key] = parse(value) as never
+    } catch (err: any) {
+      programError(options, err.message || err)
```
raineorshine: The coercion itself has no coverage. Please add to `test/initOptions.test.ts`:

- values are coerced: `concurrency: '4'` → `4`, `dep: 'prod,dev'` → `['prod', 'dev']`, `format: 'no-group,time'` → `['time']`, `cooldown: '12h'` → `0.5`
- a parse failure surfaces as a program error rather than a raw throw: `concurrency: 'abc'` rejects with `concurrency must be a number`, `cacheFile: 5` with `cacheFile must be a string`
- the `accumulate` skip keeps arrays flat: `reject: ['foo']` stays `['foo']` and not `[['foo']]`, and `initOptions({})` yields `workspace: []` — unguarded, `[...accum, value]` throws on the undefined accumulator

src/cli-options.ts
```diff
@@ -947,7 +949,9 @@ const cliOptions: CLIOption[] = [
     description:
       'Include prerelease versions, e.g. -alpha.0, -beta.5, -rc.2. Automatically set to 1 when `--target` is newest or greatest, or when the current version is a prerelease. (default: 0)',
     parse: (value: unknown): boolean => {
-      if (typeof value === 'number') {
+      if (typeof value === 'boolean') {
```
raineorshine, line 952: This branch exists only so the second parse survives: commander coerces `--pre 1` to `true`, then the new loop parses it again. I removed it and rebuilt — `ncu --pre 1` exits with `pre must be a number`. Nothing currently guards that, so please add a CLI test in `test/bin.test.ts`.

src/types/CLIOption.ts
```diff
@@ -9,7 +9,17 @@ export interface CLIOption<T = any> {
   deprecated?: boolean
   description: string
   help?: ExtendedHelp
-  /** Must be prepared to handle unknown input types since the user's ncurc.json may not match the schema. */
+  /**
+   * True if parse doubles as commander's accumulator for a repeatable option, i.e. it appends to the
+   * previous value rather than coercing a single one. Such a parse is not idempotent, so it is only
+   * applied by commander and never to values coming from the ncurc or the module API.
+   */
+  accumulate?: boolean
```
raineorshine, line 17: Both contracts documented here should be tests in `test/cli-options.test.ts`, so a future option cannot break the loop silently:

- **arity matches the flag** — every `parse` that takes an accumulator (`parse.length > 1`) is marked `accumulate`, and no arity-1 parse is. Holds exactly today across all 15 options with a `parse`.
- **idempotency** — table-driven `parse(parse(x))` equals `parse(x)` for each non-accumulate option. This is what the cli path now relies on, since commander parses and then the loop parses again.

raineorshine (changes requested):
Suggestions below

XhmikosR:
@raineorshine: thanks for catching the bug. Please check one more time.

raineorshine:
Thanks for the update.

One regression came in with the fix. `getNcuRc` no longer parses `format`, but `parseOptions` only runs on `pkgOptions` after `mergeOptions`. In `--deep --mergeConfig` mode, a nested `format` is still a string at merge time, so it replaces the root's array instead of merging into it. Root `.ncurc` `format: "time"` with a nested `format: "repo"` gives `['time', 'repo']` on `main`, but `['repo']` here.

Parsing the config before it is merged restores this. `parseOptions` is idempotent, so the later pass on `pkgOptions` is harmless.

Thanks!

fyi I've been on vacation and will be back next week

XhmikosR:
I hope everything's fixed now.

Have fun!

PR #2067, opened 2026-08-30
fix: ignore versionless packageManager field

### LINKED ISSUE #2055: Running ncu against a pnpm project with pnpm-workspace.yaml that has overrides crashes with unhandled rejection.
- [x] I am using the [latest version](https://www.npmjs.com/package/npm-check-updates?activeTab=versions) of `npm-check-updates`
- [x] I am using a [compatible version of node](https://github.com/raineorshine/npm-check-updates/blob/main/package.json#L20)

---

Regression introduced by #1767 (v22.2.6) added new call sites for `isPackageManagerProtocol(version)` without guarding against version being undefined, which occurs when ncu resolves packages from workspace overrides that are not direct dependencies.

## Steps to Reproduce

- create a pnpm project with `"packageManager": "pnpm"`
- Add a `pnpm-workspace.yaml` with overrides pointing to transitive (non-direct) dependencies:
```
overrides:
    "@typescript-eslint/utils": "8.56.0"
    "diff: "8.0.3"
```

Ru: ncu --target latest

## Current Behavior

TypeError: Cannot read properties of undefined (reading 'startsWith')
       at isPackageManagerProtocol  (index.js:4718)
Unhandled Rejection! This is a bug and should be reported: https://github.com/raineorshine/npm-check-updates/issues

## Expected Behavior

ncu completes successfully

onuracar-dev:
## Summary

- ignore a `packageManager` entry when it does not declare a version
- preserve the existing field while continuing to upgrade normal dependencies
- add an end-to-end regression for the reported `packageManager: "pnpm"` crash

## Validation

- `vitest run test/dep.test.ts` (20 passed)
- `tsc --noEmit`
- targeted ESLint and Prettier checks
- `npm run build`
- `git diff --check`

AI-assisted; tested and reviewed manually.

Fixes #2055

### DIFF
--- a/src/lib/getCurrentDependencies.ts
+++ b/src/lib/getCurrentDependencies.ts
@@ -22,7 +22,7 @@ const isGreaterThanSafe = (spec1: VersionSpec, spec2: VersionSpec) =>
 const parsePackageManager = (pkgData: PackageFile) => {
   if (!pkgData.packageManager) return {}
   const [name, version] = pkgData.packageManager.split('@')
-  return { [name]: version }
+  return version ? { [name]: version } : {}
 }
 
 /** Parses the devEngines.packageManager field into a { [name]: version } collection. */
--- a/test/dep.test.ts
+++ b/test/dep.test.ts
@@ -389,6 +389,45 @@ describe('--dep', () => {
       }
     })
 
+    it('ignore a packageManager field without a version', async () => {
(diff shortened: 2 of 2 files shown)


raineorshine (approved):

raineorshine:
Thanks!

PR #2068, opened 2026-08-31
Add a makeTempDir test helper and use it for all temp dir creation

XhmikosR:


raineorshine (approved):

XhmikosR:
It seems this landed after something else landed first and it broke CI. See #2080

PR #2073, opened 2026-09-08
Revamp build-changelog.ts

XhmikosR:


package.json
```diff
@@ -109,7 +110,7 @@
     "jsonc-parser": "^3.3.1",
     "lilconfig": "^3.1.3",
     "lockfile-lint": "^5.0.1",
-    "markdownlint": "^0.41.1",
+    "markdown-it": "^15.0.1",
```
XhmikosR, line 113: I went with v15 even though it's not deduplicated right now, but I assume markdownlint will be updated eventually.

raineorshine (approved):

PR #2074, opened 2026-09-08
package.json: use the lint:types script in test

XhmikosR:


package.json
```diff
@@ -46,7 +46,7 @@
     "prettier": "prettier . --check",
     "prettier:fix": "prettier . --write",
     "vitest": "vitest",
-    "test": "tsc --noEmit && npm run test:unit && npm run test:e2e",
+    "test": "npm run lint:types && npm run test:unit && npm run test:e2e",
```
XhmikosR, line 50: Alternatively we could use:

```suggestion
    "test": "npm-run-all lint:types test:unit test:e2e",
```

raineorshine (approved):
Looks good.

You could use `npm-run-all`, but then you'd have to add `cross-env FORCE_COLOR=1`

PR #2085, opened 2026-09-10
Keep non-violating upgrades when the peer recursion cannot resolve a conflict

### LINKED ISSUE #2072: updates available but not recognized and not applied
- [yes] I have searched for [similar issues](https://github.com/raineorshine/npm-check-updates/issues)

Not found.

- [yes] I am using the [latest version](https://www.npmjs.com/package/npm-check-updates?activeTab=versions) of `npm-check-updates`

```
$ ncu -v
23.1.0
```

- [yes] I am using a [compatible version of node](https://github.com/raineorshine/npm-check-updates/blob/main/package.json#L20)

```
$ node -v
v26.8.1
```

---

## Steps to Reproduce

.ncurc:

```
$ cat ~/.ncurc.json 
{
  "peer": true,
  "cooldown": 3
}
```

npm:

```
$ npm -v
12.0.2
```

Dependencies:

```
$ cat package.json 
{
  ...
  "dependencies": {
    "react": "19.2.8",
    "react-dom": "19.2.8"
  },
  "devDependencies": {
    "@eslint/js": "9.39.1",
    "@vitejs/plugin-react": "6.1.0",
    "eslint": "9.39.5",
    "eslint-plugin-react": "7.37.5",
    "eslint-plugin-react-hooks": "7.1.1",
    "globals": "17.11.0",
    "vite": "8.2.2"
  }
}
```

Steps:

```
$ ncu
Using config file /home/[...redacted...]/.ncurc.json
Checking /home/[...redacted...]/vite-project/package.json
[====================] 9/9 100%
[====================] 9/9 100%
[====================] 3/3 100%
All dependencies not in cooldown match the latest package versions :)
[====================] 4/4 100%

 @eslint/js             9.39.1  →   10.0.1  reason: @eslint/js requires eslint ^10.0.0
 @vitejs/plugin-react    6.1.0  →    6.1.1  reason:
 eslint                 9.39.5  →   10.9.1  reason: eslint-plugin-react requires ^3 || ^4 || ^5 || ^6 || ^7 || ^8 || ^9.7
 globals               17.11.0  →  17.12.0  reason:
```

## Current Behavior

There are updates available for packages `@vitejs/plugin-react` and `globals` and there is no reason to skip these updates (no peer dependency problems, no cooldown period), but no such updates are applied and `ncu` sees these packages in `incompatible` section (as shown above).

## Expected Behavior

Both packages `@vitejs/plugin-react` and `globals` should be updated and other incompatible packages should be skipped.

raineorshine:
Fixes #2072.

## Cause

When `--peer` is on and the peer recursion stops making progress, [`upgradePackageDefinitions`](https://github.com/raineorshine/npm-check-updates/blob/main/src/lib/upgradePackageDefinitions.ts#L158) bails out with an empty upgrade set, discarding **every** pending upgrade — including the ones that never violated a peer dependency. The correct set is already sitting in `checkPeerViolationResult.filteredUpgradedDependencies` at that point.

Those discarded packages then reappear under `Ignored incompatible updates (peer dependencies)` with an empty `reason:`, because `getIgnoredUpgradesDueToPeerDeps` reports anything missing from the upgrade set as peer-ignored.

Introduced by #1421, which is correct for the all-violate case but too broad for the mixed case.

## Repro

The reporter's `package.json`, with `--peer`:

```
All dependencies not in cooldown match the latest package versions :)

 @eslint/js             9.39.1  →   10.0.1  reason: @eslint/js requires eslint ^10.0.0
 @vitejs/plugin-react    6.1.0  →    6.1.1  reason:
 eslint                 9.39.5  →  10.10.0  reason: eslint-plugin-react requires ^3 || ^4 || ^5 || ^6 || ^7 || ^8 || ^9.7
 globals               17.11.0  →  17.12.0  reason:
```

`@vitejs/plugin-react` and `globals` have no peer conflict at all — an instrumented run confirms the bail-out discards `{@eslint/js, @vitejs/plugin-react, globals}` while the non-violating set holds exactly `{@vitejs/plugin-react: 6.1.1, globals: 17.12.0}`. Cooldown is not involved; `--peer` alone reproduces it.

After the fix:

```
 @vitejs/plugin-react  6.1.0  →  6.1.1

Minor   Backwards-compatible features
 globals  17.11.0  →  17.12.0

 @eslint/js  9.39.1  →   10.0.1  reason: @eslint/js requires eslint ^10.0.0
 eslint      9.39.5  →  10.10.0  reason: eslint-plugin-react requires ^3 || ^4 || ^5 || ^6 || ^7 || ^8 || ^9.7
```

The empty-`reason:` rows go away as a side effect.

## Change

Return `checkPeerViolationResult.filteredUpgradedDependencies` instead of `{}`. The all-violate case is unaffected — that set is empty there, which is what `ignores if post upgrade peers are unmet - no upgrades` already asserts.

## Tests

`test/peerPartialUpgrade.test.ts` covers the mixed case: `peer-shifter` only requires `pinned-pkg ^2` once upgraded, which sends the recursion one pass deep before the peer map reverts and hits the bail-out. It fails on `main` with `expected {} to strictly equal { 'clean-pkg': '1.1.0' }`.

Full suite passes. The 2 failures in `dep.test.ts` and `spawnCommand.test.ts` are pre-existing on `main`, unrelated to this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


XhmikosR:
I pushed a fix, now I think this is good to go, but have another look yourself too.

